PR #143, opened 2020-02-10
Add docker image download to download-all-data

seanpmorgan:


armory/__main__.py
```diff
@@ -118,6 +120,21 @@ def download_all_data(command_args, prog, description):
     args = parser.parse_args(command_args)
     coloredlogs.install(level=args.log_level)
 
+    print("Downloading all docker images....")
+    docker_client = docker.from_env()
+    docker_images = [
+        "twosixarmory/tf1:0.2.1",
+        "twosixarmory/tf2:0.2.1",
+        "twosixarmory/pytorch:0.2.1",
+    ]
```
davidslater: Do we want these to be at the top level of the document or elsewhere in the repo, so that we can automatically update READMEs, check images in configs, etc. in a programmatic way?

seanpmorgan: Yes :smile:

PR #162, opened 2020-02-17
Mount volumes as host user

seanpmorgan:
We previously removed this strategy since we considered mounting the host's pip installed armory in the container (which could require root to use/modify). 

Since we've since installed the armory package within the container this is no longer needed. Saving files as root onto the user's host caused a number of issues and is a pretty poor practice.

### DIFF
--- a/armory/docker/management.py
+++ b/armory/docker/management.py
@@ -7,7 +7,7 @@ import os
 
 import docker
 
-from armory import paths
+from armory.paths import HostPaths, DockerPaths
 
 
 logger = logging.getLogger(__name__)
@@ -21,17 +21,22 @@ class ArmoryInstance(object):
     def __init__(
         self, image_name, runtime: str = "runc", envs: dict = None, ports: dict = None
     ):
-        self.docker_client = docker.from_env()
+        host_paths = HostPaths()
+        docker_paths = DockerPaths()
+        self.docker_client = docker.from_env(version="auto")
 
-        os.makedirs(paths.OUTPUTS, exist_ok=True)
         container_args = {
             "runtime": runtime,
             "remove": True,
             "detach": True,
             "volumes": {
-                paths.CWD: {"bind": "/workspace", "mode": "rw"},
-                paths.USER_ARMORY: {"bind": "/.armory", "mode": "rw"},
-                paths.USER_KERAS: {"bind": "/.keras", "mode": "rw"},
+                host_paths.cwd: {"bind": "/workspace", "mode": "rw"},
+                host_paths.armory_dir: {"bind": docker_paths.armory_dir, "mode": "rw"},
+                host_paths.dataset_dir: {
+                    "bind": docker_paths.dataset_dir,
+                    "mode": "rw",
+                },
+                host_paths.output_dir: {"bind": docker_paths.output_dir, "mode": "rw"},
             },
         }
         if ports is not None:
--- a/armory/docker/management.py
+++ b/armory/docker/management.py
@@ -46,7 +46,8 @@ class ArmoryInstance(object):
         # directoriy permissions via uid and gid.
         if os.name != "nt":
             user_id = os.getuid()
-            container_args["user"] = f"{user_id}:{user_id}"
+            group_id = os.getgid()
+            container_args["user"] = f"{user_id}:{group_id}"
 
         if envs:
             container_args["environment"] = envs
--- a/armory/docker/management.py
+++ b/armory/docker/management.py
@@ -30,12 +30,18 @@ class ArmoryInstance(object):
             "detach": True,
             "volumes": {
                 paths.CWD: {"bind": "/workspace", "mode": "rw"},
-                paths.USER_ARMORY: {"bind": "/root/.armory", "mode": "rw"},
-                paths.USER_KERAS: {"bind": "/root/.keras", "mode": "rw"},
+                paths.USER_ARMORY: {"bind": "/.armory", "mode": "rw"},
+                paths.USER_KERAS: {"bind": "/.keras", "mode": "rw"},
             },
         }
         if ports is not None:
             container_args["ports"] = ports
+
+        # Windows docker does not require syncronizing file and
+        # directoriy permissions via uid and gid.
+        if os.name != "nt":
+            user_id = os.getuid()
+            container_args["user"] = f"{user_id}:{user_id}"
 
         if envs:
             container_args["environment"] = envs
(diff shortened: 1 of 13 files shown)


armory/paths.py
```diff
@@ -35,10 +35,9 @@
 
 DATASETS = config.get("datasets") or os.path.join(USER_ARMORY, "datasets")
 EXTERNAL_REPOS = os.path.join(DATASETS, "external_repos")
-MODELS = config.get("models") or os.path.join(USER_ARMORY, "models")
```
seanpmorgan: This is not used. We can add it if/when it is needed.

seanpmorgan:
Just a note. This will require users to clear their current .armory directory using `sudo rm -R ~/.armory`. We can add this in the release notes.

We could add a decent sized block of code to try to handle this gracefully, but seeing as we're in such an alpha release I think the release note is fine.

davidslater:
I don't see how these changes enable us to "Mount volumes as host user". The only major change I see is that the paths inside the docker container are different. I don't see any differences in what gets mounted from the host or what permissions we are using.

davidslater:
Oh, is it the:
```
        if os.name != "nt":
            user_id = os.getuid()
            container_args["user"] = f"{user_id}:{user_id}"
```

For some reason, I wasn't seeing them.

armory/paths.py
```diff
@@ -35,10 +35,9 @@
...
 TMP = config.get("tmp") or os.path.join(USER_ARMORY, "tmp")
 OUTPUTS = config.get("outputs") or os.path.join(USER_ARMORY, "outputs")
 
-DOCKER = "/root"
+DOCKER = "/"
```
davidslater: This breaks pulling from datasets in the container --> those go to `/root/.armory`, since armory.data.datasets uses paths.DATASETS, which uses USER_ARMORY as the base (this is the local user outside the container and "root" in the container).

Does writing to "/root" do anything to the host outside of the mounted directories?

seanpmorgan: Updated code so that we no longer use any "user" paths within the container. All docker paths as explicit.

Since we're running the container as host user there is no write permission to anything under `/root` inside of the container. I've merged this PR with master so make sure you `sudo rm -R ~/.armory` on host prior running for the first time.

armory/docker/management.py
```diff
@@ -21,22 +21,33 @@ class ArmoryInstance(object):
...
 
-        os.makedirs(paths.OUTPUTS, exist_ok=True)
         container_args = {
             "runtime": runtime,
             "remove": True,
             "detach": True,
             "volumes": {
-                paths.CWD: {"bind": "/workspace", "mode": "rw"},
-                paths.USER_ARMORY: {"bind": "/root/.armory", "mode": "rw"},
-                paths.USER_KERAS: {"bind": "/root/.keras", "mode": "rw"},
```
seanpmorgan: Dropped USER_KERAS as we can add a more scalable model dir in the future

armory/eval/evaluator.py
```diff
@@ -130,20 +133,28 @@ def run(self, interactive=False, jupyter=False, host_port=8888) -> None:
...
+            user_id = 0
         lines = [
             "Container ready for interactive use.",
             bold(
                 "*** In a new terminal, run the following to attach to the container:"
             ),
-            bold(red(f"    docker exec -itu0 {runner.docker_container.short_id} bash")),
+            bold(
+                red(
+                    f"    docker exec -itu{user_id} {runner.docker_container.short_id} bash"
```
seanpmorgan: Need to make interactive the same user as the launched docker for saving files

armory/docker/management.py
```diff
@@ -21,22 +21,33 @@ class ArmoryInstance(object):
...
             },
         }
         if ports is not None:
             container_args["ports"] = ports
 
+        # Windows docker does not require syncronizing file and
+        # directoriy permissions via uid and gid.
+        if os.name != "nt":
+            user_id = os.getuid()
+            container_args["user"] = f"{user_id}:{user_id}"
```
ng390: Apparently the user id and group id associated with the user's group may not always be the same. On archimedes the group id associated with some directories is a group owned by a different user.

seanpmorgan: Thx for the catch! Pushed a fix.

armory/eval/evaluator.py
```diff
@@ -130,20 +133,30 @@ def run(self, interactive=False, jupyter=False, host_port=8888) -> None:
...
+            group_id = 0
         lines = [
             "Container ready for interactive use.",
             bold(
                 "*** In a new terminal, run the following to attach to the container:"
             ),
-            bold(red(f"    docker exec -itu0 {runner.docker_container.short_id} bash")),
+            bold(
+                red(
+                    f"    docker exec -it -u {user_id}:{group_id} {runner.docker_container.short_id} bash"
```
seanpmorgan: Updated user/group per Neal's bug catch

ng390: The same user message should be on line 163

seanpmorgan: Good catch thanks!. Fixed.

ng390:
LGTM

armory/paths.py
```diff
@@ -5,40 +5,53 @@
...
+        try:
+            with open(path) as f:
+                config = json.load(f)
+        except json.decoder.JSONDecodeError:
+            logger.exception(f"Armory config file {path} could not be decoded")
+            raise
+        except OSError:
+            logger.exception(f"Armory config file {path} could not be read")
+            raise
+    else:
```
davidslater: If the config file exists but there are issues parsing it (e.g., it's an old config like in my case, or because they manually modified it), we should try to raise a more understandable error than:
```
(armory-base) davidslater@aleph-5:~/armory$ python -m armory run examples/fgm_attack.json 
  File "/Users/davidslater/opt/anaconda3/envs/armory-base/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/Users/davidslater/opt/anaconda3/envs/armory-base/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/davidslater/git/armory/armory/__main__.py", line 204, in <module>
    main()
  File "/Users/davidslater/git/armory/armory/__main__.py", line 200, in main
    func(sys.argv[2:], prog, description)
  File "/Users/davidslater/git/armory/armory/__main__.py", line 102, in run
    rig = Evaluator(args.filepath)
  File "/Users/davidslater/git/armory/armory/eval/evaluator.py", line 27, in __init__
    self.host_paths = HostPaths()
  File "/Users/davidslater/git/armory/armory/paths.py", line 46, in __init__
    config = _parse_global_config(self.armory_config, self.armory_dir)
  File "/Users/davidslater/git/armory/armory/paths.py", line 32, in _parse_global_config
    os.makedirs(config["cached_dataset_dir"], exist_ok=True)
KeyError: 'cached_dataset_dir'
```

Perhaps that's something to add with the `armory configure` issue?

seanpmorgan: Thanks. Added a temporary error which we can modify when #165 is complete

PR #164, opened 2020-02-17
Fix import structure for top level imports

seanpmorgan:
Enables for example:
```
import armory
armory.art_experimental.attacks.FGMBinarySearch
```

### DIFF
--- a/armory/art_experimental/attacks/poisoning_attack_backdoor.py
+++ b/armory/art_experimental/attacks/poisoning_attack_backdoor.py
@@ -3,7 +3,7 @@ Simple data poisoning backdoor attack
 """
 
 import numpy as np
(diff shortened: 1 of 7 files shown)


armory/art_experimental/attacks/poisoning_attack_backdoor.py
```diff
@@ -3,7 +3,7 @@
 """
 
 import numpy as np
-from art.attacks.attack import PoisoningAttack
+from art.attacks import PoisoningAttack
```
seanpmorgan: I believe this will break until we update to art-1.1.1

davidslater: Are there current issues preventing updating to art-1.1.1, or do we want to wait for 1.2?

PR #170, opened 2020-02-19
Docker docs

### LINKED ISSUE #119: Provide information on how to kill armory docker containers and cleanup old armory docker images
There are potentially many solutions. The most straightforward is to link to docker documentation (or potentially StackOverflow) in the README. I think these two issues should definitely be something we talk to, at least briefly.

While we could create our own scripts (possibly leveraging pydocker), it's less maintenance burden to point to docker documentation.  Docker is still a daemon running on their system that users should have some knowledge of.

davidslater:
fixes #119

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ On Mac and Windows Desktop versions, this defaults to 2 GB. See the docs to chan
 
 ### Docker Cleanup
 Subsequent versions of ARMORY will download new Docker images, but will not clean up existing ones.
-To display the set of curret images:
+To display the set of current images:
 ```
 docker images
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -86,3 +86,21 @@ Depending on the task, docker memory for an ARMORY container must be at least 8
...
 * [Windows](https://docs.docker.com/docker-for-windows/)
+
+### Docker Cleanup
+Subsequent versions of ARMORY will download new Docker images, but will not clean up existing ones.
+To display the set of current images:
+```
+docker images
+```
+To delete images, see the docs for [docker rmi](https://docs.docker.com/engine/reference/commandline/rmi/)
+or this [StackOverflow post](https://stackoverflow.com/questions/44785585/how-to-delete-all-docker-local-docker-images).
```
seanpmorgan: Not sure we want to link to how to delete all images. A safer call would be:
`docker rmi $(docker images | grep 'twosixarmory')`

davidslater: Would that delete the intermediate images as well? If not, then maybe we should just reference the docker rmi.

seanpmorgan: Not positive. Lets just reference docker rmi for now. We can handle this for the user with #174

PR #179, opened 2020-02-19
Load custom dataset from module

### LINKED ISSUE #160: Enable custom data modules
Currently our config schema only allows datasets to be loaded from the datasets available in armory. It would be helpful to allow users to load their own datasets as well. This is also needed for importing data from armory-private.

seanpmorgan:
closes #160 

* Updates the dataset schema
* Enables arbitrary dataset module
* Creates helper utils for dataset loading and model loading from config

### DIFF
--- a/armory/utils/configuration.py
+++ b/armory/utils/configuration.py
@@ -44,11 +44,13 @@ def _verify_dataset(config: Optional[dict]):
     if config is None:
         return
     else:
-        if config["name"] not in SUPPORTED_DATASETS:
-            raise ValueError(
-                f"Configured data {config['name']} not found in"
-                f" supported datasets: {list(SUPPORTED_DATASETS.keys())}"
-            )
+        if "module" not in config.keys():
+            raise ValueError("A `module` must be specified in dataset json")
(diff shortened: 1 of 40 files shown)


armory/utils/configuration.py
```diff
@@ -44,11 +44,13 @@ def _verify_dataset(config: Optional[dict]):
     if config is None:
         return
     else:
-        if config["name"] not in SUPPORTED_DATASETS:
```
seanpmorgan: Hmmm I think we can remove SUPPORTED_DATASETS now?

tests/test_data/missing_eval.json
```diff
@@ -13,7 +13,8 @@
         "name": "FastGradientMethod"
     },
     "dataset": {
-        "name": "mnist"
+        "module": "armory.data.datasets",
+        "name": "mnist_data"
```
davidslater: Little nit: can we make the `mnist_data` and `cifar10_data` methods (and the references to them) `mnist` and `cifar10`? It's a bit more intuitive and `datasets.mnist_data` feels a bit redundant.

PR #181, opened 2020-02-20
Armory clean

davidslater:


armory/docker/images.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    if not isinstance(tag, str):
+        raise ValueError(f"tag must be of type str, not type {type(tag)}")
+    if tag in ALL:
+        return False
+    tokens = tag.split(":")
+    if len(tokens) != 2:
+        return False
+    if tokens[0] in REPOSITORIES:
+        return True
```
seanpmorgan: Could we make this so that only older versions are removed instead of any versions that doesn't match. Possibly using https://docs.python.org/3/distutils/apiref.html#module-distutils.version

davidslater: Updated to only remove older versions. Tags that don't survive strict parsing will not be removed.

Issue #104, opened 2020-02-04
Return generators from armory.data


seanpmorgan:
For the 0.1 release we made it so all datasets are loaded in memory as to align with the art.utils datasets:
https://github.com/IBM/adversarial-robustness-toolbox/blob/master/art/utils.py#L471-L593
and their examples which use numpy arrays.

Going forward we'll want to return generators for large datasets though. `tfds.as_numpy` will return a NumPy generator so this is pretty straight forward, but we'll want to make sure the user interface is intuitive

PR #182, opened 2020-02-20
Imagenette dataset added, included in download_all_data

ng390:
Imagenette dataset added, included in download_all_data. Note that this requires an update to the docker images as it introduces a dependency to Pillow.

### DIFF
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -56,7 +56,13 @@ def _generator_from_tfds(
     epochs: int,
     dataset_dir: str,
     preprocessing_fn: Callable,
+    as_supervised: bool = True,
+    supervised_xy_keys=None,
 ):
+    """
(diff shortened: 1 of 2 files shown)


ng390:
Note that the CI tests will fail because the CI script will pull an older docker image without Pillow. Blocked from passing CI tests until the new Docker image is published.

davidslater:
@ng390 Can you rebase off of master? We updated the docker container to 0.4.0, and that includes Pillow, so this should simplify your merge.

ng390:
The Imagenette dataset is too big to fit in memory on the CI virtual machine. Once dataset generators are incorporated will be able to CI test. (Issue #104)

davidslater:
Please rebase off of master - should fix a number of merge issues.

davidslater:
Use TFDS ImageNette instead.

PR #195, opened 2020-02-21
German traffic sign dataset

ng390:
Add German traffic sign dataset.

### DIFF
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -2,6 +2,7 @@
 Download and preprocess common datasets.
 """
 
+import csv
 import logging
 import os
 import zipfile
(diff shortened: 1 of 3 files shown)


armory/data/datasets.py
```diff
@@ -237,11 +238,98 @@ def _parse(serialized_example):
...
+            # loop over all images in current annotations file
+            for row in gtReader:
+                try:
+                    tmp = Image.open(os.path.join(prefix, row[0]))
+                    # First column is filename
+                except IOError as e:
+                    raise IOError(f"Could not open image with PIL. {e}")
+                im_list.append(np.array(tmp))
+                tmp.close()
+                label_list.append(int(row[7]))  # the 8th column is the label
```
seanpmorgan: This dataset has annotations for bounding boxes. Seeing as we don't support this (yet) I'm fine with leaving it as a classification label but down the line we'll need to modify this.

armory/data/datasets.py
```diff
@@ -237,11 +238,98 @@ def _parse(serialized_example):
...
+
+            # Extract and clean up
+            with zipfile.ZipFile(zip_filepath, "r") as zip_ref:
+                zip_ref.extractall(dir)
+            os.remove(zip_filepath)
+
+    train_images, train_labels = [], []
+    test_images, test_labels = [], []
+
+    for c in range(0, 43):
```
seanpmorgan: This dataset is pinned to a specific hash so this is fine. We may want to look into checking the class labels in the future though.

PR #205, opened 2020-02-27
Add external repos to PYTHONPATH in the container

seanpmorgan:


armory/eval/evaluator.py
```diff
@@ -46,6 +46,9 @@ def __init__(self, config_path: str, container_config_name="eval-config.json"):
 
         if self.config["sysconfig"].get("external_github_repo", None):
             self._download_external()
+            self.extra_env_vars.update(
+                {"PYTHONPATH": self.docker_paths.dataset_dir + "/external_repos"}
+            )
```
davidslater: I don't know if this is taking effect. In the container:
```
I have no name!@476fd9027ecc:/workspace$ echo $PYTHONPATH

I have no name!@476fd9027ecc:/workspace$ 
```

seanpmorgan: Running with a config that has an external repo?

davidslater: Oh, right.

Issue #209, opened 2020-02-28
Standardize dataset directory structure


davidslater:
Ideally, the name of the dataset subdir should match the name of the function call from data.datasets
datasets.german_traffic_sign() --> datasets/german_traffic_sign/

Issue #171, opened 2020-02-19
Add a model cache to configuration


seanpmorgan:
Currently we mount a keras directory so that cached weights can be re-used. Going forward we should replace this with a model directory so cached weights can be re-used across frameworks.

seanpmorgan:
Should be handled by #263

PR #221, opened 2020-03-03
Standardize dataset directory structure (Issue #209)

ng390:
Closes Issue #209. Note that users who already have downloaded external datasets may have additional downloads given the new paths.

### DIFF
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -2,8 +2,8 @@
 Download and preprocess common datasets.
 Each standard dataset resides in its own subdirectory under dataset_dir based
 upon the name of the function in the datasets file.
-The download subdirectory under datset_dir is reserved for caching.
-The private subdirectory under dataset_dir is reserved for private datasets.
+The 'download' subdirectory under dataset_dir is reserved for caching.
+The 'private' subdirectory under dataset_dir is reserved for private datasets.
 """
 
 import csv
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -1,10 +1,11 @@
 """
 Download and preprocess common datasets.
-Each standard dataset resides in its own subdirectory under dataset_dir based
+Each standard dataset resides in its own subdirectory under <dataset_dir> based
 upon the name of the function in the datasets file. For example, the cifar10
-data is found at 'dataset_dir/cifar10'
-The 'download' subdirectory under dataset_dir is reserved for caching.
-The 'private' subdirectory under dataset_dir is reserved for private datasets.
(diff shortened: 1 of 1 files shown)


davidslater:
These changes look good. A couple of questions.

I have a couple of other folders in my datasets directory: "downloads" and "pytorch". Do you have these? And are they primarily for downloading/saving model weights, or are they dataset related?

ng390:
> These changes look good. A couple of questions.
> 
> I have a couple of other folders in my datasets directory: "downloads" and "pytorch". Do you have these? And are they primarily for downloading/saving model weights, or are they dataset related?

I don't currently have a pytorch directory, but it looks like it is cached model weights based on the source code. The downloads directory looks like it stores cached information from the in-memory datasets CIFAR and MNIST.

ng390:
As part of this issue should move the downloads directory into the appropriate subdirectories. The cached weights may be part of Issue #171?

davidslater:
Okay, then it should be solved with a different PR.

Can you add some material to the datasets.py docstring stating that:
1) each standard dataset resides in its own subdir under `dataset_dir` based on the name of the function in the datsets file
2) the `download` subdir is reserved for caching
3) the `private` subdir is reserved for private datasets

davidslater:
For `download`, we may not have that much control over where it goes - I think it's a TensorFlow thing - so I would probably leave it where it's at.

armory/data/datasets.py
```diff
@@ -1,5 +1,9 @@
 """
 Download and preprocess common datasets.
+Each standard dataset resides in its own subdirectory under dataset_dir based
+upon the name of the function in the datasets file.
+The download subdirectory under datset_dir is reserved for caching.
```
davidslater: nit: datset_dir is misspelled. 

Also, add quotes or ticks around the specific subdirectory names. 

Add an example to the standard dataset, (e.g., "cifar10")

armory/data/datasets.py
```diff
@@ -1,5 +1,10 @@
 """
 Download and preprocess common datasets.
+Each standard dataset resides in its own subdirectory under dataset_dir based
+upon the name of the function in the datasets file. For example, the cifar10
+data is found at 'dataset_dir/cifar10'
+The 'download' subdirectory under dataset_dir is reserved for caching.
+The 'private' subdirectory under dataset_dir is reserved for private datasets.
```
davidslater: Sorry, one more! (I think that this is the last one.) "dataset_dir" is not an actual directory name, but a placeholder for a key in the armory config. I would replace every instance of dataset_dir with <dataset_dir> to make it clear that it's not a string literal.

PR #228, opened 2020-03-05
updated dependencies

davidslater:
fixes #227 

We will need to update the dockerfile on docker hub, though.

Currently testing locally to ensure all dependencies are resolved properly.

PR #229, opened 2020-03-05
Ucf101

davidslater:
Add ucf101. Blocked on dependencies #228


davidslater:
TFDS takes a REALLY long time to set up this dataset (still waiting on galileo). I think that a lot of this is caused by the changing the video dimensions to 256x256. Since we're planning to look at the original aspect ratios, 320x240, it might make sense to create our own derivative dataset that just generates the aspect and split of interest.

davidslater:
This one is ready for review. Note that it takes a LONG time to extract from source (like 2 hours).

If you copy from here on galileo, it should speed that up:
`/home/davidslater/.armory/datasets/ucf101/ucf101_1`

PR #232, opened 2020-03-09
Introduce generators for datasets

seanpmorgan:
Part of #104 

### Changes:
* Adds an ArmoryDataGenerator which subclasses ART DataGenerator so it works in `.fit_generator()`
* Switches mnist and cifar10 to use generator
* Properly utilize `clip_values` for keras_mnist and keras_cifar baseline models

### Todo in future PRs (will create issues once we're in agreement):
* Fix download_all when using generators
* Make digit/imagenet_adv/TrafficSign return ArmoryDataGenerators
* Update examples on a 0.6 branch of armory-examples
* clip_values for all models
* Enforce batch_size in dataset configs


armory/data/datasets.py
```diff
@@ -20,70 +21,98 @@
...
-def _in_memory_dataset_tfds(
-    dataset_name: str, dataset_dir: str, preprocessing_fn: Callable,
-) -> (np.ndarray, np.ndarray, np.ndarray, np.ndarray):
-    default_graph = tf.compat.v1.keras.backend.get_session().graph
+class ArmoryDataGenerator(DataGenerator):
+    def __init__(self, generator, size, batch_size, preprocessing_fn=None):
+        super().__init__(size, batch_size)
+        self.preprocessing_fn = preprocessing_fn
+        self.generator = generator
+        self.total_iterations = size // batch_size
```
davidslater: Do we know how tfds handles partial batches at the end of the epoch/dataset?

armory/data/datasets.py
```diff
@@ -20,70 +21,98 @@
...
+    default_graph = tf.compat.v1.keras.backend.get_session().graph
 
-    test_x, test_y = tfds.load(
+    ds, ds_info = tfds.load(
         dataset_name,
-        batch_size=-1,
-        split="test",
+        batch_size=batch_size,
+        split=split_type,
         as_supervised=True,
```
davidslater: Not all of our datasets allow as_supervised=True

armory/data/datasets.py
```diff
@@ -20,70 +21,98 @@
...
         data_dir=dataset_dir,
+        with_info=True,
     )
 
-    train_x = tfds.as_numpy(train_x, graph=default_graph)
-    train_y = tfds.as_numpy(train_y, graph=default_graph)
-    test_x = tfds.as_numpy(test_x, graph=default_graph)
-    test_y = tfds.as_numpy(test_y, graph=default_graph)
+    ds = ds.repeat(epochs)
+    ds = ds.shuffle(batch_size * 10)
```
davidslater: Should we consider random seed setting for replicability? (Perhaps later)

armory/data/datasets.py
```diff
@@ -20,70 +21,98 @@
...
+) -> ArmoryDataGenerator:
     """
     Handwritten digits dataset:
         http://yann.lecun.com/exdb/mnist/
-
-    returns:
-        train_x, train_y, test_x, test_y
     """
     if not dataset_dir:
         dataset_dir = paths.docker().dataset_dir
```
davidslater: Recommend moving this to the ArmoryDataGenerator class init instead of replicating these lines in every function.

armory/data/datasets.py
seanpmorgan: Yeah I think we need to add a test for this and we can go from there. Looking into it I'm confused as to the default TF2 behavior for shuffle:

https://github.com/tensorflow/tensorflow/issues/31706#issuecomment-522157766

armory/data/datasets.py
seanpmorgan: Done. Thanks!

armory/data/datasets.py
seanpmorgan: Actually we need to make sure TF1 and TF2 are using the same type of shuffle (again probably best in another issue/pr)

armory/data/datasets.py
seanpmorgan: Ah thats a good catch. Can we change this when we add dataset that requires this (I think resisc45?)

It's a lot easier to analyze/work through when we have a dataset we can test with.

armory/data/datasets.py
seanpmorgan: Yeah remainders would have been returned. Updated code PTAL

armory/data/datasets.py
seanpmorgan: Per discussion it will be hard to achieve determinism anyway due to model initialization. Likely not an issue

davidslater:
I like the future PRs points. A couple of comments:
* clip_values for all models - we'll need to be careful about using this and ensuring that we provide the correct range of values
* download_all should probably move to scenario-specific processes, which may require an updated __main__ mechanism.

armory/data/datasets.py
seanpmorgan: Reverted to make the remainders still returned (needed for testing and shouldn't affect training).  Kept the new method of batching though as its more configurable in the future.

Issue #224, opened 2020-03-03
Add RESISC-45 Dataset


davidslater:
Need to download and put on s3

PR #234, opened 2020-03-09
Resisc45 generator

davidslater:
rebased off master, using generator framework

### DIFF
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -68,8 +68,8 @@ def _generator_from_tfds(
     dataset_dir: str,
     preprocessing_fn: Callable,
 ):
-    if not dataset_dir:
-        dataset_dir = paths.docker().dataset_dir
+    if dataset_dir is None:
+        dataset_dir = DEFAULT_DATASET_DIR
 
     default_graph = tf.compat.v1.keras.backend.get_session().graph
 
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -373,8 +373,12 @@ def german_traffic_sign(
 
 
 def resisc45(
-    preprocessing_fn: Callable = None, dataset_dir: str = None,
-):
+    split_type: str,
+    epochs: int,
+    batch_size: int,
+    dataset_dir: str = None,
+    preprocessing_fn: Callable = None,
+) -> ArmoryDataGenerator:
     """
     REmote Sensing Image Scene Classification (RESISC) dataset
         http://http://www.escience.cn/people/JunweiHan/NWPU-RESISC45.html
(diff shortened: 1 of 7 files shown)


davidslater:
fixes #224

armory/data/datasets.py
```diff
@@ -21,11 +21,29 @@
 from art.data_generators import DataGenerator
 from armory.data.utils import curl, download_file_from_s3
 from armory import paths
+from armory.data.resisc45 import resisc45_split  # noqa: F401
```
seanpmorgan: Could we move this inside of the resisc45 function? Or will that break tfds import hooking?

armory/data/datasets.py
```diff
@@ -21,11 +21,29 @@
...
+    global DEFAULT_DATASET_DIR
+    DEFAULT_DATASET_DIR = paths.host().dataset_dir
+
+
+def use_docker_paths():
+    global DEFAULT_DATASET_DIR
+    DEFAULT_DATASET_DIR = paths.docker().dataset_dir
+
+
+use_docker_paths()
```
seanpmorgan: Not following why these are needed? `use_host_paths()` is not a configurable option so can we remove these globals/functions and just call the paths.docker() which already caches the values so it won't be loaded many times?

armory/data/resisc45/resisc45_dataset_partition.py
```diff
@@ -0,0 +1,84 @@
...
+                split = train
+            elif i <= 600:
+                split = validation
+            else:
+                split = test
+            filename = f"{label}_{i:03}.jpg"
+            source = os.path.join(rootdir, full, label, filename)
+            target = os.path.join(rootdir, split, label, filename)
+            print(f"Copying from {source} to {target}")
+            shutil.copyfile(source, target)
```
seanpmorgan: Copies from source to target, but do we need to clean-up and remove the source files?

tests/test_dataset.py
```diff
@@ -93,6 +93,24 @@ def test_german_traffic_sign(self):
                 self.assertTrue(25 <= x.shape[0] <= 232)
                 self.assertTrue(25 <= x.shape[1] <= 266)
 
+    def test_resisc45(self):
+        for split, size in [("train", 22500), ("validation", 4500), ("test", 4500)]:
```
seanpmorgan: So adding this added 8min to test time since it has to be downloaded. Perhaps we add a SkipTest if the dataset isn't available? That way we could test locally if we want. Part of a solution for #238

armory/data/resisc45/resisc45_dataset_partition.py
davidslater: Add to docstring at the top

armory/data/datasets.py
davidslater: Removed.

tests/test_dataset.py
davidslater: Added a skip

davidslater:
Should be ready for re-review.

PR #247, opened 2020-03-10
Docker test rebase

davidslater:
Rebased off master with fixed test call.

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -28,7 +28,8 @@ jobs:
           pip install -r requirements.txt
           version=$(python -m armory --version)
           bash docker/build-dev-minimal.sh
-          docker run twosixarmory/tf1:${version} pytest -s --disable-warnings /armory_dev --ignore=tests/test_docker_management.py
+          docker run twosixarmory/tf1:${version} bash -c \
+          "cd /armory_dev; pytest -s --disable-warnings --ignore=tests/test_docker_management.py"
   flake8-test:
     name: Flake8
     runs-on: ubuntu-18.04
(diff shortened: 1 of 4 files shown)


ng390:
Should we update the docs in contrib to reflect running pytest through the Docker containers?

davidslater:
We should definitely document. However, I think we may need more discussion on how best to test everything framework related. My current thought is two suites of tests - one outside the docker container, which would run first, and a second suite that runs inside the docker container.

.github/workflows/ci_test.yml
```diff
@@ -26,9 +26,10 @@ jobs:
       - name: Run python tests
         run: |
           pip install -r requirements.txt
+          version=$(python -m armory --version)
           bash docker/build-dev-minimal.sh
-          pip install -r test-requirements.txt
-          pytest -s --disable-warnings
+          docker run twosixarmory/tf1:${version} bash -c \
+          "cd /armory_dev; pytest -s --disable-warnings --ignore=tests/test_docker_management.py"
```
seanpmorgan: nit: 
```suggestion
          docker run -w /armory_dev twosixarmory/tf1:${version} bash \
          pytest -s --disable-warnings --ignore=tests/test_docker_management.py
```

davidslater: Is the "-w" run from a specific directory? I looked for a flag that did that, but the docker docs can be hard to parse.

seanpmorgan: Yup, agree so many docs its confusing:
https://docs.docker.com/engine/reference/commandline/run/#options

davidslater: Modified.

davidslater:
@ng390 updated contributing

PR #249, opened 2020-03-10
Imagenet adv generator

### LINKED ISSUE #241: Make imagenet_adversarial dataset return an ArmoryDataGenerator

seanpmorgan:
Closes #241

### DIFF
--- a/armory/baseline_models/keras/keras_resnet50.py
+++ b/armory/baseline_models/keras/keras_resnet50.py
@@ -37,7 +37,7 @@ def get_art_model(model_kwargs, wrapper_kwargs):
                     0.0 - IMAGENET_MEANS[0],
                     0.0 - IMAGENET_MEANS[1],
                     0.0 - IMAGENET_MEANS[2],
-                    ]
+                ]
             ),
             np.array(
                 [
--- a/armory/baseline_models/keras/keras_resnet50.py
+++ b/armory/baseline_models/keras/keras_resnet50.py
@@ -31,12 +31,15 @@ def get_art_model(model_kwargs, wrapper_kwargs):
     model = ResNet50(**model_kwargs)
     wrapped_model = KerasClassifier(
         model,
-        clip_values=np.array(
-            [
-                255.0 - IMAGENET_MEANS[0],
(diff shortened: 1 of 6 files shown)


armory/baseline_models/keras/keras_resnet50.py
```diff
@@ -1,7 +1,29 @@
...
+    output = []
+    for i in range(x.shape[0]):
+        im_raw = image.array_to_img(x[i])
+        im = image.img_to_array(im_raw.resize(shape))
+        output.append(im)
+    output = preprocess_input_resnet50(np.array(output))
+    return output
 
 
 def get_art_model(model_kwargs, wrapper_kwargs):
```
seanpmorgan: What should we set the clip_values to here? There is a different upper-bound for each channel.

davidslater: It's really hard for me to say. 
https://github.com/keras-team/keras-applications/blob/master/keras_applications/imagenet_utils.py
^ has a lot of different processing paths.

It depends on what the original model was trained on. They are zero-centered on ImageNet mean. Are they also scaled? Ideally, we would take the largest inputs from the original domain for each channel ([0, 255], I think?), then apply the preprocessing to get the range for each channel. If the per-channel domains are different, then we'll need to pass in a tuple of np.arrays to the clip_values kwarg.

seanpmorgan: So I took the preprocessing functions for each model and pulled them out of keras.applications to limit confusion. This particular model uses:

I didn't know clip_values could be set per dimension, [but it looks as though they can](https://github.com/IBM/adversarial-robustness-toolbox/blob/master/art/classifiers/classifier.py#L47-L50). PTAL at the changes.

davidslater: I'll need to play around with the changes to ensure this is what ART expects, but this looks good.

davidslater: Yes, this looks good. 

ART uses `np.clip` under the hood, so it broadcasts the clip values.

PR #251, opened 2020-03-11
Update to TFDS for Imagenette

### LINKED ISSUE #168: Add Imagenette dataset

ng390:
Closes #168.

### DIFF
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -295,6 +295,30 @@ def imagenet_adversarial(
     return clean_x, adv_x, labels
 
 
+def imagenette(
+    split_type: str,
+    epochs: int,
+    batch_size: int,
+    dataset_dir: str = None,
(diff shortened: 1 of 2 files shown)


armory/data/datasets.py
```diff
@@ -295,6 +295,30 @@ def _parse(serialized_example):
...
+    batch_size: int,
+    dataset_dir: str = None,
+    preprocessing_fn: Callable = None,
+) -> ArmoryDataGenerator:
+    """
+    Smaller subset of 10 classes of Imagenet
+        https://github.com/fastai/imagenette
+    """
+    if batch_size != 1:
+        raise NotImplementedError("Due to variable length input, batch_size must be 1")
```
seanpmorgan: At some point I think we'll need to add a reshape argument to standardize variable image shapes for some datasets. (audio not as simple).

PR #252, opened 2020-03-12
Docker test rebase rebase

davidslater:
There kept being merge conflicts on test/test_datasets, so I rebased. Again.

I addressed the comments, and hopefully this provides a nicer approach to splitting docker and host tests.


davidslater:
I'm not sure why the tests were failing. If this works, maybe we have another issue for improved testing with volume mapping, release tests on all docker containers, etc.

davidslater:
We will the docker build scripts, at least, for the MITRE session tomorrow.

PR #257, opened 2020-03-12
Bump version on master

seanpmorgan:
Master should be on version ahead of release

### DIFF
--- a/armory/__init__.py
+++ b/armory/__init__.py
@@ -27,7 +27,7 @@ except ImportError as e:
 
 
 # Semantic Version
-__version__ = "0.5.0-dev"
+__version__ = "0.6.0-dev"
 DEV = "-dev"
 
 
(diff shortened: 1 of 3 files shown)


davidslater:
Makes sense. We should be careful about how we number release tags, though, as "0.6.0-dev" will be greater than "0.6.0" when doing version comparisons with distutils.

Maybe we have even patch numbers with "-dev", and odd ones with releases?

seanpmorgan:
> Makes sense. We should be careful about how we number release tags, though, as "0.6.0-dev" will be greater than "0.6.0" when doing version comparisons with distutils.
> 
> Maybe we have even patch numbers with "-dev", and odd ones with releases?

Good point. Perhaps we use pkg_resources to parse:
```
from pkg_resources import parse_version
v1 = parse_version('0.6.0-dev')
v2 = parse_version('0.6.0')
v1 > v2
```

davidslater:
Yeah, I like that solution.

Issue #254, opened 2020-03-12
Separate test runners for host/pytorch/tf


seanpmorgan:
Since we need to run tests in separate docker containers (or just on host), we need to break up our CI to run these as individual test suites

PR #260, opened 2020-03-12
Restructure host tests

seanpmorgan:
Part of #254


tests/test_host/test_config_verification.py
```diff
@@ -11,11 +11,11 @@ def test_no_config(self):
 
     def test_no_evaluation(self):
         with self.assertRaisesRegex(ValueError, "Evaluation field must contain"):
-            load_config("tests/test_configs/missing_eval.json")
+            load_config("tests/data/configs/broken/missing_eval.json")
```
davidslater: Since these are host tests, can you use Pathlib to ensure that they are not OS-dependent?

tests/test_host/test_docker_management.py
```diff
@@ -1,24 +0,0 @@
-"""
```
seanpmorgan: So I removed this. The GitHub CI doesn't have docker for macos and windows doesn't run linux images. So we could create a new test, but given that this isn't the most useful test I don't think its worth it.

davidslater:
Once you merge in, can you let the MITRE folks know that we changed the paths of the resisc45 eval script and config file?

PR #261, opened 2020-03-12
Partial refactor of Librispeech

ng390:
Rename Librispeech splits and methods, refactor tests.


armory/data/librispeech/librispeech_dev_clean_split.py
```diff
@@ -24,9 +24,9 @@
 _URL = "http://www.openslr.org/12"
 _DL_URL = "https://armory-public-data.s3.us-east-2.amazonaws.com/librispeech/"
 _DL_URLS = {
-    "dev_clean_train": _DL_URL + "librispeech-dev-clean-train.tar.gz",
-    "dev_clean_val": _DL_URL + "librispeech-dev-clean-val.tar.gz",
-    "dev_clean_test": _DL_URL + "librispeech-dev-clean-test.tar.gz",
+    "train": _DL_URL + "librispeech-dev-clean-train.tar.gz",
+    "val": _DL_URL + "librispeech-dev-clean-val.tar.gz",
```
davidslater: I have been using "validation" instead of "val" as the split name.

ng390:
Good catch -- will update.

PR #265, opened 2020-03-13
Refactor librispeech rebase

ng390:
Rebase. Rename Librispeech splits and methods, refactor tests. Uses _generator_from_tfds.


tests/test_evals/fgm_attack.py
```diff
@@ -0,0 +1,107 @@
+"""
```
davidslater: this file has been moved to `tests/evals/fgm_attack.py`. I would remove this from the PR.

PR #268, opened 2020-03-13
Simplify release

seanpmorgan:


seanpmorgan:
Removed `-minimal` script and made it so framework argument can be tf1/tf2/pytorch/all
Added the docker releases back to the release CI

PR #269, opened 2020-03-13
Docker clean

### LINKED ISSUE #258: Move from distutils to pkg_resources for versioning
pkg_resources will handle the following situation correctly. distutils will not.
```
from pkg_resources import parse_version
v1 = parse_version('0.6.0-dev')
v2 = parse_version('0.6.0')
v2 > v1  # True
```

davidslater:
fixes #258 

1) It still uses distutils to check for strict versioning, but uses pkg_resources for comparisons.

This merge also does two other things:
2) uses --force-rm when creating docker images to remove intermediate containers on failed builds
3) updates armory clean to handle "-dev" versions. (When using clean on "-dev", it only removes old "-dev" images.)

### DIFF
--- a/armory/__main__.py
+++ b/armory/__main__.py
@@ -101,7 +101,9 @@ def _pull_docker_images(docker_client=None):
             docker_client.images.get(image)
         except ImageNotFound:
             if armory.is_dev():
-                raise NotImplementedError("For dev, please build locally")
+                raise ValueError(
+                    "For '-dev', please run 'docker/build.sh' locally before running armory"
+                )
             print(f"Image {image} was not found. Downloading...")
             docker_api.pull_verbose(docker_client, image)
 
--- a/armory/__main__.py
+++ b/armory/__main__.py
@@ -102,7 +102,7 @@ def _pull_docker_images(docker_client=None):
         except ImageNotFound:
             if armory.is_dev():
                 raise ValueError(
-                    "For '-dev', please run 'docker/build.sh' locally before running armory"
+                    "For '-dev', please run 'docker/build-dev.sh' locally before running armory"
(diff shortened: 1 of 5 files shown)


armory/__main__.py
```diff
@@ -101,7 +101,9 @@ def _pull_docker_images(docker_client=None):
             docker_client.images.get(image)
         except ImageNotFound:
             if armory.is_dev():
-                raise NotImplementedError("For dev, please build locally")
+                raise ValueError(
+                    "For '-dev', please run 'docker/build.sh' locally before running armory"
```
seanpmorgan: build-dev.sh?

armory/docker/images.py
```diff
@@ -19,23 +20,23 @@
 REPOSITORIES = tuple(x.split(":")[0] for x in ALL)
 
 
-def dev_version(tag):
+def parse_version(tag):
     """
     Return version.LooseVersion class for given version tag
```
seanpmorgan: pkg_resources.version.Version

armory/__main__.py
davidslater: Good ctach.

armory/docker/images.py
davidslater: For the base images - yes, I'll need to add those. I was individually targeting them - so I'll need to add those. Alternatively, I could just look over the entire `twosixarmory/` namespace.

davidslater: `pkg_resources.version.Version` does not exist (anymore?)

davidslater: found in extern

davidslater:
Should have addressed all comments.

PR #277, opened 2020-03-18
Multiple container

### LINKED ISSUE #178: Multiple Docker container write/delete collisions
Currently, the `tmp` directory (including the evaluation file) is created and deleted by each docker container. The evaluation file may also be overwritten by a second docker container running concurrently. Similar issues may exist with the output directory and the data directory (I am unsure what would happen if two docker containers attempted to download the same model or dataset simultaneously).

For multiuser systems and evaluations with multiple containers, this behavior will need to be modified.

davidslater:
Should enable multiple docker containers to run simultaneously without impacting each others`tmp` and `output` directories. Collisions can still happen on `datasets` and `models`, but those are expected/desired.

fixes #178

### DIFF
--- a/armory/eval/evaluator.py
+++ b/armory/eval/evaluator.py
@@ -2,6 +2,7 @@
 Evaluators control launching of ARMORY evaluations.
 """
 
+import datetime
 import os
 import json
 import logging
@@ -24,6 +25,40 @@ from armory import paths
 logger = logging.getLogger(__name__)
 
 
+def tmp_output_subdir(retries=10):
+    """
+    Return (<subdir name>, <tmp subdir path>, <output subdir path>)
+
+    retries - number of times to retry folder creation before returning an error
+        if retries < 0, it will retry indefinitely.
+    """
+    tries = int(retries) + 1
+    host_paths = paths.host()
+    while tries:
+        subdir = datetime.datetime.utcnow().isoformat()
+        # Use tmp_subdir for locking
+        try:
+            tmp_subdir = os.path.join(host_paths.tmp_dir, subdir)
+            os.mkdir(tmp_subdir)
+        except FileExistsError:
+            tries -= 1
+            if tries:
+                logger.warning(f"Failed to create {tmp_subdir}. Retrying...")
+            continue
+
+        try:
+            output_subdir = os.path.join(host_paths.output_dir, subdir)
+            os.mkdir(output_subdir)
+        except OSError:
+            logger.exception(
+                f"Sync failure: created {tmp_subdir} but failed on {output_subdir}. Exiting."
+            )
+            raise
+        return subdir, tmp_subdir, output_subdir
+
+    raise ValueError("Failed to create tmp and output subdirectories")
+
+
 class Evaluator(object):
     def __init__(
         self, config_path: Union[str, dict], container_config_name="eval-config.json"
@@ -43,9 +78,13 @@ class Evaluator(object):
             self.config = config_path
         else:
             raise ValueError(f"config_path {config_path} must be a str or dict")
-        self.tmp_config = os.path.join(self.host_paths.tmp_dir, container_config_name)
+        self.container_subdir, self.tmp_dir, self.output_dir = tmp_output_subdir()
+        self.tmp_config = os.path.join(self.tmp_dir, container_config_name)
+        self.external_repo_dir = paths.get_external(self.tmp_dir)
         self.docker_config_path = Path(
-            os.path.join(self.docker_paths.tmp_dir, container_config_name)
+            os.path.join(
+                self.docker_paths.tmp_dir, self.container_subdir, container_config_name
+            )
         ).as_posix()
 
         kwargs = dict(runtime="runc")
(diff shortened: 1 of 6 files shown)


armory/eval/evaluator.py
```diff
@@ -24,6 +25,42 @@
 logger = logging.getLogger(__name__)
 
 
+def tmp_output_subdir(retries=10):
```
seanpmorgan: I would prefer this be in in some docker.utils module as there's a lot to unpack in this function and I worry it distracts from the Evaluator class which will be often looked at.

armory/eval/evaluator.py
```diff
@@ -24,6 +25,42 @@
...
+    """
+    Return (<subdir name>, <tmp subdir path>, <output subdir path>)
+
+    retries - number of times to retry folder creation before returning an error
+        if retries < 0, it will retry indefinitely.
```
seanpmorgan: Could you explain a bit why we expect this to need several retries?

armory/eval/evaluator.py
```diff
@@ -24,6 +25,42 @@
...
+    host_paths = paths.host()
+    while tries:
+        subdir = datetime.datetime.utcnow().isoformat()
+        # ":" characters violate docker-py volume specifications
+        subdir = subdir.replace(":", "-")
+        # Use tmp_subdir for locking
+        try:
+            tmp_subdir = os.path.join(host_paths.tmp_dir, subdir)
+            os.mkdir(tmp_subdir)
+        except FileExistsError:
```
seanpmorgan: This is if two containers have the exact same timestamp?

armory/eval/evaluator.py
```diff
@@ -24,6 +25,42 @@
...
+            continue
+
+        try:
+            output_subdir = os.path.join(host_paths.output_dir, subdir)
+            os.mkdir(output_subdir)
+        except OSError:
+            logger.exception(
+                f"Sync failure: created {tmp_subdir} but failed on {output_subdir}. Exiting."
+            )
+            raise
```
seanpmorgan: Why are we catching an OSError here... whats the situation that occurs in? Do we need to do clean-up of the `tmp_subdir`?

tests/test_host/test_external_repo.py
```diff
@@ -9,20 +9,20 @@
...
-        shutil.rmtree(str(pathlib.Path(f"{host_paths.external_repo_dir}/{repo_name}")))
+        self.assertTrue(os.path.exists(basedir))
+        self.assertTrue(os.path.isfile(basedir / "README.md"))
+        shutil.rmtree(basedir)
+        os.remove(external_repo_dir / (repo_name + ".tar.gz"))
+        try:
+            os.rmdir(external_repo_dir)
+        except OSError:
+            # Only delete if empty
+            pass
```
seanpmorgan: Could you elaborate on why were doing this?

armory/eval/evaluator.py
davidslater: Since the subfolder name is timestamp based, there could be collisions when simultaneously launching multiple containers (e.g., on an eval platform). Now that I'm using microsecond timestamps for naming, that chance is much smaller. I figured that 10 retries was more than plentiful to ensure no collisions.

armory/eval/evaluator.py
davidslater: Makes sense.

armory/eval/evaluator.py
davidslater: Correct.

armory/eval/evaluator.py
davidslater: This may not be something that we need to worry about, as I can't think of why this would occur. I think it would be fine to not catch and then have it die.

tests/test_host/test_external_repo.py
davidslater: The removal of the external_repo_dir is necessary for overall cleanup. This essentially ensures that all that is in the external_repo_dir is the tar file and the extracted repo. If there are additional files or folders, the end assertion will fail.

armory/eval/evaluator.py
seanpmorgan: Yeah I think it's safe to remove this catch.

davidslater: Removed.

armory/eval/evaluator.py
davidslater: Moved to `armory.docker.volume_utils`

davidslater:
Ready for another look (assuming tests pass).

tests/test_host/test_external_repo.py
```diff
@@ -9,20 +9,17 @@
...
+        download_and_extract_repo(repo, external_repo_dir=external_repo_dir)
+        basedir = external_repo_dir / repo_name
 
-        shutil.rmtree(str(pathlib.Path(f"{host_paths.external_repo_dir}/{repo_name}")))
+        self.assertTrue(os.path.exists(basedir))
+        self.assertTrue(os.path.isfile(basedir / "README.md"))
+        shutil.rmtree(basedir)
+        os.remove(external_repo_dir / (repo_name + ".tar.gz"))
+        os.rmdir(external_repo_dir)
+        os.rmdir(tmp_subdir)
```
seanpmorgan: Hmmm wouldn't this work?
```suggestion
        shutil.rmtree(tmp_subdir)
```

PR #281, opened 2020-03-18
Rename model names

seanpmorgan:
I thought it would be a good idea to remove the redundant model names. If we want to keep them this way though I can modify the PR to just rename pytorch.imagenet -> pytorch.pytorch_resnet50


davidslater:
I like the renaming. The only question I have is with regard to naming the models after datasets or classic models. Or whether it should be something like `keras/imagenet_inceptionv2.py`

seanpmorgan:
> I like the renaming. The only question I have is with regard to naming the models after datasets or classic models. Or whether it should be something like `keras/imagenet_inceptionv2.py`

I think they should be classic model names if they're a known architecture. inceptionv2 could be used for data that is not imagenet. The dataset comes into play if the user wants to load a set of pretrained weights which is done outside of the model files.

PR #282, opened 2020-03-18
Update dataset docs

seanpmorgan:
Will add more info on the ArmoryDataSet generator after we move to ART1.2

Also you'll notice we need to work on our standardization of data types. Is there any reason we can't standardize on x = float32 ; y =int64 ?

### DIFF
--- a/armory/data/README.md
+++ b/armory/data/README.md
@@ -8,9 +8,9 @@ datasets to the armory repository and standardizes them across all evaluations.
 | Dataset    | Description | x_shape | x_dtype  | y_shape  | y_dtype |
 |:---------- |:----------- |:------- |:-------- |:-------- |:------- |
 | `cifar10` | CIFAR 10 classes image dataset | (N, 32, 32, 3) | uint8 | (N,) | int64 |
-| `german_traffic_sign` | German traffic sign dataset | (N, variable_width, variable_height, 3) | uint8 | (N,) | int64 |
+| `german_traffic_sign` | German traffic sign dataset | (N, variable_height, variable_width, 3) | uint8 | (N,) | int64 |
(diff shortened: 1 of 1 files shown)


davidslater:
For data types, I don't know if we want to do any additional conversions at this point. A lot of the preprocessing (resizing, chopping, etc.) techniques are likely easier without moving from uint8 to float32. And if we need to do framework-specific things, that might need to move.

Regarding y labels, we will definitely have more complex labels later with ASR, object detection, and position tracking, so I'm not sure how much we want to standardize here.

Where would you put the standardization? Is that something that would be part of the preprocessing function or before that? If part of preprocessing, I don't think we would want to change it, but if it could be earlier (like part of TFDS loading), it might make sense, depending on what it is.

seanpmorgan:
> We still load some of the datasets in memory as numpy arrays - in particular digits and german traffic signs. We should note that below.

Do we want to include that given that with https://github.com/twosixlabs/armory/issues/284 we should try to load data without in-memory holding of data?

seanpmorgan:
> For data types, I don't know if we want to do any additional conversions at this point. A lot of the preprocessing (resizing, chopping, etc.) techniques are likely easier without moving from uint8 to float32. And if we need to do framework-specific things, that might need to move.
> Regarding y labels, we will definitely have more complex labels later with ASR, object detection, and position tracking, so I'm not sure how much we want to standardize here.

I'm okay without standardizing these as long as its apparent in the documentation what return types users will be given.

> Where would you put the standardization? Is that something that would be part of the preprocessing function or before that? If part of preprocessing, I don't think we would want to change it, but if it could be earlier (like part of TFDS loading), it might make sense, depending on what it is.

I would have this done prior to the generator in the tf.data pipeline so it's efficiently done.

Issue #255, opened 2020-03-12
Resize images using tf.data pipeline


seanpmorgan:
Currently we ensure image shape through the pre-processing function specified in each baseline model file. This is slow since we're not utilizing the tf.data pipeline which will pre-fetch and run this optimally. Instead, it should be a configuration that is passed to the dataset call and ran as a map function.

Example:
https://github.com/twosixlabs/armory/blob/master/armory/baseline_models/keras/keras_inception_resnet_v2.py#L14

seanpmorgan:
This will also allow us to remove the NotImplemented batch_size > 1 for imagnette and ucf101:

https://github.com/twosixlabs/armory/blob/master/armory/data/datasets.py#L568
https://github.com/twosixlabs/armory/blob/master/armory/data/datasets.py#L377

davidslater:
We should also think about resampling operations for audio.

davidslater:
We would like to move the preprocessing functions inside of the model boundaries, and likely remove the standard preprocessing component from the dataset pipeline.

PR #283, opened 2020-03-18
update datasets for variable length input

davidslater:
This enables datasets with variable size inputs (ucf101, librispeech, imagenette, german_traffic_sign, digits) to have batch_size > 1. This is handled differently for datasets backed by numpy arrays (in memory) and for those backed by TFDS.

The output batches are 1D np arrays of dtype object. In order to work with most systems, these will need to be modified by a preprocessing function prior to training on most models. Since this could involve a wide variety of functionality (clipping, zero-padding, masking), I think those should be handled more directly by users.

### DIFF
--- a/tests/test_docker/test_dataset.py
+++ b/tests/test_docker/test_dataset.py
@@ -190,7 +190,7 @@ class DatasetTest(unittest.TestCase):
             self.skipTest("resisc45_split dataset not locally available.")
 
         for split, size in [("train", 22500), ("validation", 4500), ("test", 4500)]:
-            batch_size = 16
+            batch_size = 15
             epochs = 1
             test_dataset = datasets.resisc45(
                 split_type=split,
--- a/tests/test_docker/test_dataset.py
+++ b/tests/test_docker/test_dataset.py
@@ -175,7 +175,10 @@ class DatasetTest(unittest.TestCase):
             )
             self.assertEqual(train_dataset.size, size)
             self.assertEqual(train_dataset.batch_size, batch_size)
-            self.assertEqual(train_dataset.total_iterations, size)
+            self.assertEqual(
+                train_dataset.total_iterations,
+                size // batch_size + bool(size % batch_size),
+            )
 
             x, y = train_dataset.get_batch()
             self.assertEqual(x.shape[0], 1)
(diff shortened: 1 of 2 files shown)


seanpmorgan:
How do we feel about removing imagenette and german_traffic_sign from variable length processing? In practice I think #255 is a better solution for those datasets. 

The audio and video can be stacked numpy arrays, though this isn't efficient as these calls are made for on batch (during training this could be substantially limiting).

davidslater:
My main concern is that using the resize functions will stretch (e.g., from rectangular to square in imagenette) and/or resample images, which is not what all models will want. Some may want the original images in their original aspect ratio and then mask or zero-pad or something like that.

tests/test_docker/test_dataset.py
```diff
@@ -190,7 +190,7 @@ def test_resisc45(self):
             self.skipTest("resisc45_split dataset not locally available.")
 
         for split, size in [("train", 22500), ("validation", 4500), ("test", 4500)]:
-            batch_size = 16
+            batch_size = 15
```
seanpmorgan: What is this change trying to test? It is skipped in the CI anyways so likely not the best place to test

seanpmorgan:
> My main concern is that using the resize functions will stretch (e.g., from rectangular to square in imagenette) and/or resample images, which is not what all models will want. Some may want the original images in their original aspect ratio and then mask or zero-pad or something like that.

That's a good point and we should support that.

tests/test_docker/test_dataset.py
davidslater: `self.assertEqual(test_dataset.total_iterations, size // batch_size)` below will actually fail due to a partial batch at the end when running with batch_size = 16. A different change would have added one to the assertEqual.

Note that I changed the total_iterations in ArmoryDataGenerator:
```
self.total_iterations = size // batch_size + bool(size % batch_size)
```

davidslater:
I'll add the digits test.

tests/test_docker/test_dataset.py
seanpmorgan: Gotcha, could you modify that all tests in the dataset_test file:
`self.assertEqual(test_dataset.total_iterations, size // batch_size + bool(size % batch_size)`

Would prefer it's explicit what we're trying to test instead of depending on the batch_size vs. total size ratio.

davidslater: Yes, I can do that.

davidslater: Only needed to do it one other place. The remaining ones are simple assertions like 
`self.assertEqual(test_dataset.total_iterations, 100)`, e.g., for MNIST.

davidslater:
Okay, ready for another review.

PR #285, opened 2020-03-18
Ds tfrecord cache

### LINKED ISSUE #270: Update dataset downloading to pull already-extracted/built TFRecords
This will require updating s3 buckets and the process for downloading/extracting them.

This should greatly reduce the overhead associated with the process.

However, the full process can still be used to generate them.

ng390:
Closes #270. Uses same hash/URL format as TFDS.

### DIFF
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -565,6 +565,9 @@ def librispeech_dev_clean(
         beam_options=beam.options.pipeline_options.PipelineOptions(flags=flags)
     )
 
+    if not dataset_dir:
+        dataset_dir = paths.docker().dataset_dir
+
     if cache_dataset and not os.path.isdir(
         os.path.join(dataset_dir, "librispeech_dev_clean_split", "plain_text", "1.1.0")
     ):
@@ -631,6 +634,9 @@ def ucf101(
     UCF 101 Action Recognition Dataset
         https://www.crcv.ucf.edu/data/UCF101.php
     """
+
+    if not dataset_dir:
+        dataset_dir = paths.docker().dataset_dir
 
     if cache_dataset and not os.path.isdir(
         os.path.join(dataset_dir, "ucf101", "ucf101_1", "2.0.0")
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -358,7 +358,7 @@ def imagenet_adversarial(
 
     ProjectedGradientDescent
         Iterations = 10
-        Max pertibation epsilon = 8
+        Max perturbation epsilon = 8
         Attack step size = 2
         Targeted = True
 
@@ -565,17 +565,11 @@ def librispeech_dev_clean(
         beam_options=beam.options.pipeline_options.PipelineOptions(flags=flags)
     )
 
-    if not dataset_dir:
-        dataset_dir = paths.docker().dataset_dir
-
-    if cache_dataset and not os.path.isdir(
-        os.path.join(dataset_dir, "librispeech_dev_clean_split", "plain_text", "1.1.0")
-    ):
-        download_verify_dataset_cache(
-            dataset_dir=dataset_dir,
-            checksum_file=os.path.join(
-                CACHED_CHECKSUMS_DIR, "librispeech_dev_clean_split.txt"
-            ),
+    if cache_dataset:
+        _cache_dataset(
+            dataset_dir,
+            name="librispeech_dev_clean_split",
+            subpath=os.path.join("plain_text", "1.1.0"),
         )
 
     return _generator_from_tfds(
@@ -596,6 +590,7 @@ def resisc45(
     batch_size: int,
     dataset_dir: str = None,
     preprocessing_fn: Callable = None,
+    cache_dataset: bool = True,
 ) -> ArmoryDataGenerator:
     """
     REmote Sensing Image Scene Classification (RESISC) dataset
@@ -612,6 +607,11 @@ def resisc45(
 
     split_type - one of ("train", "validation", "test")
     """
+    if cache_dataset:
+        _cache_dataset(
+            dataset_dir, name="resisc45_split", subpath="3.0.0",
+        )
+
     return _generator_from_tfds(
         "resisc45_split:3.0.0",
         split_type=split_type,
@@ -635,15 +635,9 @@ def ucf101(
         https://www.crcv.ucf.edu/data/UCF101.php
     """
 
-    if not dataset_dir:
-        dataset_dir = paths.docker().dataset_dir
-
-    if cache_dataset and not os.path.isdir(
-        os.path.join(dataset_dir, "ucf101", "ucf101_1", "2.0.0")
-    ):
-        download_verify_dataset_cache(
-            dataset_dir=dataset_dir,
-            checksum_file=os.path.join(CACHED_CHECKSUMS_DIR, "ucf101.txt"),
+    if cache_dataset:
+        _cache_dataset(
+            dataset_dir, name="ucf101", subpath=os.path.join("ucf101_1", "2.0.0"),
         )
 
     return _generator_from_tfds(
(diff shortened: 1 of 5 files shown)


armory/data/utils.py
```diff
@@ -72,3 +73,15 @@ def verify_sha256(filepath: str, hash_value: str, block_size: int = 4096):
     value = sha256(filepath, block_size=block_size)
     if value != hash_value:
         raise ValueError(f"sha256 hash of {filepath}: {value} != {hash_value}")
+
+
+def download_verify_dataset_cache(dataset_dir, checksum_file):
+    tmp_tar_fn = "tmp.tar.gz"
```
davidslater: recommend renaming to `tmp_tar_filename`, as I interpreted `fn` as "function", and was confused.

armory/data/utils.py
```diff
@@ -72,3 +73,15 @@ def verify_sha256(filepath: str, hash_value: str, block_size: int = 4096):
...
+    with open(checksum_file, "r") as fh:
+        url, _, hash = fh.readline().strip().split()
```
davidslater: The middle piece is file length, and would be a relatively quick check if something failed.

davidslater: Maybe create a quick "verify_size" function as well?

armory/data/utils.py
```diff
@@ -72,3 +73,15 @@ def verify_sha256(filepath: str, hash_value: str, block_size: int = 4096):
...
+    curl(url, dataset_dir, tmp_tar_fn)
+    tar_filepath = os.path.join(dataset_dir, tmp_tar_fn)
+    verify_sha256(tar_filepath, hash)
+    with tarfile.open(tar_filepath, "r:gz") as tar_ref:
+        tar_ref.extractall(dataset_dir)
```
davidslater: Instead of extracting into the target location, can you extract into a tmp directory and then move to the target location if the extraction is successful? That will prevent weird intermediate states where the target directory exists but is not properly or fully extracted. (This also mirrors how TDFS works.)

armory/data/datasets.py
```diff
@@ -496,6 +498,19 @@ def librispeech_dev_clean(
...
+    if cache_dataset and not os.path.isdir(
+        os.path.join(dataset_dir, "librispeech_dev_clean_split", "plain_text", "1.1.0")
+    ):
+        download_verify_dataset_cache(
+            dataset_dir=dataset_dir,
+            checksum_file=os.path.join(
+                CACHED_CHECKSUMS_DIR, "librispeech_dev_clean_split.txt"
+            ),
+        )
+
```
davidslater: Instead of duplicating this for every dataset of interest, can you create a single function to handle this? Something that you could call from here like:
```python
if cache_dataset:
    _cache_dataset_function(dataset_dir, name="librispeech_dev_clean_split", subpath=os.path.join("plain_text", "1.1.0"))
```

armory/data/datasets.py
```diff
@@ -550,6 +565,7 @@ def ucf101(
     batch_size: int = 1,
     dataset_dir: str = None,
     preprocessing_fn: Callable = None,
+    cache_dataset: bool = True,
 ) -> ArmoryDataGenerator:
     """
     UCF 101 Action Recognition Dataset
```
davidslater: Are UCF and Librispeech the only datasets we want to cache, or would other ones be helpful as well?

armory/data/utils.py
```diff
@@ -72,3 +73,15 @@ def verify_sha256(filepath: str, hash_value: str, block_size: int = 4096):
...
+
+
+def download_verify_dataset_cache(dataset_dir, checksum_file):
+    tmp_tar_fn = "tmp.tar.gz"
+    with open(checksum_file, "r") as fh:
+        url, _, hash = fh.readline().strip().split()
+    curl(url, dataset_dir, tmp_tar_fn)
+    tar_filepath = os.path.join(dataset_dir, tmp_tar_fn)
+    verify_sha256(tar_filepath, hash)
+    with tarfile.open(tar_filepath, "r:gz") as tar_ref:
```
davidslater: Is there a way to make the extraction more verbose (or to at least log a simple info thing for verifying the sha and extracting the tarball)?

armory/data/utils.py
davidslater: Instead of a single file with a constant name that gets saved to the base directory "datasets", how about dataset-specific names that get saved to "datasets/cache/", such as the actual url filename (ucf101.tar.gz)? Extraction can happen in the cache folder, and then the pertinent piece can get moved over to "datasets/<dataset_name>"

armory/data/datasets.py
davidslater: resisc45 took me 10+ minutes to extract through TFDS, though the full dataset is only 390MB, which took me ~30 seconds to download, hash, and extract from .tar.gz.

ng390:
PTAL -- should have resolved many of the issues. Open issue on what to do if the cached download fails -- right now it falls back to processing the input.

davidslater:
Falling back to normal processing sounds fine.

armory/data/utils.py
```diff
@@ -72,3 +76,54 @@ def verify_sha256(filepath: str, hash_value: str, block_size: int = 4096):
...
+
+def download_verify_dataset_cache(dataset_dir, checksum_file):
+    with open(checksum_file, "r") as fh:
+        url, file_length, hash = fh.readline().strip().split()
+    # download
+    cache_dir = os.path.join(dataset_dir, "cache")
+    os.makedirs(cache_dir, exist_ok=True)
+    tar_filepath = os.path.join(cache_dir, os.path.basename(url))
+    if not os.path.exists(tar_filepath):
+        curl(url, dataset_dir, tar_filepath)
```
davidslater: Recommend adding a try / except block around the curl (especially for KeyboardInterrupt), and deleting the tar_filepath if it exists after a failure. I was getting weird behavior where if I Ctrl-C to get out of the download the first time, when I run it again it skips to verification, fails, removes the tar file, and then defaults to normal processing. If I Ctrl-C out of *that*, and then run again, it will try to download the cached version again.

I would also add a logging line right before the curl to let the user know you're downloading dataset X.

armory/data/utils.py
```diff
@@ -72,3 +76,54 @@ def verify_sha256(filepath: str, hash_value: str, block_size: int = 4096):
...
+        logger.info("Falling back to processing data...")
+        return
+    except tarfile.ExtractError:
+        logger.info(f"Could not extract tarfile: {tar_filepath}")
+        logger.info("Falling back to processing data...")
+        return
+
+    # move tmp_dir subdirectory to dataset_dir (tarball does not contain files at root level)
+    for directory in os.listdir(path=tmp_dir):
+        shutil.move(os.path.join(tmp_dir, directory), dataset_dir)
```
davidslater: You can run into an issue here if the top level directory exists (e.g., for a different version of the TFDS dataset), but not this specific one:
```
2020-03-19 23:42:27 5632ef0206c7 armory.data.utils[637] INFO Extracting .tfrecord files from download...
  File "/workspace/armory/data/datasets.py", line 505, in librispeech_dev_clean
    subpath=os.path.join("plain_text", "1.1.0"),
  File "/workspace/armory/data/datasets.py", line 601, in _cache_dataset
    checksum_file=os.path.join(CACHED_CHECKSUMS_DIR, name + ".txt"),
  File "/workspace/armory/data/utils.py", line 128, in download_verify_dataset_cache
    shutil.move(os.path.join(tmp_dir, directory), dataset_dir)
  File "/opt/conda/lib/python3.7/shutil.py", line 564, in move
    raise Error("Destination path '%s' already exists" % real_dst)
shutil.Error: Destination path '/armory/datasets/librispeech_dev_clean_split' already exists
```

Since you know that the full directory path doesn't exist, I would suggest using os.makedirs to make all of the folders beneath that point and then moving over the leaf directory.

davidslater: Actually,  it is probably better to just find the first directory not in common and move that over, then cleanup the starting directory.

davidslater: If you call the following method instead of shutil.move, I think that will do the trick.
```python
import os                                                                                
import shutil                                                                            
                                                                                         
def move_merge(source='/armory/datasets/cache/tmp_esfupefxcusymnia/ucf101',              
        dest='/armory/datasets'):                                                        
    dest_location = os.path.join(dest, os.path.basename(source))                         
    if not os.path.exists(dest_location):                                                
        shutil.move(source, dest)                                                        
        return                                                                           
                                                                                         
    # Move up a directory, as shutil.move will error                                     
    filepaths = [os.path.join(source, x) for x in os.listdir(source) if not x.startswith(".")]
    if len(filepaths) != 1 or not os.path.isdir(filepaths[0]):                           
        raise ValueError(f"{source} not a single branch directory. Cannot recurse.")     
    move_merge(filepaths[0], dest_location)                                              
    os.rmdir(source)
```

davidslater: [source and dest are filled in there for clarity of my comment]

PR #287, opened 2020-03-19
Clean-up test cases

seanpmorgan:


seanpmorgan:
Might have to wait for #283 since I used the +1 on iteration batch that is part of that PR

PR #289, opened 2020-03-20
External branches

seanpmorgan:


seanpmorgan:
This cifar failed download and then borking GitHub actions couldn't be more annoying. Looking forward to moving our cached datasets to S3

PR #293, opened 2020-03-20
Lower minimum cuda requirements for pytorch

### LINKED ISSUE #292: Support CUDA 9.x limited devices
We've been asked to support GPU devices without capability to run CUDA10.1. This should be feasible for the pytorch docker container since the binary can support multiple CUDA versions. Likely little performance regression by doing this. Turing cards will be unsupported but we can work through that if the need arises:
https://github.com/twosixlabs/armory/blob/master/docker/pytorch/Dockerfile#L10

seanpmorgan:
Closes #292

### DIFF
--- a/docker/pytorch/Dockerfile
+++ b/docker/pytorch/Dockerfile
@@ -7,7 +7,7 @@ FROM twosixarmory/armory:${armory_version} AS armory-pytorch-base
 RUN /opt/conda/bin/conda install tensorflow==1.15.0 \
  pytorch==1.4 \
  torchvision==0.5.0 \
- cudatoolkit=10.1 -c pytorch
+ cudatoolkit=9.2 -c pytorch
 
 ########## PyTorch 1.4 Release #################
 


davidslater:
Is this going to work okay with TFDS? I think that I would rather have a special case for 9.X limited devices, and keep the main containers on 10.X. They'll eventually need to upgrade, anyhow - I don't want our versions pinned to the performer lowest common denominator.

Could they build a derivative docker container:
```
from twosixarmory/pytorch
RUN /opt/conda/bin/conda install cudatoolkit=9.2 -c pytorch
```
and run:
```
docker build . -t twosixarmory/pytorch-cuda9:<tag>
```
When I did it from the main pytorch image, it installed cuda9 toolkit, but the underlying cuda binaries were 10.X.

Alternatively,  build from the armory base.

docker/pytorch/Dockerfile
```diff
@@ -7,7 +7,7 @@ FROM twosixarmory/armory:${armory_version} AS armory-pytorch-base
 RUN /opt/conda/bin/conda install tensorflow==1.15.0 \
  pytorch==1.4 \
  torchvision==0.5.0 \
- cudatoolkit=10.1 -c pytorch
+ cudatoolkit=9.2 -c pytorch
```
davidslater: Building this locally, I still get the underlying CUDA version to be 10:
```
(base) root@453de08928f1:/usr/local/cuda# cat /usr/local/cuda/version.txt 
CUDA Version 10.0.130
```

seanpmorgan: Check ENV for which cuda is being used in the conda environment

davidslater: Can you see what your docker container CUDA version looks like? I don't have a good sense of how to test this. Maybe we'll need to set up AWS nodes with certain binaries on them?

davidslater:
```
(base) root@d20b5c67e882:/workspace# env  | grep CUDA
CUDA_PKG_VERSION=10-0=10.0.130-1
CUDA_VERSION=10.0.130
NVIDIA_REQUIRE_CUDA=cuda>=10.0 brand=tesla,driver>=384,driver<385 brand=tesla,driver>=410,driver<411
(base) root@d20b5c67e882:/workspace# env
```

Issue #284, opened 2020-03-18
Move all datasets to TFDS model


davidslater:
This would simplify/unify a lot of our code and dataset processing.

We would need to update the following datasets:
- [x] digit (currently numpy)
- [x] imagenet_adversarial (currently custom TFRecordDataset)
- [x] german_traffic_sign (currently numpy)

davidslater:
Then datasets methods could then simply look like:
```python
def mnist(*args, **kwargs):
    """                                                                                  
    Handwritten digits dataset:                                                          
        http://yann.lecun.com/exdb/mnist/                                                
    """ 
    return _generator_from_tfds("mnist:3.0.0", *args, **kwargs)
```

seanpmorgan:
Agree this is a nice standardization. If we need to create PyTorch generators for performance reasons, this work will not be of value to that but it still cleans up our TF generators so I would certainly be in favor of this

PR #301, opened 2020-03-23
digit dataset -> tfds

davidslater:
Part of #284


davidslater:
@seanpmorgan only the MacOS test is failing here, so I think it's ready for review.

PR #305, opened 2020-03-24
German tfds

davidslater:
Part of #284


armory/data/german_traffic_sign/german_traffic_sign.py
```diff
@@ -0,0 +1,91 @@
+"""
+TensorFlow Dataset for resisc45 with train/validate/test splits
```
ng390: Comment should be updated with German Traffic sign dataset.

davidslater:
@ng390 resolved your docstring update

PR #306, opened 2020-03-24
label class update from Hiro

davidslater:


davidslater:
I haven't tested this yet, as I have trouble running the build on my own laptop.

armory/data/librispeech/librispeech_dev_clean_split.py
```diff
@@ -207,5 +251,6 @@ def _generate_librispeech_examples(directory):
                     "chapter_id": chapter_id,
                     "speech": os.path.join(path, audio_file),
                     "text": transcript,
+                    "label": speaker_id,
```
seanpmorgan: I think this needs to be str(speaker_id) so that it can be matched with the class _LABEL map

davidslater: I've made the change, and now I'm going to test it on galileo

davidslater: Testing on galileo went fine.

davidslater:
@ng390 when this merges in, can you update the s3 cache? 

I have it built on galileo at `/home/davidslater/.armory/datasets/librispeech_dev_clean_split`

ng390:
Sure, will do.

PR #313, opened 2020-03-25
Tfds digit rebase extra

davidslater:
Additionally removes the numpy generator based material and other redundant imports.


davidslater:
Part of #284

Issue #318, opened 2020-03-26
Add download progress bar to download_file_from_s3


ng390:

davidslater:
Great idea - I was thinking this when looking at the recent cache PR.

PR #319, opened 2020-03-26
Switch from curl to boto3 for downloads

### LINKED ISSUE #317: Use boto3 instead of curl for all s3 downloads

ng390:
Closes #317. Will be more user-friendly with Issue #318 resolved (separate PR).

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,7 +3,7 @@
 FROM nvidia/cuda:10.0-cudnn7-runtime-ubuntu18.04 AS armory
 
 RUN apt-get -y -qq update && \
-    apt-get install -y wget vim build-essential git curl
+    apt-get install -y wget vim build-essential git 
 
 # Install Conda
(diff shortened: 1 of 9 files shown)


davidslater:
Be sure to assign yourself to git issues when you start working on them to prevent duplication of work.

docker/Dockerfile
```diff
@@ -3,7 +3,7 @@
 FROM nvidia/cuda:10.0-cudnn7-runtime-ubuntu18.04 AS armory
 
 RUN apt-get -y -qq update && \
-    apt-get install -y wget vim build-essential git curl
+    apt-get install -y wget vim build-essential git
```
seanpmorgan: Since this is the first step in the build of our base image it'll invalidate a lot of layer re-use. I think it's okay to keep curl in the container

ng390: Updated.

PR #321, opened 2020-03-26
Pre commit json

davidslater:
I'm not sure why this isn't in our pre-commit.sh, as it is in our ci testing.

### DIFF
--- a/tools/pre-commit.sh
+++ b/tools/pre-commit.sh
@@ -11,7 +11,7 @@ then
     exit $need_format
 fi
 
-python -m tools.format_json --check
+python -m tools.format_json --check > /dev/null 2>&1
 need_format=$?
 if [ $need_format -ne 0 ]
 then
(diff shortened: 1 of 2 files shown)


tools/pre-commit.sh
```diff
@@ -12,6 +11,17 @@ then
     exit $need_format
 fi
 
+python -m tools.format_json --check > /dev/null 2>&1
+need_format=$?
+if [ $need_format -ne 0 ]
+then
+    echo "Some JSON files need to be reformatted. Please run:"
+    echo "python -m tools.format_json"
+    exit $need_format
```
seanpmorgan: Can we have this run `python -m tools.format_json`? The black pre-commit currently runs formatting if needed.

davidslater:
I made it so that it runs `format_json` if it fails.

In doing so, I had to slightly change the tool itself, so that it returns an exit code of "0" if you run `python -m tools.format_json` and it reformats files (but does not error). However, `--check` will return an exit code of "1" if it _would_ reformat the file. This update now mirrors `black` more closely.

PR #324, opened 2020-03-26
Utilize docker build kit

seanpmorgan:
Testing to see if we see a performance improvement on test cases

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -40,6 +40,8 @@ jobs:
         with:
           python-version: '3.6'
       - name: Run armory TF1 docker tests
+        env:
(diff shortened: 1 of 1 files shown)


seanpmorgan:
Shaves a minute or two off each pytest job

davidslater:
Nice! How does the buildkit work?

davidslater:
Once we get all of the datasets cached, that should also cut off a lot of time.

seanpmorgan:
> Nice! How does the buildkit work?

It's an upgraded toolkit for docker builds. Has support for remote caching, amongst other things. Also resolves the build to see what can be built in parallel.

https://github.com/moby/buildkit
https://docs.docker.com/develop/develop-images/build_enhancements/

PR #326, opened 2020-03-26
Extend tfrecord cache to all datasets

### LINKED ISSUE #294: Extend Ds tfrecord cache to all datasets
Blocked on #284

ng390:
Closes #294

### DIFF
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -412,23 +412,21 @@ def _cache_dataset(dataset_dir: str, dataset_name: str):
         )
 
 
-def _parse_dataset_name(dataset_name):
-    if ":" not in dataset_name:
-        raise ValueError("Must have version specified to load dataset.")
-    name_config_version = dataset_name.split(":")
-    if len(name_config_version) != 2:
-        raise ValueError(f"Dataset name {dataset_name} not properly formatted.")
-    name_config, version = name_config_version
-    has_config = "/" in name_config
-    if has_config:
+def _parse_dataset_name(dataset_name: str):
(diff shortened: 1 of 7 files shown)


armory/data/datasets.py
```diff
@@ -409,6 +412,26 @@ def _cache_dataset(dataset_dir: str, name: str, subpath: str):
...
+    if has_config:
+        splits = name_config.split("/")
+        if len(splits) != 2:
+            raise ValueError(f"Dataset name {dataset_name} not properly formatted.")
+        name = splits[0]
+        subpath = os.path.join(splits[1], version)
+    else:
+        name = name_config
+        subpath = version
+    return name, subpath
```
davidslater: I might suggest a cleaner way to do this as:
```python
def _parse_dataset_name(dataset_name: str):                                              
    try:                                                                                 
        name_config, version = dataset_name.split(":")                                   
        splits = name_config.split("/")                                                  
        if len(splits) > 2:                                                              
            raise ValueError                                                             
        name = splits[0]                                                                 
        config = splits[1:]                                                              
        subpath = os.path.join(*config + [version])                                      
    except ValueError:                                                                   
        raise ValueError(                                                                
            f'Dataset name "{dataset_name}" not properly formatted.\n'                   
            'Should be formatted "<name>[/<config>]:<version>", '                        
            'where "[]" indicates "/<config>" is optional.'                              
        )                                                                                
    return name, subpath
```

ng390: This looks a good bit cleaner, thanks!

PR #327, opened 2020-03-26
ArmoryDataGenerator implements iterator interface

davidslater:
Enable iteration (and enumeration) of ArmoryDataGenerator.

It's definitely counterintuitive to have a generator that you can't directly iterate on. This small change enables that. It also enables syntactic sugar which might be nice for scenarios. Thus:
```python
    cnt = 0                                                                              
    for _ in range(test_data_generator.batches_per_epoch):                               
        x, y = test_data_generator.get_batch()                                           
        test_x_adv = attack.generate(x=x)                                                
        predictions = classifier.predict(test_x_adv)                                     
        adversarial_accuracy += np.sum(np.argmax(predictions, axis=1) == y) / len(y)     
        cnt += 1
    adversarial_accuracy = adversarial_accuracy / cnt                                    
```
would become 
```python
    for cnt, (x, y) in test_data_generator:                                              
        test_x_adv = attack.generate(x=x)                                                
        predictions = classifier.predict(test_x_adv)                                     
        adversarial_accuracy += np.sum(np.argmax(predictions, axis=1) == y) / len(y)
    adversarial_accuracy = adversarial_accuracy / cnt                                    
```

### DIFF
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -81,6 +81,12 @@ class ArmoryDataGenerator(DataGenerator):
 
         return x, y
 
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        return self.get_batch()
+
 
 def _generator_from_tfds(
     dataset_name: str,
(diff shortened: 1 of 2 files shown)


davidslater:
I can add a test case.

PR #330, opened 2020-03-27
Tqdm progress bar for S3

### LINKED ISSUE #318: Add download progress bar to download_file_from_s3

ng390:
Closes #318


armory/data/utils.py
```diff
@@ -14,21 +14,31 @@
...
 from botocore.client import Config
 
+from armory.data.progress_percentage import ProgressPercentage
 
 logger = logging.getLogger(__name__)
 
 
-def download_file_from_s3(bucket_name: str, key: str, local_path: str):
+def download_file_from_s3(
+    bucket_name: str, key: str, local_path: str, progress_bar=False
```
seanpmorgan: Would prefer if we remove API bloat and just always use progress_bar. Any reason not to?

ng390: Agreed, this option seems unnecessary.

PR #332, opened 2020-03-27
Use curl with download_verify_dataset_cache

ng390:
Exact details of the bug are unclear (could not replicate locally), but the boto3 call seems to be intermittently creating segmentation faults during CI testing. Below are two example failing CI tests:

https://github.com/twosixlabs/armory/runs/540390645?check_suite_focus=true
https://github.com/twosixlabs/armory/runs/540179652?check_suite_focus=true

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,5 +74,3 @@ armory run examples/fgm_attack.json
 # Scenarios
 Armory will provide several scenarios for various data modalities and threat models. 
 More information will be added 
-
-""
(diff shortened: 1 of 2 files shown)


seanpmorgan:
So I think we should debug this a bit more. Using boto3 is the preferred method and we already had cert issues using curl. 

Which PR do we think is the one where it started? boto3 segfaulting is quite strange but I do see some reports of it.

ng390:
> So I think we should debug this a bit more. Using boto3 is the preferred method and we already had cert issues using curl.
> 
> Which PR do we think is the one where it started? boto3 segfaulting is quite strange but I do see some reports of it.

It looks like it is PR #319 / commit hash 922dcffffd9e3fb560db41e0dc99b6252462ced0 that introduced the issue.

PR #333, opened 2020-03-29
Adding video scenario with UCF101 Armory datagen

yusong-tan:


davidslater:
Superceded by other PR

PR #335, opened 2020-04-01
Raw audio with SincNet scenario

hkakitani:


armory/baseline_models/pytorch/sincnet.py
```diff
@@ -0,0 +1,148 @@
...
+
+# from https://github.com/hkakitani/SincNet
+from SincNet import dnn_models
+
+logger = logging.getLogger(__name__)
+os.environ["TORCH_HOME"] = os.path.join(paths.docker().dataset_dir, "pytorch", "models")
+
+FS = 8000
+CW_LEN = 375
+WLEN = int(FS * CW_LEN / 1000)
```
davidslater, line 18: Can you spell these out? I presume these correspond to something like sampling_rate, window_step_size, and window_length, but it's hard to understand by inspection.

PR #348, opened 2020-04-03
Utilize GPU for pytorch models if available

### LINKED ISSUE #342: Parameterize PyTorch models for CPU/GPU
PyTorch does not have a fallback for CPU if a model/optimizer is specified to run with `.cuda()` instead we need to parameterize our devices and parse the gpu options from configs.

https://pytorch.org/tutorials/beginner/pytorch_with_examples.html#pytorch-tensors

seanpmorgan:
Closes #342 

Better than having to parse a GPU config, if it's available then it'll be used. Same strategy our TF models use.

### DIFF
--- a/armory/baseline_models/pytorch/mnist.py
+++ b/armory/baseline_models/pytorch/mnist.py
@@ -46,6 +46,10 @@ def make_mnist_model(**kwargs):
 
 def get_art_model(model_kwargs, wrapper_kwargs, weights_file=None):
     model = make_mnist_model(**model_kwargs)
+
+    if torch.cuda.is_available():
+        model.cuda()
+
     if weights_file:
         saved_model_dir = paths.docker().saved_model_dir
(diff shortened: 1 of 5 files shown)


armory/baseline_models/pytorch/mnist.py
```diff
@@ -46,6 +46,10 @@ def make_mnist_model(**kwargs):
 
 def get_art_model(model_kwargs, wrapper_kwargs, weights_file=None):
     model = make_mnist_model(**model_kwargs)
+
+    if torch.cuda.is_available():
+        model.cuda()
```
davidslater, line 51: `model.cuda()` is an in place operation? I.e., we don't need to do `model = model.cuda()` ?

seanpmorgan: Yup it's inplace:
https://discuss.pytorch.org/t/solved-make-sure-that-pytorch-using-gpu-to-compute/4870/2

PR #350, opened 2020-04-03
Audio_video_poison_scenario_branch

yusong-tan:
Finished audio, video, and poison scenarios.  Applied Black, Flake8, and JSON formatting.  Tested with upstream/master changes from April 2.

PR #351, opened 2020-04-06
Poisoning scenario

ng390:
Partial split of PR #350.


ng390:
Now tested locally and with GPU.

.github/workflows/ci_test.yml
```diff
@@ -11,10 +11,6 @@ on:
     branches:
       - master
       - r*
-
-env:
-  DOCKER_BUILDKIT: 1
-
```
seanpmorgan, line 17: Hmmm don't think this file should have been modified?

armory/baseline_models/keras/keras_gtsrb_micronnet.py
```diff
@@ -0,0 +1,68 @@
+"""
```
seanpmorgan, line 1: Rename the file to `gtsrb_micronnet` keras_ is redundant since the directory

armory/baseline_models/keras/keras_gtsrb_micronnet.py
```diff
@@ -0,0 +1,68 @@
...
+        loss=tf.keras.losses.sparse_categorical_crossentropy,
+        metrics=["accuracy"],
+    )
+
+    return model
+
+
+def get_art_model(model_kwargs, wrapper_kwargs, weights=None):
+    model = make_model(**model_kwargs)
+    wrapped_model = KerasClassifier(model, clip_values=(0.0, 255.0), **wrapper_kwargs)
```
seanpmorgan, line 67: Curious as to why we don't normalize input? Can be changed later on though

armory/baseline_models/pytorch/cifar.py
```diff
@@ -46,10 +46,6 @@ def make_cifar_model(**kwargs):
 
 def get_art_model(model_kwargs, wrapper_kwargs, weights_file=None):
     model = make_cifar_model(**model_kwargs)
-
-    if torch.cuda.is_available():
-        model.cuda()
```
seanpmorgan, line 51: Reason to remove these? We do need to change this method to match the way it's device placement is done, but probably not this PR

tests/configs/pytorch/fgm_attack_test.json
```diff
@@ -1,36 +0,0 @@
-{
```
seanpmorgan, line 1: Why remove?

PR #354, opened 2020-04-07
Cifar10 scenario

davidslater:
The built-in ART Madry adversarial defense didn't implement fit_generator.

I still need to integrate that defense into the scenario.

I could use some review of the `scenarios.base` module, as I think it could be a very clean way of unifying all of the scenarios together and reducing boilerplate code and strange one-offs.


armory/scenarios/base.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class Scenario(abc.ABC):
+    def __init__(self):
+        pass
+
+    def validate_config(self, config):
+        """
+        Validate the scenario config
+        """
```
seanpmorgan, line 25: What would be done in here that is different from our config file validation?

armory/scenarios/base.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def evaluate(self, config: dict):
+        """
+        Evaluate a config for robustness against attack.
+        """
+        self.validate_config(config)
+        results = self._evaluate(config)
+        if results is None:
+            logger.warning(f"{self._evaluate} returned None, not a dict")
```
seanpmorgan, line 34: Is this an expected return? Or just a warning for improper evaluation files?

armory/scenarios/base.py
```diff
@@ -0,0 +1,102 @@
...
+        "--debug",
+        dest="log_level",
+        action="store_const",
+        const=logging.DEBUG,
+        default=logging.INFO,
+        help="Debug output (logging=DEBUG)",
+    )
+    args = parser.parse_args()
+    coloredlogs.install(level=args.log_level)
+    run_config(args.config_path)
```
seanpmorgan, line 102: Can you give an example cli for how you imagine users would call this script?

armory/utils/metrics.py
```diff
@@ -0,0 +1,22 @@
...
+def categorical_accuracy(y, y_pred):
+    """
+    Return the categorical accuracy of the predictions
+    """
+    if y.shape == y_pred.shape:
+        return [bool(x) for x in list(y == y_pred)]
+    elif y.ndim + 1 == y_pred.ndim:
+        return [bool(x) for x in list(y == np.argmax(y_pred, axis=1))]
+    else:
+        raise ValueError(f"{y} and {y_pred} accuracy cannot be computed")
```
seanpmorgan, line 13: Expand on this a little please to describe why it couldn't be computed.

armory/scenarios/configs/cifar10_baseline.json
```diff
@@ -0,0 +1,50 @@
...
+        "module": "armory.baseline_models.keras.cifar",
+        "name": "get_art_model",
+        "weights_file": null,
+        "wrapper_kwargs": {}
+    },
+    "scenario": {
+        "kwargs": {},
+        "module": "armory.scenarios.cifar10",
+        "name": "Cifar10"
+    },
```
seanpmorgan, line 44: A little worried about how this will be handled for configs without a scenario. Perhaps we take this time to enforce evaluations always be ran on a scenario (new scenarios could obviously be added by endusers).

armory/scenarios/base.py
davidslater: I wanted a way to ensure evaluation files returned results. But I'm not tied to this.

armory/scenarios/base.py
davidslater: This was for any scenario-specific validation. Perhaps that can be handled when instantiating the subclasses.

armory/scenarios/base.py
davidslater: Yep, I'll add it to the docstring.

armory/scenarios/base.py
davidslater: Probably something that we would want our tests to include, instead.

armory/scenarios/base.py
davidslater: Better handled at subclass instantiation.

armory/scenarios/configs/cifar10_baseline.json
davidslater: Agreed. Will standardize when we have all of the scenarios incorporated.

davidslater:
@seanpmorgan I removed the WIP and hopefully addressed your comments. It works fine if you don't use a defense. I'm running it currently with my madry adversarial trainer; interested in any other feedback at this point.

armory/scenarios/cifar10.py
```diff
@@ -0,0 +1,97 @@
...
+        """
+        Evaluate the config and return a results dict
+        """
+
+        model_config = config["model"]
+        classifier, preprocessing_fn = load_model(model_config)
+
+        if not model_config["weights_file"]:
+            classifier.set_learning_phase(True)
+            logger.info(
```
seanpmorgan, line 32: Do we think a null weight's file is inutitive enough for users to know it will train a model. I lean toward yes, but we've seen flags like `model_status` and `pretrained` etc. so wondering if we should think about that a bit. (Not gaiting this PR though)

davidslater: I think it's intuitive enough; however, maybe we want to a different flag for whether to train it. I personally don't like the "model_status" and "pretrained" flags. A wider discussion of specific flags seems in order after we have the other branches merged.

PR #362, opened 2020-04-09
Resisc45 class

davidslater:


davidslater:
Adding `__len__` to the generator allows `tqdm` to display total size and percent complete.

armory/scenarios/resisc45.py
```diff
@@ -2,99 +2,81 @@
...
-from importlib import import_module
 
-import numpy as np
+from tqdm import tqdm
 
-from armory.utils.config_loading import load_dataset, load_model
-from armory import paths
+from armory.scenarios.base import Scenario
+from armory.utils import metrics
+from armory.utils.config_loading import load_dataset, load_model, load_attack
```
seanpmorgan, line 11: Could we add the defense loading for the scenario

davidslater: I think the defense loading requires a little more work - I don't think the CIFAR setup will work with pre/post-processing defenses. Maybe we do a pass at the end to add defenses to all scenarios?

PR #364, opened 2020-04-09
Utilize scenario base class in test cases

seanpmorgan:


tests/scenarios/demo_fgm.py
```diff
@@ -0,0 +1,100 @@
...
+
+        attack = load_attack(config["attack"], classifier)
+        test_data_generator = load_dataset(
+            config["dataset"],
+            epochs=1,
+            split_type="test",
+            preprocessing_fn=preprocessing_fn,
+        )
+
+        adversarial_accuracies, perturbations = [], []
```
seanpmorgan, line 83: Remove these since we're not adjusting or plotting?

davidslater: I would remove perturbations, and then just adjust the output dict so that it only returns benign and adversarial accuracy

seanpmorgan: Done. I also only return the means. Thoughts?

davidslater: Looks fine. The means are generated from the lists.

PR #368, opened 2020-04-09
Librispeech raw scenario

davidslater:


armory/baseline_models/pytorch/sincnet.py
```diff
@@ -138,7 +136,7 @@ def preprocessing_fn(batch):
 
 # NOTE: PyTorchClassifier expects numpy input, not torch.Tensor input
 def get_art_model(model_kwargs, wrapper_kwargs, model_weights=None):
-    model = sincnet(**model_kwargs)
+    model = sincnet(model_weights=model_weights, **model_kwargs)
```
seanpmorgan, line 139: So we unpack the `model_kwargs` but it would error. Is it safer for sincnet to accept kwargs? Or preferable to fail?

armory/scenarios/configs/librispeech_dev_clean_baseline_raw_sincnet.json
```diff
@@ -23,18 +20,24 @@
...
-        "model_kwargs": {
-            "model_file": "model_sincnet.pth",
-            "pretrained": true
+        "fit_kwargs": {
+            "nb_epochs": 20000
         },
+        "model_kwargs": {},
         "module": "armory.baseline_models.pytorch.sincnet",
         "name": "get_art_model",
+        "weights_file": "SincNet/model_sincnet.pth",
```
seanpmorgan, line 33: Will need an update on S3. Do we want this to be `pytorch/sincnet/model_sincnet`? Or prefer we work out the directory structure once they're all merged?

armory/scenarios/librispeech_dev_clean.py
```diff
@@ -0,0 +1,86 @@
...
+        logger.info(f"Loading dataset {config['dataset']['name']}...")
+        test_data_generator = load_dataset(
+            config["dataset"],
+            epochs=1,
+            split_type="test",
+            preprocessing_fn=preprocessing_fn,
+        )
+        logger.info("Running inference on benign examples...")
+
+        task_metric = metrics.categorical_accuracy
```
seanpmorgan, line 54: Do we want this to be picked up from the config `metric`? or should we remove metric from the configs and have them be coded into the scenario so they're compatible?

davidslater: It would be nice to pick it up from the config. That way you could track different (or additional metrics) as desired.

armory/scenarios/configs/librispeech_dev_clean_baseline_raw_sincnet.json
davidslater: Good catch - I'll update the structure so that it works with the present setup, and we should work out the directory structure all at once.

armory/baseline_models/pytorch/sincnet.py
davidslater: If model_kwargs is an empty dict, it will work fine. If you add kwargs into it, it will fail. I would prefer it to fail over silently accepting unknown kwargs.

armory/scenarios/configs/librispeech_dev_clean_baseline_raw_sincnet.json
davidslater: Should work now.

PR #370, opened 2020-04-09
Poisoning scenario (with corrected merge)

ng390:
Partial split of PR #350.


armory/baseline_models/keras/keras_gtsrb_micronnet.py
```diff
@@ -0,0 +1,68 @@
+"""
```
seanpmorgan, line 1: During PR for Scenario subclass we should make this `baseline_models/keras/gtsrb_micronnet.py`

armory/scenarios/gtsrb_scenario.py
```diff
@@ -0,0 +1,170 @@
...
+from importlib import import_module
+
+import numpy as np
+
+from armory.utils.config_loading import load_dataset, load_model
+from armory import paths
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+coloredlogs.install(logging.INFO)
```
seanpmorgan, line 19: During PR for Scenario subclass we should remove the logger level being set here

PR #372, opened 2020-04-09
Spectrogram scenario

seanpmorgan:


armory/data/datasets.py
```diff
@@ -44,13 +44,26 @@ class ArmoryDataGenerator(DataGenerator):
...
+        batch_size,
+        preprocessing_fn=None,
+        variable_length=False,
     ):
         super().__init__(size, batch_size)
         self.preprocessing_fn = preprocessing_fn
         self.generator = generator
+
+        self.epochs = epochs
+        self.samples_per_epoch = size
```
seanpmorgan, line 60: Renamed this to a more intuitive name

armory/data/datasets.py
```diff
@@ -88,7 +101,7 @@ def __next__(self):
         return self.get_batch()
 
     def __len__(self):
-        return self.size
+        return self.batches_per_epoch * self.epochs
```
seanpmorgan, line 104: For tqdm we want to show a total number of batches. Previously it was the size (i.e. number of samples per epoch) which doesn't match the iteration

armory/scenarios/librispeech_dev_clean_spectrogram.py
```diff
@@ -0,0 +1,171 @@
...
+
+                classifier.fit(
+                    x_seg,
+                    y,
+                    batch_size=config["dataset"]["batch_size"],
+                    nb_epochs=1,
+                    verbose=True,
+                )
+
+                if cnt % train_data_generator.batches_per_epoch == 0:
```
davidslater, line 69: Won't this do the validation at the beginning of the epoch, after the first batch? Would we actually want
```if (cnt + 1) % train_data_generator.batches_per_epoch == 0:```

File: armory/scenarios/librispeech_dev_clean_spectrogram.py [hunk 4]
```diff
@@ -0,0 +1,171 @@
...
+                # The model accepts a fixed size spectrogram, so x_trains need to
+                # be sampled.
+
+                x_seg = []
+                for xt in x:
+                    rand_t = np.random.randint(xt.shape[1] - n_tbins)
+                    x_seg.append(xt[:, rand_t : rand_t + n_tbins])
+                x_seg = np.array(x_seg)
+                x_seg = np.expand_dims(x_seg, -1)
+
```
davidslater, line 60: It seems like this should be a preprocessing function.

armory/scenarios/librispeech_dev_clean_spectrogram.py
```diff
@@ -0,0 +1,171 @@
...
+                                    xt[:, ii * n_tbins : (ii + 1) * n_tbins]
+                                )
+                                y_val_seg.append(yt)
+                        x_val_seg = np.array(x_val_seg)
+                        x_val_seg = np.expand_dims(x_val_seg, -1)
+                        y_val_seg = np.array(y_val_seg)
+
+                        y_pred = classifier.predict(x_val_seg)
+                        validation_accuracies.extend(task_metric(y_val_seg, y_pred))
+                        cnt += len(y_val_seg)
```
davidslater, line 96: I'm confused. Why are the validation and test labels segmented, but the training labels are not?

davidslater: Is this because inference can take arbitrary length inputs but training cannot?

armory/scenarios/librispeech_dev_clean_spectrogram.py
```diff
@@ -0,0 +1,171 @@
...
+            x_seg, y_seg = [], []
+            for xt, yt in zip(x, y):
+                n_seg = int(xt.shape[1] / n_tbins)
+                xt = xt[:, : n_seg * n_tbins]
+                for ii in range(n_seg):
+                    x_seg.append(xt[:, ii * n_tbins : (ii + 1) * n_tbins])
+                    y_seg.append(yt)
+            x_seg = np.array(x_seg)
+            x_seg = np.expand_dims(x_seg, -1)
+            y_seg = np.array(y_seg)
```
davidslater, line 124: Since this repeatedly shows up, I would recommend cleaning it up and making a function out of it:
```
def segment(x, y, n_time_bins):
    """Return segmented batch with dims ..."""
    ...
    return x_seg, y_seg
```

armory/scenarios/librispeech_dev_clean_spectrogram.py
seanpmorgan: Unfortunately it can't be since it requires adjusting the labels as well as input data

armory/scenarios/librispeech_dev_clean_spectrogram.py
seanpmorgan: No I didn't find a good reason this this. Instead of sampling all of the inputs it only sampled a single one. Updated.

PR #378, opened 2020-04-10
Store all model weights in single directory

seanpmorgan:


davidslater:
Now basically all of our models do this:
```python
    if weights_file:                                                                           
        saved_model_dir = paths.docker().saved_model_dir                                       
        filepath = os.path.join(saved_model_dir, weights_file)                                 
                                                                                               
        if not os.path.isfile(filepath):                                                       
            download_file_from_s3(                                                             
                "armory-public-data",                                                          
                f"model-weights/{weights_file}",                                               
                f"{saved_model_dir}/{weights_file}",                                           
            )                                                                                  
                                                                                               
        model.load(filepath)
```
The last line may differ. Can we have a util function like:
```
def get_s3_weights_file(weights_file):
    saved_model_dir = paths.docker().saved_model_dir                                           
    filepath = os.path.join(saved_model_dir, weights_file)                                     
                                                                                               
    if not os.path.isfile(filepath):                                                           
        download_file_from_s3(                                                                 
            "armory-public-data",                                                              
            f"model-weights/{weights_file}",                                                   
            f"{saved_model_dir}/{weights_file}",                                               
        )
    return filepath
```

Then in the model file, you would do:
```
     if weights_file:
        filepath = get_s3_weights_file(weights_file)
        model.load_weights(filepath)
```

seanpmorgan:
> This specific change looks good. Thoughts on a more general util function for loading file?

Yes I like the proposed fn. Will add it in a sec

PR #381, opened 2020-04-10
Unify cuda delegation for pytorch models

seanpmorgan:


armory/baseline_models/pytorch/ucf101_mars.py
```diff
@@ -137,6 +137,7 @@ def preprocessing_fn(inputs):
 # NOTE: PyTorchClassifier expects numpy input, not torch.Tensor input
 def get_art_model(model_kwargs, wrapper_kwargs, weights_file):
     model, optimizer = make_model(weights_file=weights_file, **model_kwargs)
+    model.to(DEVICE)
```
davidslater, line 140: I don't think this is necessary for this, as line 64:
```python
        checkpoint = torch.load(filepath, map_location=DEVICE)
```
should handle it

davidslater: I see how this is working across the files. While it may be a no-op, I think it is helpful to have this here to be explicit.

PR #384, opened 2020-04-12
Move check sums to Armory dataset dir

seanpmorgan:
Preferable to have these not downloaded into site-packages.

### DIFF
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -31,9 +31,10 @@ os.environ["KMP_WARNINGS"] = "0"
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


davidslater:
```CACHED_CHECKSUMS_DIR``` is not where we download to. It's where we keep the ground truth checksum values:
https://github.com/twosixlabs/armory/tree/master/armory/data/cached_s3_checksums

This follows the tensorflow datasets convention.
https://github.com/tensorflow/datasets/tree/v2.1.0/tensorflow_datasets/url_checksums

The cached dataset objects are downloaded to ```<datasets_dir>/cache```

PR #390, opened 2020-04-13
Update Poisoning Scenario

seanpmorgan:
Checkpointing, in order to discuss. The output I'm getting is:
```
    "results": {
        "test_accuracy": "0.4233570863024545",
        "validation_accuracy": "0.4203483768804434"
    }
```

This seems plausible since we're only targetting certain classes, but are these the metrics we want to report for poisoning success?


ng390:
Another checkpoint (albeit a bit more stochastic), the metric "targeted_accuracy" has varied from 82.82% to 95.43% on various runs

Issue #380, opened 2020-04-10
curl display errors


ng390:
The `curl` convenience wrapper method doesn't run into intermittent failures as the `download_file_from_s3` has, however its display sometimes glitches out as shown below. A separate download utility method, perhaps using `requests` or a wrapper for wget may be useful.

' 77 7565M   77 5882M    0     0  11.5M      0  0:10:53  0:08:28  0:
02:25 11.2M
 77 7565M   77 5894M    0     0  11.5M      0  0:10:53  0:08:29  0:02:24
 1
1.6M
 78 7565M   78 5906M    0     0  11.5
M      0  0:10:5
3
 0:08:30  0:02:23
 11.7M
 78 7565M   78 5918M    0     0  11.5M      0  0:10:53  0:08:31  0:0
2:22 11.5M
 78 7565M   78 5929M    0     0  11.5M      0  0:10:53  0:08:32  0:02
:21 11.6M
 78 7565M   78 5941M    0     0  11.5M      0  0:10:53  0:08:33  0:02:20 11.7M
 78 7565M   78 5953M
    0     0  11.5M      0  0:10:53  0:08:34  0:02:19 11.7M
 78 7565M   78 5965M    0     0  11.5M      0  0:10:53  0:08:35  0:02:18 11.7M
 78 7565M   78 5976M    0     0  11.5M      0
0:10:53  0:08
:36  0:02:17 1
1.7M
 79 7565M   79 5988M    0     0  11.5M      0  0:10
:53  0:08:
37  0:0
2:1
6 11
.6M
 79 7565M   79 5999M    0     0  11.5M      0  0:10:53  0:08:38  0:0
2:15 11.6
M
 79 7565M   79 6011M    0     0  11.5M      0  0:10:53  0:08:39  0:02:14 11.5M
 79 7565M   79 6022M    0     0  11.5M
0  0:10:53
0:08:40  0:0
2:13
11.5M

PR #392, opened 2020-04-13
Download using requests/tqdm

ng390:
Closes Issue #380


armory/data/utils.py
```diff
@@ -81,6 +83,20 @@ def curl(url: str, dirpath: str, filename: str) -> None:
         raise subprocess.CalledProcessError
 
 
+def download_requests(url: str, dirpath: str, filename: str):
+    filepath = os.path.join(dirpath, filename)
+    chunk_size = 1024
```
davidslater, line 88: This seems like a small chunk_size to me. Should we use a 4k (4096) chunk, as that is a common file block size?

PR #397, opened 2020-04-13
Metrics refactor

davidslater:
- [x] Add top_5_categorical_accuracy for UCF scenario
- [x] Add tests for metrics logger
- [x] Determine better name for MetricsListCounter

### DIFF
--- a/armory/utils/metrics.py
+++ b/armory/utils/metrics.py
@@ -16,6 +16,8 @@ def categorical_accuracy(y, y_pred):
     """
     Return the categorical accuracy of the predictions
     """
+    y = np.asarray(y)
+    y_pred = np.asarray(y_pred)
     if y.shape == y_pred.shape:
         return [int(x) for x in list(y == y_pred)]
     elif y.ndim + 1 == y_pred.ndim:
@@ -35,6 +37,10 @@ def top_n_categorical_accuracy(y, y_pred, n):
     if n < 1:
         raise ValueError(f"n must be a positive integer, not {n}")
     n = int(n)
+    if n == 1:
+        return categorical_accuracy(y, y_pred)
+    y = np.asarray(y)
+    y_pred = np.asarray(y_pred)
     if len(y) != len(y_pred):
         raise ValueError("y and y_pred are of different length")
     if y.shape == y_pred.shape:
@@ -50,6 +56,8 @@ def norm(x, x_adv, ord):
     """
     Return the given norm over a batch, outputting a list of floats
     """
+    x = np.asarray(x)
+    x_adv = np.asarray(x_adv)
     # cast to float first to prevent overflow errors
     diff = (x.astype(float) - x_adv.astype(float)).reshape(x.shape[0], -1)
     values = np.linalg.norm(diff, ord=ord, axis=1)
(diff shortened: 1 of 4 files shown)


armory/utils/metrics.py
```diff
@@ -68,6 +71,129 @@ def l0(x, x_adv):
     return norm(x, x_adv, 0)
 
 
+class MetricCounter:
```
seanpmorgan, line 74: MetricLog?
MetricHistory?

armory/utils/metrics.py
```diff
@@ -68,6 +71,129 @@ def l0(x, x_adv):
...
+        self._values.extend(value)
+
+    def values(self):
+        return self._values
+
+    def mean(self):
+        return sum(float(x) for x in self._values) / len(self._values)
+
+
+class MetricsListCounter:
```
seanpmorgan, line 101: MetricsContainer?
MetricsHistoryContainer?

davidslater: MetricsLogger?

armory/utils/metrics.py
davidslater: Maybe this could be MetricList (since it effectively stores a list of metric values), and I could update the function names to be more similar to list syntax, with append instead of update?

davidslater:
I think I will do the UCF changes when I make that PR.

I will go ahead and make some tests, though.

davidslater:
@seanpmorgan ready for another look

davidslater:
Only the stochastic pytorch piece is failing (unrelated to this PR).

armory/utils/metrics.py
```diff
@@ -68,6 +101,127 @@ def l0(x, x_adv):
...
+    """
+    Keeps track of all results from a single metric
+    """
+
+    def __init__(self, name, function=None):
+        if function is None:
+            try:
+                self.function = globals()[name]
+            except KeyError:
+                raise KeyError(f"{name} is not part of armory.utils.metrics")
```
seanpmorgan, line 114: Maybe safer to have something like `SUPPORTED_DATASETS` to check against?

armory/utils/metrics.py
```diff
@@ -68,6 +101,127 @@ def l0(x, x_adv):
     return norm(x, x_adv, 0)
 
 
+class MetricList:
```
seanpmorgan, line 104: Would it be useful to add an __iter__ method here if people are using this as a List?

armory/utils/metrics.py
```diff
@@ -68,6 +101,127 @@ def l0(x, x_adv):
...
+    """
+    Uses the set of task and perturbation metrics given to it.
+    """
+
+    def __init__(self, task=None, perturbation=None, means=True, full=False):
+        """
+        task - single metric or list of metrics
+        perturbation - single metric or list of metrics
+        means - whether to return the mean values for each metric
+        full - whether to return the full values for each metric
```
seanpmorgan, line 146: I think the word full here is a bit confusing.. especially when read from a config. Maybe something like `record_metric_per_sample` or something?

seanpmorgan:
Also small conflict

davidslater:
@seanpmorgan ready for another pass

PR #398, opened 2020-04-13
Config standardization

ng390:
Add "gpus" field to config files


docs/docker.md
```diff
@@ -77,6 +77,8 @@ The `use_gpu` flag takes a boolean true/false value, and specifies whether to us
 The `gpus` flag is optional, and is ignored if `use_gpu` is false. If `use_gpu` is true, it defaults to using all GPUs.
     If present, the value should be a `,`-separated list of numbers specifying the GPU index in `nvidia-smi`.
     For instance, `"gpus": "2,4,7"` would enable three GPUs with indexes 2, 4, and 7.
+    Setting the field to be `null` or setting the field to be `all` will enable use of all available gpus.
+    In other words, `"gpus": null` or `"gpus": "all"` will enable all available GPUs.
```
seanpmorgan, line 81: This seems confusing to me. @ng390 @davidslater how about we set default to "all" make it required to be a string (non-nullable)

ng390: This sounds good to me. I believe the behavior should be exactly the same, though in the current code a value of null would result in the NVIDIA_VISIBLE_DEVICES environmental variable being unset, versus the NVIDIA_VISIBLE_DEVICES environmental variable being set to "all"

PR #408, opened 2020-04-14
Enable all defenses

seanpmorgan:


armory/utils/config_loading.py
```diff
@@ -64,9 +64,28 @@ def load_attack(attack_config, classifier):
...
+    # TODO: Ensure preprocessor or postprocessor
+
+    defense_module = import_module(defense_config["module"])
+    defense_fn = getattr(defense_module, defense_config["name"])
+    defense = defense_fn(**defense_config["kwargs"])
+
+    if defense_config["type"] == "Preprocessor":
+        classifier.preprocessing_defences = [defense]
+    if defense_config["type"] == "Postprocessor":
+        classifier.postprocessing_defences = [defense]
```
davidslater, line 87: Probably want to check to see if the classifier already has one or both of these set. In that case, either error or append?

armory/scenarios/image_classification.py
```diff
@@ -41,10 +48,11 @@ def _evaluate(self, config: dict) -> dict:
...
             )
-            if config["defense"] is not None:
-                logger.info("loading defense")
-                defense = load_defense(config["defense"], classifier)
-                defense.fit_generator(train_data, **fit_kwargs)
+            if defense:
+                if defense_type in ["Trainer", "Transformer"]:
+                    logger.info(f"Training with {defense_type} defense...")
+                    defense = load_defense(config["defense"], classifier)
+                    defense.fit_generator(train_data, **fit_kwargs)
```
davidslater, line 55: This code won't work with DefensiveDistillation, the only concrete "Trainer" in ART.

armory/scenarios/image_classification.py
```diff
@@ -26,6 +27,12 @@ def _evaluate(self, config: dict) -> dict:
 
         model_config = config["model"]
         classifier, preprocessing_fn = load_model(model_config)
+        defense = config.get("defense")
+        if defense:
+            defense_type = defense["type"]
+            if defense_type in ["Preprocessor", "Postprocessor"]:
```
davidslater, line 33: to prevent nested double "if" statements everywhere you use this, you could do:
```
defense = config.get("defense", {})
defense_type = defense.get("type")
if defense_type in ["Preprocessor", "Postprocessor"]:
    ...
...
if defense_type in ["Trainer", "Transformer"]:
    ...
```

armory/scenarios/image_classification.py
seanpmorgan: I believe we'll need to raise an issue with ART to support fit_generator

davidslater: I think that Transformer should probably go after the main training loop in its own space. For DefensiveDistillation, you need to first train the main network, and then train the distilled network.

davidslater:
Rebased in #420

PR #414, opened 2020-04-14
[WIP] Poison defense

ng390:
Add defense and configurable poisoning rate (tuning needed). Uses numpy array for training rather than generator.


davidslater:
superceded by #422

PR #420, opened 2020-04-14
Add defenses

### LINKED ISSUE #366: Update Defenses
Ensure that each ART defense API can be run with our scenarios.

davidslater:
Fixes #366

### DIFF
--- a/armory/scenarios/image_classification.py
+++ b/armory/scenarios/image_classification.py
@@ -11,6 +11,7 @@ from armory.utils.config_loading import (
     load_model,
     load_attack,
     load_defense,
+    load_defense_internal,
 )
 from armory.utils import metrics
 from armory.scenarios.base import Scenario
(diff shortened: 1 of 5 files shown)


davidslater:
Ready for review

armory/scenarios/image_classification.py
```diff
@@ -41,9 +49,9 @@ def _evaluate(self, config: dict) -> dict:
                 split_type="train",
                 preprocessing_fn=preprocessing_fn,
             )
-            if config["defense"] is not None:
-                logger.info("loading defense")
-                defense = load_defense(config["defense"], classifier)
+            if defense_type == "Trainer":
```
seanpmorgan, line 52: So if you had a Transformer here, it would never get called by `load_defense_wrapper` and display the error. It would just run silently without the defense right?

davidslater: Correct.

I can have it try to load from the scenario and raise the NotImplementedError in the loading. That is more explicit. Thoughts?

seanpmorgan: Yeah I would prefer that I think

davidslater: Updated. Let me know what you think.

PR #424, opened 2020-04-14
Image DenseNet scenario: if not resisc-trained weights, use imagenet pretrained weights

grobertson-ext:
In the Image DenseNet RESISC45 scenario, use the imagenet pretrained weights for training instead of the randomly initialized ones if 'pretrained' set to `False`


seanpmorgan:
Add functionality for this in #426 Thanks!

PR #433, opened 2020-04-15
First pass at getting started documentation

seanpmorgan:


docs/getting_started.md
```diff
@@ -1,7 +1,51 @@
...
+pip install armory-testbed
+```
 
-### Running an evaluation
+When a user runs a given configuration file, the necessary docker image, datasets and 
+model weights will be pulled as needed. We do have convenience functions to download 
+all images, datasets and model weights for a scenario set release. This can take a 
+while so you may want to run it overnight:
+```
+armory download scenarios-set1.json
```
davidslater, line 14:
```suggestion
armory download scenarios-set1
```
The ".json" ending should be dropped.

PR #435, opened 2020-04-15
Add logging

seanpmorgan:


armory/scenarios/video_ucf101_scenario.py
```diff
@@ -73,6 +73,7 @@ def _evaluate(self, config: dict) -> dict:
                     if defense_type == "Trainer":
                         defense.fit(x, y, batch_size=batch_size, nb_epochs=1)
                     else:
+                        logger.info(f"Fitting classifier on clean train dataset...")
```
davidslater, line 76: Since this uses fit on each batch instead of fit_generator over the whole dataset, put this in an else block after line 59.

seanpmorgan: Good catch, fixed.

PR #438, opened 2020-04-15
Document baseline models

seanpmorgan:


docs/baseline_models.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Pretrined Weights
+Pretrained weights can be loaded in to the baseline models or custom models. This is 
+acheived by specifying the name in the `weights_file` field of a model's config. 
+
+When the model is loaded it will first try to load the file from the armory 
+`saved_model_dir`. This enables you to place your own custom weights in that directory 
+for loading. If the weights file is not found it'll then try to download the file from 
+our S3 bucket. Files that are available in the S3 bucket are listed in the table 
+below.
```
davidslater, line 14: from the armory S3 bucket?

Should we add a note that if the file doesn't exist locally or in the armory S3 bucket, this will return S3 errors?

seanpmorgan: Updated.

PR #444, opened 2020-04-15
Update help string

ng390:
Update help string


armory/__main__.py
```diff
@@ -168,7 +168,7 @@ def download(command_args, prog, description):
         dest="download_config",
         type=str,
         action=DownloadConfig,
-        help="Configuration for download of data, e.g. scenarios-set1",
+        help="Configuration for download of data as relative path, e.g. ./official_scenario_configs/scenarios-set1.json",
```
davidslater, line 171: @ng390 you'll need to modify DownloadConfig as well (lines 66-84)

Issue #310, opened 2020-03-24
Do not shuffle imagenet_adversarial dataset


davidslater:
I could not figure out how to get TFDS to *not* shuffle the items when writing the `imagenet_adversarial` dataset to file (which results in the "clean" and "adversarial" splits returning in a different order):
```
Downloading and preparing dataset imagenet_adversarial (?? GiB) to /armory/datasets/imagenet_adversarial/1.0.0...
Dl Completed...: 0 url [00:00, ? url/s]
Extraction completed...: 0 file [00:00, ? file/s]
Extraction completed...: 0 file [00:00, ? file/s]
Dl Size...: 0 MiB [00:00, ? MiB/s]

Dl Completed...: 0 url [00:00, ? url/s]
Shuffling and writing examples to /armory/datasets/imagenet_adversarial/1.0.0.incompleteDC9UV6/imagenet_adversarial-adversarial.tfrecord
Shuffling and writing examples to /armory/datasets/imagenet_adversarial/1.0.0.incompleteDC9UV6/imagenet_adversarial-clean.tfrecord                    
Computing statistics...: 100%|██████████████████████████████████████████████████████████████████████████████████████| 2/2 [00:04<00:00,  2.47s/ split]
Dataset imagenet_adversarial downloaded and prepared to /armory/datasets/imagenet_adversarial/1.0.0. Subsequent calls will reuse this data.
```

davidslater:
We are moving to a model where adversarial and clean are stored together in the same split, instead of trying to sort them separately in the same order, which does not appear possible at the moment.

Superceded by #487

PR #447, opened 2020-04-15
Densenet patch attack scenario

grobertson-ext:
Fails now because config's `attack `field cannot have `kwargs` or `module` or `name` set to null


armory/data/adversarial/resisc_adversarial_patch.py
```diff
@@ -0,0 +1,82 @@
...
+            ),
+            supervised_keys=("image", "label"),
+        )
+
+    def _split_generators(self, dl_manager):
+        """Returns SplitGenerators."""
+        path = _PATH  # dl_manager.download_and_extract(_URL)
+        splits = [
+            tfds.core.SplitGenerator(name=x, gen_kwargs={"path": path, "split": x})
+            for x in ("adversarial", "clean")
```
davidslater, line 47: What's the motivation for having a "clean" split for this? We already have a clean version of resisc45. I would prefer not to have the user have two large duplicative datasets.

Also, how would you split this into train and test? Is it already split? If so, that is unclear.

armory/data/adversarial/resisc_adversarial_patch.py
```diff
@@ -0,0 +1,82 @@
...
+
+    VERSION = tfds.core.Version("1.0.0")
+
+    def _info(self):
+        return tfds.core.DatasetInfo(
+            builder=self,
+            description=_DESCRIPTION,
+            features=tfds.features.FeaturesDict(
+                {
+                    "image": tfds.features.Tensor(shape=[224, 224, 3], dtype=tf.uint8),
```
davidslater, line 35: The dimensions here don't match the input dimensions of the original dataset, so if a user tries to use their own model that doesn't do the same preprocessing you have for densenet121, this will fail.

armory/data/datasets.py
```diff
@@ -25,6 +25,7 @@
 from armory import paths
 from armory.data.librispeech import librispeech_dev_clean_split  # noqa: F401
 from armory.data.resisc45 import resisc45_split  # noqa: F401
+from armory.data.adversarial import resisc45_patch_adversarial
```
davidslater, line 28: Add `  # noqa: F401` to the end of the line to prevent flake8 errors.
Also, name it something other than the name in your function of the same name below.

armory/scenarios/image_classification_load_adv_ds.py
```diff
@@ -0,0 +1,127 @@
...
+
+        attack_type = config["attack"]["type"]
+        if attack_type == "transfer":
+            # test_data_generator = attack     # suggested by DSlater but load_attack loads an ART attack function, not the attack dataset, right now. Plus, the dataset in the config json file is the adversarial dataset anyways, so load_dataset makes sense
+            test_data_generator = load_dataset(
+                config["dataset"],
+                epochs=1,
+                split_type="adv",
+                preprocessing_fn=preprocessing_fn,
+            )
```
davidslater, line 110: This, if I understand correctly, would do attacks on the same images as those trained on above. This should be split into train/test.

armory/scenarios/image_classification_load_adv_ds.py
```diff
@@ -0,0 +1,127 @@
...
+
+        # for x, y in tqdm(test_data_generator, desc="Attack"):
+        #     x_adv = attack.generate(x=x)
+        #     y_pred_adv = classifier.predict(x_adv)
+        #     metrics_logger.update_task(y, y_pred_adv, adversarial=True)
+        #     metrics_logger.update_perturbation(x, x_adv)
+        ####################################################
+        attack = load_attack(config["attack"], classifier)
+
+        attack_type = config["attack"]["type"]
```
davidslater, line 102: My recommendation is to make a new attack class in `armory/art_experimental/attacks`, like `TransferAttack`, that takes dataset information as input. Then the dataset information here can be loaded as attack parameters into the attack config.

scenario_configs/resisc45_patch_adversarial_densenet121.json
```diff
@@ -0,0 +1,45 @@
...
+        "fit_kwargs": {},
+        "model_kwargs": {},
+        "module": "armory.baseline_models.keras.densenet121_resisc45",
+        "name": "get_art_model",
+        "weights_file": "densenet121_resisc45_v1.h5",
+        "wrapper_kwargs": {}
+    },
+    "scenario": {
+        "kwargs": {},
+        "module": "armory.scenarios.image_classification",
```
davidslater, line 36: This should point to `armory.scenarios.image_classification_load_adv_ds.py`

tests/configs/densenet121_resisc45_scenario_baseline.json
```diff
@@ -0,0 +1,36 @@
+{
```
davidslater, line 1: This file looks like it's in the old format and should be removed.

armory/scenarios/image_classification_load_adv_ds.py
yusong-tan: I'm not sure creating a new type of attack is necessarily the simpler pathway, although I may not have thought about this from an implementation perspective as much as you have David.

Also, I think attack_type = "transfer" in the configuration file is not necessarily correct either.  The type of attack, given an adversarial dataset, depends on the model under attack, so it could be transfer or white-box.  Instead, the configuration should support a attack_type = "preloaded" (or something to that effect), to suggest that adversarial examples are not going to be generated on the fly in the scenario but rather from a file.

In the scenario, if attack_type = "preloaded", then an attack object will not be the instantiated; instead, the preloaded adversarial dataset will be loaded via a datagen and inference can be performed directly on that dataset.  On the other hand, if attack_type != "preloaded", then an attack object needs to be instantiated and used against a clean test dataset - of course, this will necessitate a change in the json format to allow two datasets to be declared under "dataset" field, one for the clean dataset and (optionally) one for the adversarial dataset.

davidslater: I like the idea of a "preloaded" attack_type, instead of a "transfer attack" (which might imply needing to load an additional model), with logic surrounding that attack_type. 

In terms of configuration, I think that it will be more understandable to have the dataset-specific information in the attack field instead of the dataset field - and this would only require a small (backwards-compatible) change to the configuration.

Another change to the logic in this case is that we'll need to provide paired clean and adversarial samples from the dataset in order to compute perturbation metrics for them.

armory/data/adversarial/resisc_adversarial_patch.py
davidslater: The "clean" version makes sense if it is otherwise difficult to pair with the "adversarial" split.

However, we do need to figure out train/test splits in this context, so that we're not returning adversarial test results on the training set.

yusong-tan: This is something that we've been discussing internally on how to best address.  Obviously, including the clean examples with every adversarial dataset is doubles the size of the dataset, which could be prohibitive to storage and downloading.

The other way to address it is to deliver adversarial datasets with only adversarial examples.  If examples for all datasets (clean and adv) are tagged in the tfds with a unique name, which I think is already done for many, if not all, datasets, and then we can do a brute-force pairing of adversarial and clean examples (this would take N^2 time), but it's probably okay since this doesn't need to be done often.

armory/data/adversarial/resisc_adversarial_patch.py
yusong-tan: For this scenario, resizing the adversarial examples up and then down again to attack the DenseNet probably would only minimally impact white-box attack efficacy, due to having 1:1 aspect ratio.

For scenarios, like video, where the aspect ratio of the original data (240:320) is different from the delivered adversarial example (112:112), we may need to create two different tfds scripts, one that keeps the adversarial examples at (112:112) and one that resizes to (240:320). The reason is that MARS preprocessing takes a center crop before resizing down.

armory/scenarios/image_classification_load_adv_ds.py
yusong-tan: I agree on the need to pair.  I don't know if tfds has an API to pull data by the "key" argument.

davidslater: If your team can look into pairing / aligning datasets, that would be very helpful. This would solve an existing bug on the imagenet adversarial dataset (adversarial and clean splits don't follow the same ordering): https://github.com/twosixlabs/armory/issues/310

davidslater:
Superceded by #530

Issue #468, opened 2020-04-21
Expand on the scenario documentation


seanpmorgan:
We need to add documentation for each scenario. Including description, baseline information, etc.

PR #472, opened 2020-04-21
Add placeholders for scenario information

seanpmorgan:
Part of #468


seanpmorgan:
> I think that this is fine, though we may want them to add something of a "threat scenario" or metrics of interest. Thoughts?

Good idea... updated, PTAL

PR #477, opened 2020-04-21
Updated docker release process

### LINKED ISSUE #476: Make Docker layer updates reuse cache
Currently, the docker release process is done in parallel using GitHub actions, which means that cache layers are not reused. This results in two problems:
1) Performers have to download the entire docker image from every patch release version update of armory
2) We have to push the full Docker image for every patch release version of armory

davidslater:
Fixes #476 

There are three things to resolve before removing WIP status:
1) How do we test this?
2) Is the `__previous_docker_version__` (or alternatively, `__previous_version__`) what we want to use to pull older docker containers? We could use the "latest" tag, but that may make it harder to test things. (And I don't know if we want performers trying to pull "latest".)
3) (minor) For our first build, we'll need a previous version of the `armory` container so that it doesn't error when it tries to pull one.

However, I think review before resolving these would be helpful.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,6 +24,26 @@ jobs:
         with:
           user: __token__
           password: ${{ secrets.pypi_token }}
+  release-armory-docker:
+    name: Build and release armory docker image
+    needs: [release-wheel]
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.client_payload.branch }}
+      - uses: actions/setup-python@v1
+        with:
+          python-version: '3.6'
+      - name: Build and release docker images
+        run: |
+          python -m pip install -r requirements.txt
+          version=$(python -c "import armory; print(armory.__version__)")
+          previous_version=$(python -c "import armory; print(armory.__previous_docker_version__)")
+          docker login --username ${{ secrets.DOCKER_USERNAME }} --password ${{ secrets.DOCKER_PASSWORD }}
+          docker pull twosixarmory/armory:${previous_version}
+          bash docker/build.sh armory
+          docker push twosixarmory/armory:${version}
   release-tf1-docker:
     name: Build and release tf1 docker image
     needs: [release-wheel]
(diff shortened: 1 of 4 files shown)


davidslater:
I looked into using some tag like "latest" or "stable" instead of the explicit `__previous_version__`, but it could cause issues if we need to replace an image, and it's hard to test, due to being implicit. On the other hand, if it's just used for caching layers, that may not be so bad - if we have to replace an image, it may just mess up the caching for that particular image.

The only other way forward I could think of is having a dedicated AWS VM for building images. But that is a bit more complicated to integrate with GitHub actions.

davidslater:
@seanpmorgan regarding testing, do you know how I could do this with my own dockerhub account? (Was that what you wer suggesting in our meeting?)

davidslater:
It tested this on my own docker account `davidslater`, with `tf1`: [link](https://hub.docker.com/repository/docker/davidslater/tf1). 

I tried a simple version change and a larger change to the armory base image, using the following script pulled from the release yml:
```
version=$(python -c "import armory; print(armory.__version__)")
  
#  release-armory-docker:
echo "armory:${version}"
docker pull davidslater/armory:latest
bash docker/build-ds.sh armory
docker push davidslater/armory:${version}
docker tag davidslater/armory:${version} davidslater/armory:latest
docker push davidslater/armory:latest

#  release-tf1-docker:
echo "tf1:${version}"
docker pull davidslater/armory:${version}
docker pull davidslater/tf1:latest
bash docker/build-ds.sh tf1
docker push davidslater/tf1:${version}
docker tag davidslater/tf1:${version} davidslater/tf1:latest
docker push davidslater/tf1:latest
```

For testing, I also had to replace `"twosixarmory"` with `"davidslater"` in `docker/build.sh` and `docker/tf1/Dockerfile`. Everything seemed to work as expected, with short uploads/downloads and proper caching. I'm not sure how to test it further.

davidslater:
@seanpmorgan ready for another review

Not sure if we want to integrate any of this into `build-dev.sh`. It would likely be faster to pull from dockerhub than build from scratch, if nothing has changed.

seanpmorgan:
> @seanpmorgan ready for another review
> 
> Not sure if we want to integrate any of this into `build-dev.sh`. It would likely be faster to pull from dockerhub than build from scratch, if nothing has changed.

I think it's okay to keep it as is for now.

PR #480, opened 2020-04-22
Run armory without docker

### LINKED ISSUE #457: running armory without docker
Hi, some of use cannot install docker in our grids. Can you provide some instructions of how to manually configure our environment to be able to run armory without docker? e.g., list of python packages, environment variables, other dependencies, etc.
Thanks

ng390:
Closes #457

### DIFF
--- a/armory/eval/evaluator.py
+++ b/armory/eval/evaluator.py
@@ -178,7 +178,14 @@ class Evaluator(object):
                     "jupyter, interactive, or bash commands only supported when running Docker containers."
                 )
             runner = self.manager.start_armory_instance(envs=self.extra_env_vars)
-            self._run_config(runner)
+            logger.warning(f"Outputs will be written to {self.output_dir}")
+            try:
+                self._run_config(runner)
+            except KeyboardInterrupt:
+                logger.warning("Keyboard interrupt caught")
+            finally:
+                logger.warning("Cleaning up...")
+            self._delete_tmp()
             return
 
         container_port = 8888
--- a/armory/eval/evaluator.py
+++ b/armory/eval/evaluator.py
@@ -75,6 +75,13 @@ class Evaluator(object):
 
             if self.config["sysconfig"].get("external_github_repo", None):
                 self._download_external()
+                self.extra_env_vars.update(
+                    {
+                        "PYTHONPATH": os.environ["PATH"]
+                        + os.pathsep
+                        + self.external_repo_dir
+                    }
+                )
 
             self.manager = HostManagementInstance()
             return
@@ -169,8 +176,9 @@ class Evaluator(object):
                 raise ValueError(
                     "jupyter, interactive, or bash commands only supported when running Docker containers."
                 )
-            runner = self.manager.start_armory_instance()
+            runner = self.manager.start_armory_instance(envs=self.extra_env_vars)
             self._run_config(runner)
+            return
 
         container_port = 8888
         ports = {container_port: host_port} if jupyter else None
--- a/armory/eval/evaluator.py
+++ b/armory/eval/evaluator.py
@@ -77,7 +77,7 @@ class Evaluator(object):
                 self._download_external()
                 self.extra_env_vars.update(
                     {
-                        "PYTHONPATH": os.environ["PATH"]
+                        "PYTHONPATH": os.environ["PYTHONPATH"]
                         + os.pathsep
                         + self.external_repo_dir
                     }
--- a/armory/eval/evaluator.py
+++ b/armory/eval/evaluator.py
@@ -15,6 +15,7 @@ import requests
 from docker.errors import ImageNotFound
 
 from armory.docker.management import ManagementInstance
+from armory.docker.host_management import HostManagementInstance
 from armory.docker import volumes_util
 from armory.utils.configuration import load_config
 from armory.utils import external_repo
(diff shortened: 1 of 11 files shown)


ng390:
Tested on Ubuntu 18.04 LTS with python 3.6 on cifar10_baseline and librispeech_baseline_sincnet

armory/eval/evaluator.py
```diff
@@ -144,8 +171,17 @@ def _delete_tmp(self):
...
-        container_port = 8888
         self._write_tmp()
+        if self.no_docker:
+            if jupyter or interactive or command:
+                raise ValueError(
+                    "jupyter, interactive, or bash commands only supported when running Docker containers."
+                )
+            runner = self.manager.start_armory_instance(envs=self.extra_env_vars)
+            self._run_config(runner)
+            return
```
davidslater, line 182: You need to do the `self._delete_tmp()` cleanup before returning.

davidslater: Similar logging to below would also be helpful.

davidslater:
Is there some sort of automated test that we can set up for this? Seems difficult to tell if/when we introduce a regression bug down the line.

davidslater:
In looking through this and considering the AWS evaluations (as well as the fact that evaluations are now done as scenarios and not explicitly in the Evaluator class), maybe we want to consider restructuring how the `Evaluator` and `Management` classes work and interact with each other (and how `__main__` interacts with them). @seanpmorgan thoughts?

davidslater:
Here's a thought: could we run armory recursively in "host" mode inside of an armory docker container to test?
```
python -m armory exec tf1 -- armory run --no-docker ...
```

seanpmorgan:
> Here's a thought: could we run armory recursively in "host" mode inside of an armory docker container to test?
> 
> ```
> python -m armory exec tf1 -- armory run --no-docker ...
> ```

Not quite following why that's needed. Can't we just run this on a fresh ubuntu VM in the CI?

davidslater:
> > Here's a thought: could we run armory recursively in "host" mode inside of an armory docker container to test?
> > ```
> > python -m armory exec tf1 -- armory run --no-docker ...
> > ```
> 
> Not quite following why that's needed. Can't we just run this on a fresh ubuntu VM in the CI?

Wouldn't we then have to work out the full build process for python, cuda, tensorflow, etc., on the fresh ubuntu VM? Aren't we trying to avoid supporting that?

ng390:
> > > Here's a thought: could we run armory recursively in "host" mode inside of an armory docker container to test?
> > > ```
> > > python -m armory exec tf1 -- armory run --no-docker ...
> > > ```
> > 
> > 
> > Not quite following why that's needed. Can't we just run this on a fresh ubuntu VM in the CI?
> 
> Wouldn't we then have to work out the full build process for python, cuda, tensorflow, etc., on the fresh ubuntu VM? Aren't we trying to avoid supporting that?

armory exec will run into permissions issues trying to access directories on the container in host mode, e.g. trying to read the default config in .armory. This can be circumvented by adding a flag to not set container_args["user"] or to docker run the command, e.g.
```
docker run -w /armory_dev twosixarmory/tf1:${version} python -m armory run --no-docker tests/scenarios/tf1/image_classification.json
```

AngusG:
Hi, thanks for this PR. I'm on an academic cluster with docker but no root access and was having permissions issues so this is helpful. 

I confirmed this PR works on CIFAR-10 and librispeech, but wondering how to specify the pytorch/tensorflow version which was previously supplied to `docker_image": "twosixarmory/pytorch:0.6.0",`. I'm assuming is ignored when using `--no-docker`. The cifar10 example seems to use aspects of pytorch and tensorflow at the same time, but defaults to tensorflow for evaluation. For example: 

```
armory run --no-docker official_scenario_configs/cifar10_baseline.json
...
2020-04-25 22:15:43 ws14 art.classifiers.pytorch[17952] INFO Inferred 1 hidden layers on PyTorch classifier.
...
tensorflow/core/platform/cpu_feature_guard.cc:142] Your CPU supports instructions that this TensorFlow binary was not compiled to use: AVX2 FMA
```
Is it possible to have it use one or the other, and is there a flag to do so? Thanks!

davidslater:
> Hi, thanks for this PR. I'm on an academic cluster with docker but no root access and was having permissions issues so this is helpful.
> 
> I confirmed this PR works on CIFAR-10 and librispeech, but wondering how to specify the pytorch/tensorflow version which was previously supplied to `docker_image": "twosixarmory/pytorch:0.6.0",`. I'm assuming is ignored when using `--no-docker`. The cifar10 example seems to use aspects of pytorch and tensorflow at the same time, but defaults to tensorflow for evaluation. For example:
> 
> ```
> armory run --no-docker official_scenario_configs/cifar10_baseline.json
> ...
> 2020-04-25 22:15:43 ws14 art.classifiers.pytorch[17952] INFO Inferred 1 hidden layers on PyTorch classifier.
> ...
> tensorflow/core/platform/cpu_feature_guard.cc:142] Your CPU supports instructions that this TensorFlow binary was not compiled to use: AVX2 FMA
> ```
> 
> Is it possible to have it use one or the other, and is there a flag to do so? Thanks!

In the config, you should be able to set `"docker_image": null`.

In terms of PyTorch and TensorFlow, we use TFDS under the hood for data loading, which is why TensorFlow is showing up. We are working on a native PyTorch data loader, but there is still some work to be done on that front. For evaluation, the framework used is based on the model selected.

PR #481, opened 2020-04-22
Enable tf.data.Dataset return

### LINKED ISSUE #452: Add tfds loader option to datasets
If someone wanted to train a model directly or play with the dataset in pure tensorflow, it would be difficult, as they would need to copy most of the logic from `datasets._generator_from_tfds`. However, if we made an optional return from https://github.com/twosixlabs/armory/blob/master/armory/data/datasets.py#L172 , it would enable easy usage, and would only require adding a parameter to each of the main dataset functions.

PyTorch, I think, would require much more effort, but TensorFlow should be pretty easy.

seanpmorgan:
Closes #452 

Wanted to discuss the implementation a little bit before finalizing. It's a bit tricker than expected since in TF the Dataset we're working with prior to `as_numpy` is a tf.data.Dataset which is not iterable since it's used in graph execution (This applies to TF1 and TF2). That means we can't pass it to ArmoryDataGenerator and it will not have a `get_batch` method for ART to use. It may still be useful for users to have this type of dataset be returned for their own training etc. but it doesn't fit as nicely in the framework as we had hoped. 

We can coerce the dataset into an iterable by executing the session, but not sure if that's preferable. Similar to how it's done for ART:
https://github.com/IBM/adversarial-robustness-toolbox/blob/c3434e6e9406df4a25d91dd7c746beb019be2bd2/tests/test_data_generators.py#L280

It would be nice to return ArmoryDataGenerators instead of a Union of ArmoryDataGenerators and tf.data.Datasets so after writing this I'm leaning toward working on that. 

Two other questions:
1. Do we want to cast the `framework` field to lowercase so there's not issue with say `NumPy` and `numpy` or would we rather be explicit.
2. Do we want to make the framework a required field in the config? Currently it will default to numpy but there's pros and cons to that.

### DIFF
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -10,7 +10,7 @@ The 'downloads' subdirectory under <dataset_dir> is reserved for caching.
 
 import logging
 import os
-from typing import Callable
+from typing import Callable, Union
 
 import numpy as np
 import tensorflow as tf
@@ -120,13 +120,25 @@ def _generator_from_tfds(
     variable_length=False,
     shuffle_files=True,
     cache_dataset: bool = True,
-):
+    framework: str = "numpy",
+) -> Union[ArmoryDataGenerator, tf.data.Dataset]:
     """
     If as_supervised=False, must designate keys as a tuple in supervised_xy_keys:
         supervised_xy_keys=('video', 'label')  # ucf101 dataset
     if variable_length=True and batch_size > 1:
         output batches are 1D np.arrays of objects
     """
+    supported_frameworks = ["tf", "pytorch", "numpy"]
+    if framework not in supported_frameworks:
+        raise ValueError(
+            f"Dataset framework field {framework} must be one of {supported_frameworks}"
+        )
+
+    if framework == "pytorch":
+        raise NotImplementedError(
+            "PyTorch native dataloaders are not yet supported. See issue https://github.com/twosixlabs/armory/issues/455"
+        )
+
     if not dataset_dir:
         dataset_dir = paths.docker().dataset_dir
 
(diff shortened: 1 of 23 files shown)


davidslater:
For the time being, I would rather get TF and PyTorch dataset objects that performs can play around with directly and then move to working on the integration with ART generators, so I'm fine with returning a `tf.data.Dataset`, at least for now. I would like to make sure that performers can access some native framework version of their datasets so that they can get start on research, even if it's not fully integrated with what we're doing yet.

If we execute the session, will that prevent other sorts of graph-based map functions?

Any thoughts on what Keras or TF Eager models would want for a dataset object for training? I think that we may want to think/talk through the main usage patterns for model training and inference.

As for the other two questions:
1) I would rather just stick with lowercase, explicit inputs.
2) Let's keep the config an optional field for now, defaulting to numpy, and only change it when we have fully working DataGenerator objects as output (with those fully integrated into the scenarios).

seanpmorgan:
> For the time being, I would rather get TF and PyTorch dataset objects that performs can play around with directly and then move to working on the integration with ART generators, so I'm fine with returning a `tf.data.Dataset`, at least for now. I would like to make sure that performers can access some native framework version of their datasets so that they can get start on research, even if it's not fully integrated with what we're doing yet.
> 
> If we execute the session, will that prevent other sorts of graph-based map functions?
> 
> Any thoughts on what Keras or TF Eager models would want for a dataset object for training? I think that we may want to think/talk through the main usage patterns for model training and inference.
> 
> As for the other two questions:
> 
> 1. I would rather just stick with lowercase, explicit inputs.
> 2. Let's keep the config an optional field for now, defaulting to numpy, and only change it when we have fully working DataGenerator objects as output (with those fully integrated into the scenarios).

Sounds good. Updates made PTAL

PR #490, opened 2020-04-28
Ucf101 adv dataset deliverables

yusong-tan:
- Added an adversarial databuilder for UCF101 (data/adversarial/ucf101_mars_perturbation_and_patch_adversarial_112x112.py), where the a dict of features are used to pair clean examples with adversarial examples 

- Minimally modified datasets.py to accommodate the adversarial dataset.

- Added a config (scenario_configs/ucf101_baseline_adversarial.json) showing one way of integrating the adversarial dataset under the "attack" module.  Currently, the batch size for adversarial dataset must be 1.  May require more modification to datasets.py for batch size > 1, but the effort may not be worth it if batch size = 1 could be enforced.

- Added a scenario file (scenarios/video_ucf101_scenario_adversarial_preloaded.py) showing an example of how to call preloaded adversarial dataset, run inference on it, and calculate distance measures between clean and adversarial examples.


yusong-tan:
Superseded by a new PR

PR #493, opened 2020-04-28
UCF101_adversarial_dataset_deliverables

yusong-tan:
Replaces a previous PR.

- Added an adversarial databuilder for UCF101 (data/adversarial/ucf101_mars_perturbation_and_patch_adversarial_112x112.py), where the a dict of features are used to pair clean examples with adversarial examples

- Minimally modified datasets.py to accommodate the adversarial dataset.

- Added a config (scenario_configs/ucf101_baseline_adversarial.json) showing one way of integrating the adversarial dataset under the "attack" module. Currently, the batch size for adversarial dataset must be 1. May require more modification to datasets.py for batch size > 1, but the effort may not be worth it if batch size = 1 could be enforced.

- Added a scenario file (scenarios/video_ucf101_scenario_adversarial_preloaded.py) showing an example of how to call preloaded adversarial dataset, run inference on it, and calculate distance measures between clean and adversarial examples.

### DIFF
--- a/armory/data/adversarial/ucf101_mars_perturbation_and_patch_adversarial_112x112.py
+++ b/armory/data/adversarial/ucf101_mars_perturbation_and_patch_adversarial_112x112.py
@@ -1,6 +1,6 @@
 """
-UFC101 actiona recognition adversarial dataset generated using non-universal,
-non-patch perturbation attack
+UFC101 action recognition adversarial dataset generated using
+perturbation and patch attacks
 """
 
 from __future__ import absolute_import
@@ -31,7 +31,7 @@ _CITATION = """
 }
 """
 _DESCRIPTION = """
-Dataset contains five randomly chosen videos from each class,
+Dataset contains five randomly chosen videos from each class, taken from test split,
 totaling 505 videos. Each video is broken into its component frames.  Therefore,
 the dataset comprises sum_i(N_i) images, where N_i is the number of frames in video
 i.  All frames are of the size (112, 112, 3). For each video, a clean version,
--- a/armory/data/adversarial/ucf101_mars_perturbation_and_patch_adversarial_112x112.py
+++ b/armory/data/adversarial/ucf101_mars_perturbation_and_patch_adversarial_112x112.py
@@ -1,0 +1,248 @@
+"""
+UFC101 actiona recognition adversarial dataset generated using non-universal,
+non-patch perturbation attack
+"""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import os
+
+import tensorflow.compat.v2 as tf
+import tensorflow_datasets.public_api as tfds
+
+_CITATION = """
+@article{DBLP:journals/corr/abs-1212-0402,
+  author    = {Khurram Soomro and
+               Amir Roshan Zamir and
+               Mubarak Shah},
+  title     = {{UCF101:} {A} Dataset of 101 Human Actions Classes From Videos in
+               The Wild},
+  journal   = {CoRR},
+  volume    = {abs/1212.0402},
+  year      = {2012},
+  url       = {http://arxiv.org/abs/1212.0402},
+  archivePrefix = {arXiv},
+  eprint    = {1212.0402},
+  timestamp = {Mon, 13 Aug 2018 16:47:45 +0200},
+  biburl    = {https://dblp.org/rec/bib/journals/corr/abs-1212-0402},
+  bibsource = {dblp computer science bibliography, https://dblp.org}
+}
+"""
+_DESCRIPTION = """
+Dataset contains five randomly chosen videos from each class,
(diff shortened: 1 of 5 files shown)


armory/scenarios/video_ucf101_scenario_adversarial_preloaded.py
```diff
@@ -0,0 +1,162 @@
...
+                    x_adv = x
+                else:
+                    # each x is of shape (n_stack, 3, 16, 112, 112)
+                    #    n_stack varies
+                    attack.set_params(batch_size=x.shape[0])
+                    x_adv = attack.generate(x=x)
+
+                # combine predictions across all stacks
+                y_pred = np.mean(classifier.predict(x_adv), axis=0)
+                metrics_logger.update_task(y, y_pred, adversarial=True)
```
davidslater, line 140: This isn't pulling the pairs of (clean, adversarial), so when the perturbation is measured in line 141, it will be 0.

yusong-tan: I didn't see line 141.  Is it suppose to calculate some distance metric?  I can add in pulling of clean example before line 141.

Lines 146-162 is a template to pulling the pairs.

davidslater: Yes, 141 is calculating perturbation distance(s) between clean and adversarial for the list of metrics provided by the user. I think that having the default be to load both clean and adversarial and calculate those perturbations would work best here. I don't think that the overhead of loading both is going to be overly high.

armory/data/adversarial/ucf101_mars_perturbation_and_patch_adversarial_112x112.py
```diff
@@ -0,0 +1,248 @@
...
+  eprint    = {1212.0402},
+  timestamp = {Mon, 13 Aug 2018 16:47:45 +0200},
+  biburl    = {https://dblp.org/rec/bib/journals/corr/abs-1212-0402},
+  bibsource = {dblp computer science bibliography, https://dblp.org}
+}
+"""
+_DESCRIPTION = """
+Dataset contains five randomly chosen videos from each class,
+totaling 505 videos. Each video is broken into its component frames.  Therefore,
+the dataset comprises sum_i(N_i) images, where N_i is the number of frames in video
```
davidslater, line 36: Can you specify whether these videos were taken from the test split, train split, or some combination?

armory/scenarios/video_ucf101_scenario_adversarial_preloaded.py
yusong-tan: Got it.  I just saw that there's a "metric" key in the config.  Is your intention of "metric/perturbation" to match the distance used in generating the attack, or any user-defined distance metric?  If the latter, is there a way to calculate multiple metrics (L-0, L-2, L-inf, Wasserstein) in one pass, like passing a list to "metric/perturbation?"

armory/data/adversarial/ucf101_mars_perturbation_and_patch_adversarial_112x112.py
yusong-tan: Will do.

davidslater:
"Currently, the batch size for adversarial dataset must be 1. May require more modification to datasets.py for batch size > 1, but the effort may not be worth it if batch size = 1 could be enforced."

Is this a constraint of the underlying dataset, or a constraint of the model? What breaks if batch size = 2, for instance?

yusong-tan:
Batch_size = 1 seems to be a constraint in get_batch() or _generate_from_tfds()
https://github.com/twosixlabs/armory/blob/03be34f654f0787452c91c79a73b5e7adf6ec954/armory/data/datasets.py#L74

https://github.com/twosixlabs/armory/blob/03be34f654f0787452c91c79a73b5e7adf6ec954/armory/data/datasets.py#L110

When batch_size > 1, then the dict of features fails to accommodate a batch of variable length (N_i, 3, 16, 112, 112) videos.

Not sure exactly where the issue is, but I suspect, ds returned by tfds.load() isn't handled well by the current code.

https://github.com/twosixlabs/armory/blob/03be34f654f0787452c91c79a73b5e7adf6ec954/armory/data/datasets.py#L152

armory/scenarios/video_ucf101_scenario_adversarial_preloaded.py
davidslater: The intention is for any distance metric, though right now it only uses ones baked in to Armory, which is currently Lp (and L0) norms. You should be able to pass a list into the perturbation part of the metrics config and it will generate multiple values.

davidslater:
I'm pulling the dataset, presently, so that I can do full end-to-end testing. It looks like it will take several hours to download.

davidslater:
Merged, with modifications, on https://github.com/twosixlabs/armory/pull/519

PR #501, opened 2020-04-30
Gather external repo from within scenario

seanpmorgan:


armory/scenarios/base.py
```diff
@@ -34,11 +36,25 @@ def evaluate(self, config: dict):
...
         if results is None:
             logger.warning(f"{self._evaluate} returned None, not a dict")
         self.save(config, results)
 
+    def setup(self, config: dict):
+        """
+        Performs scenario setup on the evaluating system.
+        """
+        if config["sysconfig"].get("external_github_repo", None):
+            external_repo_dir = paths.docker().external_repo_dir
```
davidslater, line 50: This line will break when running in `--no-docker` mode, I think, since it will point to `/armory/tmp/external`.

PR #503, opened 2020-05-04
Refactor scenario launching

### LINKED ISSUE #470: Move scenario setup inside the Docker container
When running a scenario the main Armory executable performs various setup tasks (downloading and unpacking external GitHub repos, downloading datasets, etc) on the host before launching the scenario Docker container.

AWS ECS Tasks do not allow us to run setup commands on the EC2 host before starting a Docker container. The scenario setup tasks should thus be moved into the scenario container. This could be accomplished by adding a new command to armory that takes the scenario configuration file as input, performs the setup tasks, then starts the scenario script. This command would be specified as the entrypoint for the scenario Docker container. (Currently Armory specifies the scenario script as the Docker container entrypoint)

seanpmorgan:
Closes #470 

This PR removes a good amount of complexity and accomplishes:
* Scenario configs can now be passed as serialized strings. This enables the AWS task definition to contain all necessary information for a run
* Output and tmp directories are created inside the container. This means we no longer need a host to launch evaluations and enables cloud evaluations


armory/scenarios/base.py
```diff
@@ -41,11 +43,18 @@ def evaluate(self, config: dict):
...
         """
+
+        id = uuid.uuid4()
+        self.scenario_output_dir = os.path.join(paths.docker().output_dir, str(id))
+        os.mkdir(self.scenario_output_dir)
+        logger.warning(f"Outputs will be written to {self.scenario_output_dir}")
+
         if config["sysconfig"].get("external_github_repo", None):
             external_repo_dir = paths.docker().external_repo_dir
             external_repo.download_and_extract_repo(
```
davidslater, line 60: Suggestion here:
Have `paths` handle the creation of output and tmp directories the first time `paths.docker()` is called. Since this is then cached, all the other code shouldn't need changing. The tmp and output subdirs could just be returned directly from paths.

armory/eval/evaluator.py
```diff
@@ -193,18 +151,28 @@ def run(
                 )
             else:
                 logger.error("Is Docker Daemon running?")
-        self._delete_tmp()
+        if interactive:
+            self._delete_tmp_config()
+
+    def _escape_config_string(self):
+        escaped_config = str(self.config).replace(" ", "").replace("'", '\\"')
```
adamj26, line 158: As per our discussion over Zoom, given that users can put whatever they want in the `adhoc` section of the config file, including keys or values containing spaces, it'd be safer and more robust to encode the config in base64

seanpmorgan: Done.

armory/scenarios/base.py
seanpmorgan: I'm in favor of directory creation being explicit to the someone reading the scenario code instead of done behind the scenes when accessing paths. I'm planning on submitting a paths refactor today. Perhaps we could discuss in that PR?

davidslater: That works for me.

PR #522, opened 2020-05-06
Enable multiple workers in PyTorch DataLoader

### LINKED ISSUE #520: Performance of PyTorch's DataLoader in Armory
There seems to be a performance issue when using [torch.utils.data.DataLoader](https://pytorch.org/docs/stable/data.html) in our code. When we set `num_workers` to something greater than 1, the DataLoader is unable to use all of the CPU resources because all of the workers will be on same core. I haven't root-caused why this happens, but I do know that if we import torch before tensorflow, that fixes the issue. The earliest import of tensorflow I can find in Armory is at [armory/data/datasets.py#L16](https://github.com/twosixlabs/armory/blob/master/armory/data/datasets.py#L16).

davidslater:
fixes #520

### DIFF
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -23,8 +23,8 @@ except ImportError:
 import tensorflow as tf
 import tensorflow_datasets as tfds
 import apache_beam as beam
-
 from art.data_generators import DataGenerator
+
 from armory.data.utils import (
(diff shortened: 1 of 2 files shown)


davidslater:
As far as I can tell, this is the first time tensorflow is imported when using armory scenarios or importing datasets.

PR #525, opened 2020-05-06
added example config and scenario for using ART preprocessing defense…

yusong-tan:
…s as baseline

Scenario manually breaks video 5D input tensors into multiple 4D tensors, each independently processed by ART image-based preprocessing defenses.

### DIFF
--- a/armory/scenarios/video_ucf101_scenario_adversarial_preloaded_and_defended.py
+++ b/armory/scenarios/video_ucf101_scenario_adversarial_preloaded_and_defended.py
@@ -1,0 +1,215 @@
+"""
+Classifier evaluation within ARMORY
+
+Scenario Contributor: MITRE Corporation
+"""
+
+import logging
+
+import numpy as np
+from tqdm import tqdm
+
+from armory.utils.config_loading import (
+    load_dataset,
+    load_model,
+    load_attack,
+    load_defense_wrapper,
+    load_defense_internal,
+    load,
+)
+from armory.utils import metrics
+from armory.scenarios.base import Scenario
+
+logger = logging.getLogger(__name__)
+
+
+class Ucf101(Scenario):
+    def _evaluate(self, config: dict) -> dict:
+        """
+        Evaluate the config and return a results dict
+        """
+
+        model_config = config["model"]
+        classifier, preprocessing_fn = load_model(model_config)
+
+        defense_config = config.get("defense") or {}
+        defense_type = defense_config.get("type")
+
+        IMAGE_DEFENSE_SUPPORT_ONLY_4D = True
+
+        if defense_type in ["Preprocessor", "Postprocessor"]:
+            if IMAGE_DEFENSE_SUPPORT_ONLY_4D:
+                # Architectures that use 3D convolution takes 5D
+                # inputs, usually in the shape of (nb, c, t, h, w)
+                # To use image-based defense that support 4D inputs,
+                # need to manually reshape 5D inputs into 4D, process
+                # defense, and then reshape back to 5D
+                defense = load(config.get("defense"))
+            else:
+                logger.info(f"Applying internal {defense_type} defense to classifier")
+                classifier = load_defense_internal(config["defense"], classifier)
+
+        if model_config["fit"]:
+            classifier.set_learning_phase(True)
+            logger.info(
+                f"Fitting model {model_config['module']}.{model_config['name']}..."
+            )
+            train_epochs = config["model"]["fit_kwargs"]["nb_epochs"]
+            batch_size = config["dataset"]["batch_size"]
+
+            logger.info(f"Loading train dataset {config['dataset']['name']}...")
+            train_data = load_dataset(
+                config["dataset"],
+                epochs=train_epochs,
+                split_type="train",
+                preprocessing_fn=preprocessing_fn,
+            )
+
+            if defense_type == "Trainer":
+                logger.info(f"Training with {defense_type} defense...")
+                defense = load_defense_wrapper(config["defense"], classifier)
+            else:
+                logger.info(f"Fitting classifier on clean train dataset...")
+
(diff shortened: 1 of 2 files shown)


armory/scenarios/video_ucf101_scenario_adversarial_preloaded_and_defended.py
```diff
@@ -0,0 +1,215 @@
...
+            if IMAGE_DEFENSE_SUPPORT_ONLY_4D:
+                # Architectures that use 3D convolution takes 5D
+                # inputs, usually in the shape of (nb, c, t, h, w)
+                # To use image-based defense that support 4D inputs,
+                # need to manually reshape 5D inputs into 4D, process
+                # defense, and then reshape back to 5D
+                defense = load(config.get("defense"))
+            else:
+                logger.info(f"Applying internal {defense_type} defense to classifier")
+                classifier = load_defense_internal(config["defense"], classifier)
```
davidslater, line 50: This hard coding makes it difficult for performers to apply to defenses that they add, since they would have to extend / or modify the scenario directly.

armory/scenarios/video_ucf101_scenario_adversarial_preloaded_and_defended.py
```diff
@@ -0,0 +1,215 @@
...
+                                    # jpegcompression requires unnormalized inputs
+                                    xt[..., 0] += 114.7748
+                                    xt[..., 1] += 107.7354
+                                    xt[..., 2] += 99.4750
+                                xt, _ = defense(xt)
+                                if defense_config.get("name") == "JpegCompression":
+                                    xt[..., 0] -= 114.7748
+                                    xt[..., 1] -= 107.7354
+                                    xt[..., 2] -= 99.4750
+                                x[ns, ...] = np.transpose(xt, (3, 0, 1, 2))
```
davidslater, line 103: This shows up several places in the script. Can you abstract this into a Preprocessing ART function in armory.art_experimental.defences that wraps the underlying JpegCompression and works seamlessly in the original scenario?

Ideally, the hard-coded means and transpose args would be stored in the defense's `kwargs` portion of the config.

armory/scenarios/video_ucf101_scenario_adversarial_preloaded_and_defended.py
yusong-tan: I agree.  The performers could use this scenario only for baseline defense evaluation, and then use another scenario for evaluating their own defense.  Unless image-based defenses accepts 5D inputs, I'm not sure what cleaner workaround there is.

armory/scenarios/video_ucf101_scenario_adversarial_preloaded_and_defended.py
yusong-tan: If you can point me to an example of how an experimental defense is used in a scenario, then I can try to take a crack at it if the effort isn't too large.  Otherwise, I'm leaving the ART portion to the IBM team and going to focus on scenarios instead.

davidslater: I tested this out and it seemed to work. This is my file `armory.art_experimental.defences.jpeg_compression_5d.py`:
```
from art.defences import JpegCompression                                                 
import numpy as np                                                                       
                                                                                         
                                                                                         
class JpegCompression5D(JpegCompression):                                                
    def __init__(                                                                        
        self,                                                                            
        clip_values,                                                                     
        quality=50,                                                                      
        channel_index=3,                                                                 
        apply_fit=True,                                                                  
        apply_predict=False,                                                             
        transpose=None,                                                                  
        means=None,                                                                      
    ):                                                                                   
        super().__init__(                                                                
            clip_values,                                                                 
            quality=quality,                                                             
            channel_index=channel_index,                                                 
            apply_fit=apply_fit,                                                         
            apply_predict=apply_predict,                                                 
        )                                                                                
        # transpose = (1, 2, 3, 0),                                                      
        if transpose is None:                                                            
            transpose = (0, 1, 2, 3)  # identify operation                               
        if len(transpose) != 4:                                                          
            raise ValueError("transpose must have 4 dimensions specified")               
        self.transpose = transpose                                                       
                                                                                         
        # means = (114.7748, 107.7354, 99.4750),                                         
        if means is None:                                                                
            means = (0.0, 0.0, 0.0)  # identity operation                                
        if len(means) != 3:                                                              
            raise ValueError("means must have 3 values, one per channel")                
        self.means = means                                                               
                                                                                         
    def __call__(self, x, y=None):                                                       
        x_new = []                                                                       
        for x_i in x:                                                                    
            x_i = x_i.transpose(self.transpose)                                          
            x_i = x_i + self.means  # use broadcasting                                   
            x_i, _ = super().__call__(x_i)                                               
            x_i = x_i - self.means                                                       
            x_i = x_i.transpose(np.argsort(self.transpose))                              
            x_new.append(x_i)                                                            
        return np.stack(x_new), y 
```

To test:
```
from armory.art_experimental.defences import jpeg_compression_5d
defense = jpeg_compression_5d.JpegCompression5D(clip_values=(0, 255))
ds = datasets.ucf101()
defense(x)
```
For use with the MARS preprocessing function:
```
from armory.art_experimental.defences import jpeg_compression_5d
defense = jpeg_compression_5d.JpegCompression5D(clip_values=(0, 255), tranpose=(1, 2, 3, 0), means=(114.7748, 107.7354, 99.4750))
ds = datasets.ucf101(preprocessing_fn=preprocessing_fn)
defense(x)
```

davidslater: I think that if you add this file and then modify your config, this should work with the original scenario, but let me know if you run into issues.

yusong-tan:
Replacing with a new PR that adds art_experimental.defences.JpegCompression5D

Issue #505, opened 2020-05-04
Use argparse subparsers in __main__


davidslater:

davidslater:
Currently, new parsers are created inside each command.

davidslater:
Current setup seems to work fine.

PR #536, opened 2020-05-12
Clean up of __main__ and interface with evaluator

### LINKED ISSUE #292: Support CUDA 9.x limited devices

### LINKED ISSUE #478: Use of logging and print in __main__
Currently, we use the built-in `print` function in `armory/__main__.py`, but logging everywhere else. This was likely done to make sure logging levels were set up prior to trying to log output, and to ensure that if there was only the null handler set up, it would still output to the shell.

It is unclear if this is needed and/or true.

davidslater:
Primary changes:
* Updated usage of `print` and `logger`, so that the former only happens either before the log hander has been installed or when directly interacting with the user (e.g., in `configure`)
* Making functions for adding parser elements - this cleans up a lot of duplicate code and shortens each of the command functions quite a bit.
* Making a cleaner abstraction between `__main__` and `Evaluator` by making the latter only take a configuration dictionary and removing a lot of the kwargs (except `no_docker`, which can later be expanded to a target, e.g. one of `docker`, `host`, `aws`).
* It also updates documentation for this and using CUDA 9.

Fixes #478 
Closes #292 

It doesn't solve #505 , but would make that transition much easier.

### DIFF
--- a/armory/__main__.py
+++ b/armory/__main__.py
@@ -10,6 +10,7 @@ This runs an arbitrary config file. Results are output to the `outputs/` directo
 """
 
 import argparse
+import json
 import logging
 import os
 import sys
@@ -24,6 +25,9 @@ from armory.configuration import save_config
 from armory.eval import Evaluator
 from armory.docker import images
 from armory.utils import docker_api
+from armory.utils.configuration import load_config
+
+logger = logging.getLogger(__name__)
 
 
 class PortNumber(argparse.Action):
@@ -76,11 +80,10 @@ class DownloadConfig(argparse.Action):
             )
 
 
-def run(command_args, prog, description):
-    parser = argparse.ArgumentParser(prog=prog, description=description)
-    parser.add_argument(
-        "filepath", metavar="<json_config>", type=str, help="json config file"
-    )
+# Helper functions for parsers
+
+
+def _debug(parser):
     parser.add_argument(
         "-d",
         "--debug",
@@ -90,6 +93,9 @@ def run(command_args, prog, description):
         default=logging.INFO,
         help="Debug output (logging=DEBUG)",
     )
+
+
+def _interactive(parser):
     parser.add_argument(
         "-i",
         "--interactive",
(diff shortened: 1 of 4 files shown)


armory/__main__.py
```diff
@@ -99,6 +105,9 @@ def run(command_args, prog, description):
         default=False,
```
adamj26, line 105: This combination of `action`, `const`, and `default` can be replaced with `action="store_true"`, and `dest` doesn't need to be set when it's the same as the long version of the flag.

armory/__main__.py
```diff
@@ -108,6 +117,9 @@ def run(command_args, prog, description):
...
```
adamj26, line 117: `dest` can be removed and `action`, `const`, and `default` collapsed to `action="store_true"`

armory/__main__.py
adamj26: There're some other places in this file where these changes could be made; but I don't want to spam comments all over the place

docs/docker.md
```diff
@@ -130,6 +132,47 @@ The `gpus` flag is optional, and is ignored if `use_gpu` is false. If `use_gpu`
...
+due to an inability to upgrade your local server, and we do not have it baked in to our docker
+containers. To use CUDA 9 in our docker container, you will need to replace the line
+```
+ cudatoolkit=10.1 -c pytorch && \
+```
+with
+```
+ cudatoolkit=9.2 -c pytorch && \
+```
+in `docker/pytorch/Dockerfile` and build them the pytorch container locally.
```
ng390, line 173:
```suggestion
in `docker/pytorch/Dockerfile` and build the pytorch container locally.
```

ng390:
The --use-gpu and --gpus functionality looks good to me.

PR #540, opened 2020-05-12
Example clean/adversarial label pairing for UCF101

yusong-tan:
- Modified UCF101 adversarial dataset to return dict for labels

- Added "adversarial_goal" key under "attack/kwargs" in the scenario config.  For "targeted" goal, the clean and adversarial labels can be used to calculate classification accuracy and targeted fooling rate.  For "untargeted" goal, the adversarial labels can be ignored.

### DIFF
--- a/armory/data/adversarial/ucf101_mars_perturbation_and_patch_adversarial_112x112.py
+++ b/armory/data/adversarial/ucf101_mars_perturbation_and_patch_adversarial_112x112.py
@@ -171,11 +171,17 @@ class Ucf101MarsPerturbationAndPatchAdversarial112x112(tfds.core.GeneratorBasedB
                             shape=(None, 112, 112, 3)
                         ),
                     },
-                    "label": tfds.features.ClassLabel(names=_LABELS),
+                    "labels": {
+                        "clean": tfds.features.ClassLabel(names=_LABELS),
+                        "adversarial_perturbation": tfds.features.ClassLabel(
+                            names=_LABELS
+                        ),
+                        "adversarial_patch": tfds.features.ClassLabel(names=_LABELS),
+                    },
                     "videoname": tfds.features.Text(),
                 }
             ),
-            supervised_keys=("videos", "label"),
+            supervised_keys=("videos", "labels"),
             homepage=_URL,
             citation=_CITATION,
         )
@@ -201,7 +207,7 @@ class Ucf101MarsPerturbationAndPatchAdversarial112x112(tfds.core.GeneratorBasedB
             os.path.join(data_dir_path, root_dir, split_dirs[0])
         )
         labels.sort()
-        for label in labels:
+        for i, label in enumerate(labels):
             videonames = tf.io.gfile.listdir(
                 os.path.join(data_dir_path, root_dir, split_dirs[0], label)
             )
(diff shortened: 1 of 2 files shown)


armory/data/adversarial/ucf101_mars_perturbation_and_patch_adversarial_112x112.py
```diff
@@ -171,11 +171,17 @@ def _info(self):
...
                         ),
                     },
-                    "label": tfds.features.ClassLabel(names=_LABELS),
+                    "labels": {
+                        "clean": tfds.features.ClassLabel(names=_LABELS),
+                        "adversarial_perturbation": tfds.features.ClassLabel(
+                            names=_LABELS
+                        ),
+                        "adversarial_patch": tfds.features.ClassLabel(names=_LABELS),
+                    },
```
davidslater, line 180: This looks like it will break existing things that currently use the adversarial dataset, yes?

So merging this is blocked on updates to the how these loaders work.

armory/data/adversarial/ucf101_mars_perturbation_and_patch_adversarial_112x112.py
```diff
@@ -249,7 +255,11 @@ def _generate_examples(self, data_dir_path):
                         "adversarial_perturbation": adv_pert,
                         "adversarial_patch": adv_patch,
                     },
-                    "label": label,
+                    "labels": {
+                        "clean": label,
+                        "adversarial_perturbation": label,  # untargeted, label not used
+                        "adversarial_patch": labels[(i + 1) % len(_LABELS)],  # targeted
```
davidslater, line 261: Can you document this a little more? You're doing round-robin targeted attacks, specifically?

armory/data/adversarial/ucf101_mars_perturbation_and_patch_adversarial_112x112.py
davidslater: Do we have similar targeted labels for the other adversarial datasets?

davidslater:
Blocked on #571

armory/data/adversarial/ucf101_mars_perturbation_and_patch_adversarial_112x112.py
yusong-tan: This, this will break existing things.  This is an example of how we could pair the labels, but not necessarily the final version.  To separate out the labels, I imagine you could use a method similar to that used to separate out the clean/adv inputs?

Image and audio will also have targeted labels, but the audio targeted adversarial dataset is not ready, so the current delivery only includes clean and untargeted examples.  Going forward, for most scenarios will have both targeted and untargeted examples

armory/data/adversarial/ucf101_mars_perturbation_and_patch_adversarial_112x112.py
yusong-tan: Yes, I can document and update the VERSION.

armory/data/adversarial/ucf101_mars_perturbation_and_patch_adversarial_112x112.py
yusong-tan: We can discuss more tomorrow on how you'd want to proceed with this

armory/data/adversarial/ucf101_mars_perturbation_and_patch_adversarial_112x112.py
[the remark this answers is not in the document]
yusong-tan: That's a good catch. The target labels were generated according to the order in `_LABELS`.

The following are the differences between `_LABELS` and `labels.sort()`:

`_LABELS` `labels.sort()`
Hammering HammerThrow
HammerThrow Hammering
JumpingJack JumpRope
JumpRope JumpingJack

Since the set of `labels` is the same as `_LABELS`, I think one way to correct the bug is to replace

`labels.sort()` with `labels = _LABELS`. In that case, https://github.com/yusong-tan/armory/blob/90ba4ac31b4e2e8fae6d6b21c704e4aad2087cd6/armory/data/adversarial/ucf101_mars_perturbation_and_patch_adversarial_112x112.py#L206-L208 can be deleted.

davidslater:
Merged via #622

PR #554, opened 2020-05-15
Update poisoning scenario

### LINKED ISSUE #528: % of Poisoned Data
According to the scenario descrpition, 1/5/10% of the source class will be poisoned at training time. However the current scenario seems to be poisoning 70% of the source test at training time.

ng390:
Closes #528

### DIFF
--- a/armory/scenarios/poisoning_gtsrb_scenario.py
+++ b/armory/scenarios/poisoning_gtsrb_scenario.py
@@ -153,11 +153,23 @@ class GTSRB(Scenario):
             preprocessing_fn=preprocessing_fn,
         )
         validation_metric = metrics.MetricList("categorical_accuracy")
+        target_class_benign_metric = metrics.MetricList("categorical_accuracy")
         for x, y in tqdm(test_data, desc="Testing"):
             y_pred = classifier.predict(x)
             validation_metric.append(y, y_pred)
+            y_pred_tgt_class = y_pred[y == src_class]
+            if len(y_pred_tgt_class):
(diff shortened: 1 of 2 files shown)


ng390:
With 10% poisoning of the training set and 70% of the test set poisoned (both are configurable parameters), as expected the attack is somewhat weaker. The following empirical results are over 5 trials. If the model is undefended, the attack succeeds 56.8% (std dev 2.0%) of the time. If the model is defended the attack succeeds only 0.1% of the time (std dev 0.1%). The defense is not completely effective in the sense that unpoisoned accuracy drops 14.9% (std dev 4.3%).

armory/scenarios/poisoning_gtsrb_scenario.py
```diff
@@ -169,10 +169,12 @@ def _evaluate(self, config: dict) -> dict:
...
             targeted_test_metric = metrics.MetricList("categorical_accuracy")
+            fraction_poisoned_test = config["adhoc"].get("fraction_poisoned_test", 1)
             for x_test, y_test in tqdm(test_data, desc="Testing"):
-                x_test, _ = poison_batch(
-                    x_test, y_test, src_class, tgt_class, len(y_test), attack
-                )
+                if np.random.rand() < fraction_poisoned_test:
+                    x_test, _ = poison_batch(
+                        x_test, y_test, src_class, tgt_class, len(y_test), attack
+                    )
```
davidslater, line 177: I don't understand the purpose of only poisoning a fraction of the test points if we already test on all of the benign points.

Issue #570, opened 2020-05-18
sampling rate librispeech data


jesus-villalba:
Hi,
In this file
https://github.com/twosixlabs/armory/blob/master/armory/baseline_models/pytorch/sincnet.py
the sampling_rate is hardcoded to 8kHz,
but librispeech is 16kHz
Are the tfrecord files that armory downloads at 8 or 16k?
We trained all our models at 16k already for this.
Thanks
Jesus

davidslater:
We definitely need to document sampling rate for both of the audio datasets.

I will look into the our dataset and verify. I think that the underlying dataset is 16kHz, as shown [here](https://github.com/twosixlabs/armory/blob/master/armory/data/librispeech/librispeech_dev_clean_split.py#L162), and we leverage the underlying TFDS dataset, which also uses 16kHz [here](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/audio/librispeech.py#L129). 

However, it does look like we're using 8kHz in the SincNet example, so I'll follow up with the MITRE folks on that one after verifying that the data is at 16 kHz.

davidslater:
Confirmed. The underlying dataset is 16 kHz. (I saved the dataset output as a wavfile in both sampling rates, and it was clearly 16 kHz.) I'll follow up with MITRE on why SincNet uses 8 kHz.

davidslater:
@hkakitani The SincNet example uses a sample rate of 8 kHz:
https://github.com/twosixlabs/armory/blob/master/armory/baseline_models/pytorch/sincnet.py#L22
but the underlying LibriSpeech dataset is 16 kHz. Is this a bug, or is there something else going on here?

jesus-villalba:
Even if it is a bug in the sincnet, the results are ~90% acc, so it seems to have minor impact. I think because the fs is only used to calculate the window length and it is quite large.

davidslater:
The fs is also used internally by the SincNet model, but I haven't dug in deep enough to figure out what it's doing there. Based on the performance, I doubt it's doing any downsampling.

hkakitani:
The original SincNet code uses 8 kHz when using LibriSpeech: https://github.com/hkakitani/SincNet/blob/master/cfg/SincNet_Librispeech.cfg#L11  

The filters SincNet learns are all rectangular band-pass filters, and the sampling rate is used to calculate the high cutoff frequency of the band-pass filter: https://github.com/hkakitani/SincNet/blob/master/dnn_models.py#L140

PR #575, opened 2020-05-19
Made SincNet results deterministic

hkakitani:
Used the clip's length as the seed for np.random before getting a random chunk in order to get deterministic results for benign LibriSpeech


davidslater:
Fixes #570

PR #581, opened 2020-05-20
Optionally send evaluation results to a MongoDB instance

adamj26:
Adds a `--mongo` flag to the base scenario that takes the address of the MongoDB instance to send the results to.

Currently the exact same results are sent as are written to `.armory/outputs/`.

### DIFF
--- a/armory/scenarios/base.py
+++ b/armory/scenarios/base.py
@@ -139,18 +139,13 @@ if __name__ == "__main__":
     )
     parser.add_argument(
         "--no-docker",
-        dest="no_docker",
-        action="store_const",
-        const=True,
-        default=False,
+        action="store_true",
         help="Whether to use Docker or a local environment with armory run",
     )
     parser.add_argument(
         "--load-config-from-file",
         dest="from_file",
-        action="store_const",
-        const=True,
-        default=False,
+        action="store_true",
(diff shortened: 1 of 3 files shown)


armory/scenarios/base.py
```diff
@@ -139,22 +174,24 @@ def run_config(config_json, from_file=False):
...
-        const=True,
-        default=False,
+        action="store_true",
         help="If the config argument is a path instead of serialized JSON",
     )
+    parser.add_argument(
+        "--mongo",
+        dest="mongo_host",
+        default=None,
+        help="Send scenario results to a MongoDB instance at the given host",
```
seanpmorgan, line 190: Could we add an example of what a Mongo host URI with username/password looks like so it's clear how to use credentials. Alternatively we could add some notes in the documentation.

adamj26: I pushed an update to this text providing examples of how to specify the host. The code as currently written to connect to the host assumes that the MongoDB instance does not use authentication and thus does not handle usernames or passwords, which I can document either here or elsewhere, as you prefer

adamj26: Apologies; I just reread https://pymongo.readthedocs.io/en/stable/api/pymongo/mongo_client.html#pymongo.mongo_client.MongoClient and see that the code as written can take a username and password

PR #586, opened 2020-05-21
Poisoning scenario: update randomization

### LINKED ISSUE #583: Poisoning scenario - better randomization
Currently it is set such that an entire batch is either poisoned or not. This can lead to pretty high variance with the larger batch sizes once the poisoning % drops to the correct levels (10%).

ng390:
Closes #583

### DIFF
--- a/armory/scenarios/poisoning_gtsrb_scenario.py
+++ b/armory/scenarios/poisoning_gtsrb_scenario.py
@@ -97,6 +97,10 @@ class GTSRB(Scenario):
         if poison_dataset_flag:
             total_count = np.bincount(y_train_all)[src_class]
             poison_count = int(fraction_poisoned * total_count)
+            if poison_count == 0:
+                logger.warning(
+                    f"No poisons generated with fraction_poisoned {fraction_poisoned} for class {src_class}."
+                )
             src_indices = np.where(y_train_all == src_class)[0]
             poisoned_indices = np.random.choice(
                 src_indices, size=poison_count, replace=False
--- a/armory/scenarios/poisoning_gtsrb_scenario.py
+++ b/armory/scenarios/poisoning_gtsrb_scenario.py
@@ -23,7 +23,7 @@ from armory.scenarios.base import Scenario
 logger = logging.getLogger(__name__)
 
 
-def poison_batch(src_imgs, src_lbls, src, tgt, batch_size, attack, fraction_poisoned):
+def poison_dataset(src_imgs, src_lbls, src, tgt, ds_size, attack, poisoned_indices):
     # In this example, all images of "src" class have a trigger
     # added and re-labeled as "tgt" class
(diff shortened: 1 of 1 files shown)


davidslater:
I've been thinking more about this, and I think that there is still an additional stochastic issue to resolve. For poisoning, we shouldn't actually be flipping coins for each sample - we should be randomly selected a fixed-size subset of the data to poison.

Some classes have as few as 210 examples, which would be 21 for 10%, 10 for 5%, or 2 for 1% poisoned. However, if we randomize each selection, it's actually a binomial with standard deviations of 4.3, 3.2, and 1.4, respectively, which is quite large. Not as high variance as before, where you are doing it on full batches, but still higher than we want, I think. (Especially for 5% or 1% poisoned, where it can drastically change the number of poisoned examples).

I think what we should actually do is collect the counts of each class and then randomly select the percentage of instances from it. Something like:
```python
fraction = 0.1
ds = datasets.german_traffic_sign(split_type="train")
y = np.hstack([y_i for (x_i, y_i) in ds])
target_class = 18
total_count = np.bincount(y)[target_class]
poison_count = int(fraction * total_count)  # raise error if poison_count is 0
target_index = np.where(y == target_class)[0]
poisoned_index = np.random.choice(target_index, size=poison_count, replace=False)
# poison those
...
```

armory/scenarios/poisoning_gtsrb_scenario.py
```diff
@@ -88,14 +89,28 @@ def _evaluate(self, config: dict) -> dict:
...
-                    x_train, y_train, src_class, tgt_class, len(y_train), attack
-                )
             x_train_all.append(x_train)
             y_train_all.append(y_train)
         x_train_all = np.concatenate(x_train_all, axis=0)
         y_train_all = np.concatenate(y_train_all, axis=0)
+
+        if poison_dataset_flag:
+            total_count = np.bincount(y_train_all)[src_class]
+            poison_count = int(fraction_poisoned * total_count)
```
davidslater, line 99: Maybe we should consider issuing a log warning if poison_count == 0 ?

davidslater:
Have you measured the attack variance you're getting with this change?

ng390:
| Poison rate | Undefended attack success rate (std dev) | Validation accuracy (std dev) |
|-------------|------------------------------------------|-------------------------------|
| 1%          | 5.7% (7.1%)                              | 92.4% (3.0%)                  |
| 5%          | 73.4% (40.1%)                            | 90.4% (6.6%)                  |
| 10%         | 86.2% (11.4%)                            | 91.7% (2.9%)                  |

ng390:
There is not much variance in the attack success with the Activation Filtering defense (the attack success rate is near 0 even at 10% poisoning), but as seen above at 5% and 1% poisoning there is a fair bit of variance in the attack performance.

PR #588, opened 2020-05-21
Added art_experiemental JpegCompression wrapper to work with normaliz…

yusong-tan:
…ed inputs.  This will be used as baseline defense for image scenario.


davidslater:
Will merge if these tests pass

PR #590, opened 2020-05-21
--check mode for armory run

### LINKED ISSUE #382: Enable quick test mode to rapidly check whether scenario code runs

davidslater:
Fixes #382 

Can now quickly test code in a scenario by using the CLI flag `--check`. This works by essentially setting all loaded datasets to have a total size equal to a single batch (note that this means that if the batch size is larger, it will take longer). This required skipping the filtering defense portion of the poisoning scenario, but otherwise works fine on all scenarios.

To test:
```
python -m armory scenario_configs/mnist_baseline.json --check
python -m armory scenario_configs/mnist_baseline.json --check --no-docker
```

I also ensure that the `--debug` flag is carried into the scenario from the calling function, which partially addresses #500 .

### DIFF
--- a/armory/eval/evaluator.py
+++ b/armory/eval/evaluator.py
@@ -129,7 +129,7 @@ class Evaluator(object):
         jupyter=False,
         host_port=8888,
         command=None,
-        check=False,
+        check_run=False,
     ) -> None:
         if self.no_docker:
             if jupyter or interactive or command:
@@ -138,7 +138,7 @@ class Evaluator(object):
                 )
             runner = self.manager.start_armory_instance(envs=self.extra_env_vars)
             try:
-                self._run_config(runner, check=check)
+                self._run_config(runner, check_run=check_run)
             except KeyboardInterrupt:
                 logger.warning("Keyboard interrupt caught")
             finally:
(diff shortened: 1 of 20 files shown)


armory/eval/evaluator.py
```diff
@@ -124,7 +124,12 @@ def _cleanup(self):
             pass
 
     def run(
-        self, interactive=False, jupyter=False, host_port=8888, command=None
+        self,
+        interactive=False,
+        jupyter=False,
+        host_port=8888,
+        command=None,
+        check=False,
```
seanpmorgan, line 132: Perhaps `dry-run`? Feels a bit more informative than `check` IMO.

armory/scenarios/audio_classification.py
```diff
@@ -42,13 +42,16 @@ def _evaluate(self, config: dict) -> dict:
                 f"Fitting model {model_config['module']}.{model_config['name']}..."
             )
             fit_kwargs = model_config["fit_kwargs"]
+            if self.check:
+                fit_kwargs["nb_epochs"] = 1
```
seanpmorgan, line 46: So if users submit their own scenarios (which is likely) will be requiring this as part of their _evaluate method? Maybe we should consider the `check` flag modifying the configuration dictionary?

armory/eval/evaluator.py
davidslater: Using `check_run`

PR #595, opened 2020-05-27
Increase default SHM size

### LINKED ISSUE #535: Increase shm_size in Docker container
The `shm_size` is 64M by default. However, the multiprocessing data pipeline mentioned in #520 requires much larger `shm_size`.

```console
$ docker exec -it -u 1000:1000 183ec408cb bash

183ec408cb8d:/workspace$ df -h
Filesystem      Size  Used Avail Use% Mounted on
overlay         1.8T  566G  1.2T  33% /
tmpfs            64M     0   64M   0% /dev
tmpfs            32G     0   32G   0% /sys/fs/cgroup
shm              64M     0   64M   0% /dev/shm
```

A potential fix could be

```diff
diff --git a/armory/docker/management.py b/armory/docker/management.py
index 8e665e4..e3edc8f 100644
--- a/armory/docker/management.py
+++ b/armory/docker/management.py
@@ -53,6 +53,7 @@ class ArmoryInstance(object):
                 host_output_dir: {"bind": docker_paths.output_dir, "mode": "rw"},
                 host_tmp_dir: {"bind": docker_paths.tmp_dir, "mode": "rw"},
             },
+            "shm_size": '16G',
         }
         if ports is not None:
             container_args["ports"] = ports
```

With that change, we have /dev/shm in size of 16G:

```console
$ docker exec -it -u 1000:1000 c731bd7106 bash

c731bd71068c:/workspace$ df -h
Filesystem      Size  Used Avail Use% Mounted on
overlay         1.8T  566G  1.2T  33% /
tmpfs            64M     0   64M   0% /dev
tmpfs            32G     0   32G   0% /sys/fs/cgroup
shm              16G     0   16G   0% /dev/shm
```

seanpmorgan:
So went through a couple of iterations on deciding if we wanted to introduce docker configs to the scenario configurations. Ideally, we don't since they require special parsing from ECS task conversion. Looking into /dev/shm and the tempfs spec it doesn't appear to have negative consequences to hard code this as a higher value:
https://unix.stackexchange.com/questions/259823/if-tmpfs-has-bigger-size-than-your-ram-how-much-ram-does-it-use-are-applicatio

I tried setting this to 64GB (2x my laptops memory) and everything worked fine so shouldn't be an issue for systems < 16GB RAM. This guide also recommends increasing shm by default:
https://www.pugetsystems.com/labs/hpc/How-To-Setup-NVIDIA-Docker-and-NGC-Registry-on-your-Workstation---Part-5-Docker-Performance-and-Resource-Tuning-1119/

@adamj26 should this get merged we'll need to hardcode this into the ECS task conversion.

Closes #535

### DIFF
--- a/armory/docker/management.py
+++ b/armory/docker/management.py
@@ -50,6 +50,7 @@ class ArmoryInstance(object):
                 host_output_dir: {"bind": docker_paths.output_dir, "mode": "rw"},
                 host_tmp_dir: {"bind": docker_paths.tmp_dir, "mode": "rw"},
             },
+            "shm_size": "16G",
         }
         if ports is not None:
             container_args["ports"] = ports


davidslater:
That seems reasonable to me. This is the only docker config we've really been asked for, and they just wanted this value increased (it didn't look like they needed to configure it).

PR #597, opened 2020-05-27
PGD fix (also applies to other attacks that need ground truth labels)

### LINKED ISSUE #578: PGD attack generates incorrect values
This is due to a couple of issues: attack generation does not take the ground truth value into account, and the attack metrics are computed regardless of whether the initial classification was correct.

The fix should make use of ART's full attack.generate functionality

davidslater:
Fixes #578

### DIFF
--- a/docs/configuration_files.md
+++ b/docs/configuration_files.md
@@ -16,6 +16,9 @@ All configuration files are verified against the jsonschema definition at run ti
     kwargs: [Object] Keyword arguments to pass to attack instatiation    
     module: [String] Python module to load attack from 
     name: [String] Name of the attack class to be instatiated
+    use_label: [Bool] Default: False. Wheather attack should use the true label when 
+          attacking the model. Without this, it is not possible to drive the accuracy 
+          down to 0% when the model has misclassifications.
   }
 `dataset`: [Object]
   {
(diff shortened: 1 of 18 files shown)


davidslater:
Fixes #578

davidslater:
This enables the attacker to use the true label when attacking the model. Without this, it is not possible to drive the accuracy down to 0% when the model has misclassifications. By setting the optional attack config "use_label" to True, this causes the attack to work properly on misclassifications.

(An alternative would be to simply ignore misclassifications in the attack generation, though this would require computing the benign and attack predictions in the same loop.)

scenario_configs/resisc45_baseline_densenet121.json
```diff
@@ -4,10 +4,14 @@
...
-            "eps": 0.2
+            "eps": 0.031,
+            "eps_step": 0.007,
+            "max_iter": 20,
+            "num_random_init": 1
         },
         "module": "art.attacks",
-        "name": "FastGradientMethod"
+        "name": "ProjectedGradientDescent",
+        "use_label": true
```
davidslater, line 14: I think that just modifying the cifar example should be sufficient.

docs/configuration_files.md
```diff
@@ -16,6 +16,9 @@ All configuration files are verified against the jsonschema definition at run ti
     kwargs: [Object] Keyword arguments to pass to attack instatiation    
     module: [String] Python module to load attack from 
     name: [String] Name of the attack class to be instatiated
+    use_label: [Bool] Default: False. Wheather attack should use the true label when
```
ng390, line 19:
```suggestion
    use_label: [Bool] Default: False. Whether attack should use the true label when 
```

PR #607, opened 2020-06-02
SincNet and Adversarial LibriSpeech update

hkakitani:
1. Updated the adversarial LibriSpeech dataset with targeted PGD attack. Updated dataset found here: https://git.codev.mitre.org/projects/GARD/repos/gard-scenarios/browse/scenarios_phase_1/sincnet_librispeech/UP-and-pgd-sincnet-librispeech

2. SincNet preprocessing and audio classification scenario updated to use multiple cuts from entire audio rather than just using one random cut.


yusong-tan:
Just want to comment on performers potentially need to access the whole clip.  This update performs inference on consecutive cuts of the same clip as an (n_variable_cuts, 3000) shaped input. A defense could work with 2D as is or by reshaping to 1D.  During training, a single random cut from each clip is used instead.  If performers need to access all the cuts from a clip during training, then the scenario will need to be slightly modified.

davidslater:
@hkakitani there appears to be some issues with the dataset here. The values on the clean and adversarial datasets are in completely different ranges. It appears that the clean values are in the range of 16 bit ints, where the adversarial values are in the range of 32 bit ints.

Since the adversarial examples may have different maxes, normalization is not applied evenly to both, which results in very large epsilon values. (And if normalization is not used, the epsilon values are extremely large.)

hkakitani:
I've modified the dataset so that all of the data is in the int16 range. The updated datasets and checksum file are in the codev: https://git.codev.mitre.org/projects/GARD/repos/gard-scenarios/browse/scenarios_phase_1/sincnet_librispeech/UP-and-pgd-sincnet-librispeech

davidslater:
@hkakitani I'm still seeing a number of issues.

1) The first is that some of the pairs return an epsilon (inf norm) of 2 for the normalized samples, which is the max possible. My guess is that this is due to some sort of casting to int issue (possibly an integer overflow).
2) Second is that the epsilon values (when they don't return ~2), appear to be around 0.03, not 0.3. Is the documentation wrong on your PGD parameters?
3) Finally, while being in the ballpark of 0.03, they can vary by more than 0.001, causing epsilon values beyond the desired bound. This is likely due to different normalizations on the clean and adversarial samples (due to different max values). I don't know of a simple way to fix this one, but the other two should be fixed.

yusong-tan:
@davidslater 

For #1, I believe the attack on audio uses L-2 norm, so the L-inf being 2 isn't very informative - it's quite possible a sample of clean audio is +1 while the corresponding sample in the adversarial being -1. Now that I've taken a look at the PGD documentation, it didn't specify L-2 norm being used.

As for #2, is 0.03 the L-inf norm?

davidslater:
#2 is 0.03 for the L-inf norm. For the L2 norm, it is always quite large (L2 on left column, Linf on right column):
```
>>> ds = adversarial_datasets.librispeech_adversarial(cache_dataset=False, adversarial_key="adversarial_univperturbation")
>>> for i in range(10):
...     (x1, x2), y = next(ds)
...     x1normal, x2normal = [x / np.max(np.abs(x)) for x in (x1[0], x2[0])]
...     print(np.linalg.norm(x1normal - x2normal, ord=2), np.linalg.norm(x1normal - x2normal, ord=np.inf))
... 
1.424759002895783 0.028721278659769478
2.592007132509364 2.0
7.08017683317084 1.9989296226919988
2.588001927760749 2.0
2.512648520987732 1.9884096870615506
2.5251349494635345 2.0
1.660504767441881 0.031156540479349792
1.312103057736463 0.03659722169495483
1.5091728891879825 0.022149936342432265
1.5322410553674601 0.030690638741298182
```

From the PGD code, Linf is the default:
https://github.com/IBM/adversarial-robustness-toolbox/blob/master/art/attacks/evasion/projected_gradient_descent/projected_gradient_descent.py#L76

davidslater:
Similarly for the untargeted case:
```
>>> ds = adversarial_datasets.librispeech_adversarial(cache_dataset=False, adversarial_key="adversarial_perturbation")
>>> for i in range(10):
...     (x1, x2), y = next(ds)
...     x1normal, x2normal = [x / np.max(np.abs(x)) for x in (x1[0], x2[0])]
...     print(np.linalg.norm(x1normal - x2normal, ord=2), np.linalg.norm(x1normal - x2normal, ord=np.inf))
... 
1.3107720126395492 0.04144935907111087
1.2332842100707848 0.038548400353303314
1.0785662430778182 0.03885992375734901
1.1031459445926948 0.029743328312207214
2.3157075516263874 1.9996025680220115
1.0998781209052564 0.043776419909990816
0.8850355560552551 0.03081677013324028
2.2297290154275506 1.9952906638940706
2.360328473507063 2.0
1.0102036365903713 0.03290565379901905
```

yusong-tan:
I'm not sure how sensitive the results are to the normalization.  I know when I ran some tests in the past using PGD to generate L-2 norm with eps=0.3, the perturbation is constrained by 0.3. But this calculated in the same loop when the adversarial example is generated i.e., is where the clean and adv have the same normalization.

I suspected the normalization would make distance metric measures uninformative for a single adversarial dataset, but a trend should appear when comparing across adversarial datasets with different eps.

davidslater:
Screen Shot 2020-06-15 at 4 44 20 PM

Here is the clean (orange) and adversarial (blue), where we're getting an Linf distance of 2.0. I think this is probably due to integer underflow, as it only happens when the clean is near -1.0 (or 1.0).

davidslater:
For the other issue, are you generating the attacks independly on a cut-by-cut basis? (So the L2 distance is actually a sum of a number of cuts?) Or is there some other pre-processing that the norm is calculated after?

hkakitani:
Yes, we're generating adversarial examples for each cut and concatenating all of the adversarial examples together to create the full length adversarial audio.

hkakitani:
Just updated the dataset and checksum file on CoDev: https://git.codev.mitre.org/projects/GARD/repos/gard-scenarios/browse/scenarios_phase_1/sincnet_librispeech/UP-and-pgd-sincnet-librispeech

I had forgotten to clip the original adversarial audio to the range [-1.0, 1.0], and that was what was causing the underflow/overflow when converting to int16.

davidslater:
Looks good! I've split off from this branch and made a PR under #625

davidslater:
#625 was merged.

PR #608, opened 2020-06-03
Improving UCF-101 inference speed by 2.25x

mzweilin:
Let the MARS model consume a batch of 16 consecutive frames via PyTorchClassifier in ART, bringing 2.25x speedup (1,910 seconds Vs. 848 seconds on 3,792 videos).

### DIFF
--- a/armory/scenarios/video_ucf101_scenario.py
+++ b/armory/scenarios/video_ucf101_scenario.py
@@ -6,6 +6,7 @@ Scenario Contributor: MITRE Corporation
 
 import logging
 
+import torch
 import numpy as np
 from tqdm import tqdm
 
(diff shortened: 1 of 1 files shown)


seanpmorgan:
Hi @mzweilin thanks for the PR! So unforuntely this scenario needs to work for all frameworks so importing torch in this particular file would cause issues. Leaving this open though because we'll need to figure out how to incorporate this into the structure.

mzweilin:
The torch.no_grad() statement could be added to ART.

https://github.com/IBM/adversarial-robustness-toolbox/blob/3cfa6de196cb32a3efafab2ff6bbf44247c9ddbd/art/classifiers/pytorch.py#L147

davidslater:
I made an ART issue for this: https://github.com/IBM/adversarial-robustness-toolbox/issues/448

If they accept the issue, we'll make a PR there instead.

PR #613, opened 2020-06-08
Add warning for evaluating model without weights nor training

seanpmorgan:
Only question would be whether this should instead raise an error?

### DIFF
--- a/armory/utils/config_loading.py
+++ b/armory/utils/config_loading.py
@@ -61,7 +61,7 @@ def load_model(model_config):
     )
     if not isinstance(model, Classifier):
         raise TypeError(f"{model} is not an instance of {Classifier}")
-    if weights_file is None and not model_config["fit"]:
+    if not weights_file and not model_config["fit"]:
         logger.warning(
             "You're attempting to evaluate an unfitted model with no "
             "pre-trained weights!"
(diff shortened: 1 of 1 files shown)


armory/utils/config_loading.py
```diff
@@ -58,6 +61,11 @@ def load_model(model_config):
     )
     if not isinstance(model, Classifier):
         raise TypeError(f"{model} is not an instance of {Classifier}")
+    if weights_file is None and not model_config["fit"]:
```
davidslater, line 64: Suggestion - make this:
```
if weights_file and not model_config["fit"]:
```
Right now, if weights_file == "", it successfully runs but does not output a warning. See:
https://github.com/twosixlabs/armory/blob/master/armory/baseline_models/keras/mnist.py#L55

davidslater:
I would prefer not to raise an error. Sometimes it's nice to ensure inference runs before training.

armory/utils/config_loading.py
seanpmorgan: Thanks. Updated to `if not weights_file and not model_config["fit"]:` is that what you meant?

davidslater: Yes, that's what I meant.

PR #616, opened 2020-06-08
Poisoning scenario including variable fractions of train poisoned samples

### LINKED ISSUE #355: Create Poisoning Attacks which download generated poisoned examples

DavidWillmes:
Closes #355


armory/scenarios/poisoning_gtsrb_scenario.py
```diff
@@ -224,18 +270,14 @@ def _evaluate(self, config: dict) -> dict:
...
-
                 y_pred_targeted = y_pred[y_test == src_class]
                 if not len(y_pred_targeted):
                     continue
                 targeted_test_metric.append(
                     [tgt_class] * len(y_pred_targeted), y_pred_targeted
                 )
-            results["test_accuracy"] = test_metric.mean()
-            results["targeted_misclassification_accuracy"] = targeted_test_metric.mean()
-            logger.info(f"Test accuracy: {test_metric.mean():.2%}")
```
davidslater, line 237: We should keep test accuracy, as this should not be greatly impacted by targeted attacks (and is a strong detection signal if it is).

armory/scenarios/poisoning_gtsrb_scenario.py
```diff
@@ -200,7 +228,27 @@ def _evaluate(self, config: dict) -> dict:
             "validation_accuracy_targeted_class": target_class_benign_metric.mean(),
         }
 
-        if poison_dataset_flag:
```
davidslater, line 203: We have lost this functionality (ability to set poisoning to False for a benign run through) - can you add it back?

DavidWillmes: Sure, this should be easy enough to add back in.

PR #617, opened 2020-06-09
Enable multiple external repos

### LINKED ISSUE #496: Enable multiple external repositories to be pulled
In order to prevent breaking API change we can make it a blackslash separated string or some other invalid git repo character.

seanpmorgan:
Closes #496

### DIFF
--- a/armory/utils/external_repo.py
+++ b/armory/utils/external_repo.py
@@ -42,6 +42,7 @@ def download_and_extract_repo(
     repositories project root will be added to the sys path.
 
     Private repositories require an `ARMORY_GITHUB_TOKEN` environment variable.
+    :param external_repo_name: String name of "organization/repo-name" or "organization/repo-name@branch"
     """
     verify_ssl = get_verify_ssl()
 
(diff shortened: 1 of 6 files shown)


armory/utils/external_repo.py
```diff
@@ -24,7 +42,6 @@ def download_and_extract_repo(
     repositories project root will be added to the sys path.
 
     Private repositories require an `ARMORY_GITHUB_TOKEN` environment variable.
-    :param external_repo_name: String name of "organization/repo-name"
```
davidslater, line 27: Reason why this was deleted?

seanpmorgan: Thought it was misleading since it should say:
```
String name of "organization/repo-name" or "organization/repo-name@branch"
```

Then I realized we hadn't documented the other params so wasn't sure it was worth updating since that information is available in the documentation. I can update it if you'd prefer

davidslater: Let's just update it.

seanpmorgan: Done. I need to update https://github.com/twosixlabs/armory/blob/master/docs/external_repos.md as well though

PR #618, opened 2020-06-09
Initial PyTorch dataloaders from TFRecords

### LINKED ISSUE #455: Add pytorch Datasets loader to datasets
This requires using PyTorch DataLoader to read directly from the TFDS datasets. There may be some issues making this performant, as PyTorch uses very different methods of loading the data.
See this thread: https://discuss.pytorch.org/t/read-dataset-from-tfrecord-format/16409

Most of what I have found focuses on reading TFRecord files (which are protobufs), e.g., https://github.com/vahidk/tfrecord
but does not leverage the larger dataset_info.json and related txt/json files that would be necessary for parsing an entire dataset automatically.

seanpmorgan:
Closes #455 

Works for cifar10 and mnist. Some design decisions we'd need to make and then we can expand this to other datasets.

### DIFF
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -34,6 +34,7 @@ from armory.data.librispeech import librispeech_dev_clean_split  # noqa: F401
 from armory.data.resisc45 import resisc45_split  # noqa: F401
 from armory.data.german_traffic_sign import german_traffic_sign as gtsrb  # noqa: F401
 from armory.data.digit import digit as digit_tfds  # noqa: F401
+from armory.data.pytorch_loaders import ImageTFRecordDataSet
 
 
 os.environ["KMP_WARNINGS"] = "0"
@@ -47,7 +48,7 @@ CACHED_CHECKSUMS_DIR = os.path.join(os.path.dirname(__file__), "cached_s3_checks
 
 class ArmoryDataGenerator(DataGenerator):
     """
-    Returns batches of data.
+    Returns batches of numpy data.
 
     variable_length - if True, returns a 1D object array of arrays for x.
     """
@@ -198,12 +199,6 @@ def _generator_from_tfds(
             f"Dataset framework field {framework} must be one of {supported_frameworks}"
         )
 
-    if framework == "pytorch":
-        raise NotImplementedError(
-            "PyTorch native dataloaders are not yet supported. "
-            "See issue https://github.com/twosixlabs/armory/issues/455"
-        )
-
     if not dataset_dir:
         dataset_dir = paths.runtime_paths().dataset_dir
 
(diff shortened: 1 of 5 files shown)


armory/data/datasets.py
```diff
@@ -212,58 +206,73 @@ def _generator_from_tfds(
...
-    if shuffle_files:
-        ds = ds.shuffle(batch_size * 10, reshuffle_each_iteration=True)
-    if variable_length and batch_size > 1:
-        ds = ds.batch(1, drop_remainder=False)
+        if not shuffle_files:
+            raise ValueError("PyTorch DataLoaders from Armory are shuffled by default")
+        ds_name, ds_version = dataset_name.split(":")
+
+        # TODO: Make this configurable. Possibly a mapping from dataset_name???
+        ds = ImageTFRecordDataSet(ds_name, ds_version, split_type)
```
seanpmorgan, line 217: It'd be nice if we could parse the relevant JSON files from each dataset and have a generic dataset loader here, but it may be easier to just map them to respective loaders.

armory/data/pytorch_loaders.py
```diff
@@ -0,0 +1,50 @@
...
+        return np.array(Image.open(BytesIO(bytes_img[0])))
+
+    def preprocess(self, x):
+        label, image = x
+        image = self.decode_image(image)
+
+        if image.ndim == 2:
+            image = np.expand_dims(image, axis=-1)
+
+        # TODO: permute the shape to channels first as is typical for PyTorch???
```
seanpmorgan, line 44: Need a decision on this

tests/test_pytorch/test_framework_dataset.py
```diff
@@ -2,20 +2,47 @@
...
-            epochs=1,
-            batch_size=16,
-            dataset_dir=DATASET_DIR,
-            framework="pytorch",
-        )
+def test_pytorch_generator_cifar10():
+    batch_size = 16
+    dataset = datasets.cifar10(
+        split_type="train",
+        epochs=1,
```
seanpmorgan, line 17: Need to verify how this works for multiple epochs as well

armory/data/pytorch_loaders.py
davidslater: This is tricky, as we would then want to permute the inference time channels when using numpy mode in the attack. Hmmm.

armory/data/datasets.py
```diff
@@ -212,58 +206,73 @@ def _generator_from_tfds(
...
+
+        generator = torch.utils.data.DataLoader(
+            ds, batch_size=batch_size, num_workers=2
+        )
+        return generator
```
davidslater, line 222: It would be nice to make `num_workers` configurable.

armory/data/datasets.py
davidslater: For now, let's just hard map them in.

armory/data/datasets.py
```diff
@@ -212,58 +206,73 @@ def _generator_from_tfds(
...
-        ds = ds.batch(1, drop_remainder=False)
+        if not shuffle_files:
+            raise ValueError("PyTorch DataLoaders from Armory are shuffled by default")
+        ds_name, ds_version = dataset_name.split(":")
+
+        # TODO: Make this configurable. Possibly a mapping from dataset_name???
+        ds = ImageTFRecordDataSet(ds_name, ds_version, split_type)
+
+        generator = torch.utils.data.DataLoader(
+            ds, batch_size=batch_size, num_workers=2
```
davidslater, line 220: This needs to have a `worker_init_fn` defined to handle multi-process data loading. Currently, each worker will return duplicate batches.

See the second example in: https://pytorch.org/docs/stable/data.html#torch.utils.data.IterableDataset

Example:
```
dataset = datasets.cifar10(
         split_type="train",
         epochs=1,
         batch_size=batch_size,
         dataset_dir=DATASET_DIR,
         framework="pytorch",
     )
iter_ds = iter(dataset)
x1, y1 = next(iter_ds)
x2, y2 = next(iter_ds)
print((x1 == x2).all())
# tensor(True)
print((y1 == y2).all())
# tensor(True)
```

armory/data/pytorch_loaders.py
davidslater: Let's go ahead and switch to channels first, as channels last is still experimental in pytorch.

tests/test_pytorch/test_framework_dataset.py
davidslater: It only generates a single epoch, regardless of what `epochs` is.

armory/data/datasets.py
```diff
@@ -212,58 +206,73 @@ def _generator_from_tfds(
...
-        ds = ds.map(lambda_map)
-
-    ds = ds.repeat(epochs)
-    if shuffle_files:
-        ds = ds.shuffle(batch_size * 10, reshuffle_each_iteration=True)
-    if variable_length and batch_size > 1:
-        ds = ds.batch(1, drop_remainder=False)
+        if not shuffle_files:
+            raise ValueError("PyTorch DataLoaders from Armory are shuffled by default")
+        ds_name, ds_version = dataset_name.split(":")
```
davidslater, line 214: We should probably document that this shuffling is due to dareblopy, not due to the defaults of PyTorch iter dataloaders.

armory/data/datasets.py
davidslater: If we are mapping, should we move all of the NotImplementedError throwing into here, instead of replicating across datasets?

tests/test_pytorch/test_framework_dataset.py
davidslater: You can call `iter(dataset)` again to start over, but the iteration is in the same order as the initial shuffle (it is not reshuffled).

armory/data/pytorch_loaders.py
seanpmorgan: Per discussion we'll be leaving it in channels last format so that pytorch baseline models take the same inputs in numpy and torch

ng390:
LGTM

PR #625, opened 2020-06-11
Librispeech variable length input

### LINKED ISSUE #591: Odd Audio adquisition pipeline
Hi,
This is about the audio preprocessing function in the Sincnet example.
In MITRE's slides, it indicates that adv examples would be 5sec chunks, while librispeech are variable length recordings with duration>= 1.5 sex.

However, in the sincnet example, you read the full utterance and then you just randomly take a single 375/2=187.5 milliseconds frame  and discard the rest. That is less than a syllable (~250 msecs).  

I assume that this odd pipeline is the result of a misunderstanding reading Mirco's code. He trains his network in a frame by frame basis and the forward function of his model only takes single frame batches.
But when he evaluates, he evaluates all the frames of the recording with a 10msecs shift. See lines 290-309 in https://github.com/hkakitani/SincNet/blob/master/speaker_id.py
He gets frame level decision and combine them to obtain an utterance level decision in line 316
[val,best_class]=torch.max(torch.sum(pout,dim=0),0)

It also puzzles me, why MITRE hardcoded 8 kHz and 375 for librispeech considering that, in Mirco's Timit example, he used 200ms window and 16kHz.

In summary, for evaluating our own models, we need to know if we are allowed to use the full utterances, or if we should use 5secs out of the full utt. as we inferred from MITRE's slides, or any other duration. This should be documented somewhere. From my point of view, using 187 msecs is not realistic. 
Thanks

davidslater:
Fixes #591 

Can be checked with:
```
python -m armory run scenario_configs/librispeech_baseline_sincnet.json --check
```

### DIFF
--- a/scenario_configs/librispeech_baseline_sincnet.json
+++ b/scenario_configs/librispeech_baseline_sincnet.json
@@ -28,7 +28,7 @@
         ]
     },
     "model": {
-        "fit": true,
+        "fit": false,
         "fit_kwargs": {
             "nb_epochs": 20000
         },
--- a/scenario_configs/librispeech_baseline_sincnet.json
+++ b/scenario_configs/librispeech_baseline_sincnet.json
@@ -1,6 +1,8 @@
 {
     "_description": "Librispeech_dev_clean raw audio classification, contributed by MITRE Corporation",
-    "adhoc": null,
+    "adhoc": {
+        "fit_batch_size": 16
+    },
     "attack": {
         "knowledge": "white",
         "kwargs": {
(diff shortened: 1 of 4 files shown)


scenario_configs/librispeech_baseline_sincnet.json
```diff
@@ -26,12 +28,12 @@
         ]
     },
     "model": {
-        "fit": false,
+        "fit": true,
```
seanpmorgan, line 31:
```suggestion
        "fit": false,
```

PR #635, opened 2020-06-17
Hkakitani dataset update

davidslater:


davidslater:
Should be able to test with:
```
python
from armory import paths
paths.set_mode("host")
from armory.data import adversarial_datasets
ds = adversarial_datasets.librispeech_adversarial(cache_dataset=False)
```
and 
```
ds = adversarial_datasets.librispeech_adversarial(cache_dataset=True)
```

You'll need to delete the `.armory/datasets/librispeech_adversarial/1.1.0` directory after testing with cache_dataset=False; otherwise, it won't pull the cached dataset from s3 (it will just use your local one).

PR #637, opened 2020-06-17
Fix gpu memory leak to 0.9

seanpmorgan:


davidslater:
Should we update version numbers here?

PR #644, opened 2020-06-18
support for art 1.3 APIs

davidslater:
Changes API calls to work with both ART 1.2 and 1.3. There are couple of try/except blocks for breaking changes, but the rest of the changes should be backwards compatible.

This *does not* update the docker images or build scripts.

This was due to at least one performer needing ART 1.3 to get their scenario to run. I am going to recommend that they build a custom Dockerfile with ART 1.3 instead of switching the Dockerfiles for everyone at this point.


davidslater:
The deletion (armory/art_experimental/attacks/poisoning_attack_backdoor.py) was due to it not even being compatible with ART 1.2. @ng390 was this material that we integrated into ART?

ng390:
> The deletion (armory/art_experimental/attacks/poisoning_attack_backdoor.py) was due to it not even being compatible with ART 1.2. @ng390 was this material that we integrated into ART?

Good catch. The same functionality (with more flexibility with pattern choice) was independently implemented in ART 1.2, so that the art_experimental code became somewhat redundant.

davidslater:
@davidslater needs to fix the json files to match

davidslater:
Should be superceded by #648

PR #648, opened 2020-06-19
Art 1.3 Dependency updates

### LINKED ISSUE #479: Consider bumping to PyTorch 1.5
I think we want to avoid chasing each release of TF and PyTorch, but there is newly added support for "Channel Last" memory format in PyTorch 1.5:
https://pytorch.org/blog/pytorch-1-dot-5-released-with-new-and-updated-apis/#channels-last-memory-format-for-computer-vision-models-experimental

This could simplify our dataset generators because both frameworks would have support for the data format we currently load. There would no longer be a need to permute the shapes of data in models as is done here:
https://github.com/twosixlabs/armory/blob/master/armory/baseline_models/pytorch/cifar.py#L30

It is true that we are actively looking to build PyTorch data generators https://github.com/twosixlabs/armory/issues/455 that could handle the dimension swap, but since ART is slowly adding support for torch Tensors, it's likely we'll need NumPy support for some attacks/defenses into the foreseeable future.

### LINKED ISSUE #585: Pillow 7.1.2 in pytorch docker image
I pulled the latest docker images and noticed Pillow 7.1.2 was installed rather than Pillow 7.0.0 as specified in the Dockerfile: https://github.com/twosixlabs/armory/blob/b81d681b638e3fe4e7dab89706a88867dd8b41e0/docker/Dockerfile#L25

Which version of Pillow should we expect?

### LINKED ISSUE #649: Support ART 1.3

ng390:
Fixes #649
Fixes #585 
Fixes #479


docker/pytorch/Dockerfile
```diff
@@ -1,18 +1,18 @@
-########## PyTorch 1.4 Base #################
+########## PyTorch 1.5 Base #################
```
davidslater, line 1: Let's remove the minor version numbers from these and just have PyTorch 1 Base, TF 2 Base, etc.

host-requirements.txt
```diff
@@ -12,17 +12,18 @@ apache-beam==2.18.0
 pytest==5.3.5
 torchvision==0.5.0
 ffmpeg==4.2
+pymongo==3.1.0
```
davidslater, line 15: You'll want to update Pillow, ART version, etc., above

docker/Dockerfile
```diff
@@ -21,9 +21,9 @@ RUN /opt/conda/bin/pip install --no-cache-dir \
     tensorflow-datasets==2.0.0 \
     jupyterlab==1.2.6 \
     boto3==1.11.13 \
-    adversarial-robustness-toolbox==1.2.0 \
-    Pillow==7.0.0 \
-    pydub==0.23.1 \
+    adversarial-robustness-toolbox==1.3.0 \
+    Pillow==7.1.2 \
+    pydub==0.24.1 \
     apache-beam==2.18.0 \
```
davidslater, line 27: Let's update the following (and also in the host-requirements.txt):
apache-beam==2.22.0
dill==0.3.1.1

Also, let's fix ffmpeg to 4.2.3 (from 4.2), as that seems to clean up the conda process a bit.

davidslater:
Once ART 1.3.1 is released, will need to update docker/Dockerfile as follows:
Remove lines 34-38
Update ART from 1.3.0 to 1.3.1

And update host-requirements.txt by updating ART from 1.3.0 to 1.3.1

PR #655, opened 2020-06-25
Label adversarial examples for resisic45 adversarial datasets 

grobertson-ext:
Edited armory/data/adversarial/resisc45_densenet121_univpatch_and_univperturbation_adversarial_224x224.py and added labels for targeted attacks. Also updated label_map function as in the ucf part of adversarial_datasets.py. Getting error from _generator_from_tfds in datasets.py related to label.


grobertson-ext:
@ng390 @seanpmorgan Do you have any ideas about this error I'm getting? It's this from `_generator_from_tfds` in datasets.py. 

>  /workspace/armory/data/datasets.py:257 None * 
>        ds = ds.map(lambda x: (x[x_key], x[y_key])) 
>    KeyError: 'label'  
> 

(The last 2 commits are the only relevant ones. Only changes to L111-117 and 170-174 in `armory/data/adversarial/resisc45_densenet121_univpatch_and_univperturbation_adversarial_224x224.py` + updating the `resisc adversarial` section of `adversarial_datasets.py` to match the `ucf adversarial` `label_map` function)

armory/data/adversarial/resisc45_densenet121_univpatch_and_univperturbation_adversarial_224x224.py
```diff
@@ -107,7 +108,13 @@ def _info(self):
                             shape=[224, 224, 3], dtype=tf.uint8, encoding_format="png"
                         ),
                     },
-                    "label": tfds.features.ClassLabel(names=_LABELS),
+                    "labels": {
```
davidslater, line 111: The key here was changed from "label" to "labels". This requires a change below.

armory/data/adversarial_datasets.py
```diff
@@ -138,7 +150,7 @@ def resisc45_adversarial_224x224(
         variable_length=False,
         cache_dataset=cache_dataset,
         framework=framework,
-        lambda_map=lambda x, y: ((x[clean_key], x[adversarial_key]), y),
+        lambda_map=lambda_map,
     )
```
davidslater, line 154: ```supervised_xy_keys=("images", "label"),``` should be changed to ```supervised_xy_keys=("images", "labels"),```

grobertson-ext: Thanks, David. Still getting the error: 
`tensorflow.python.framework.errors_impl.InvalidArgumentError: {{function_node __inference_Dataset_map_ExampleParser.parse_example_20107}} Feature: labels/adversarial_univpatch (data type: int64) is required but could not be found.
`

ng390: Please set [cache_dataset](https://github.com/twosixlabs/armory/blob/b8b9c35c94ac2b47fb07100de3fceaca003b1213/armory/data/adversarial_datasets.py#L110) to be false. Right now the workflow requires us to update that manually when a dataset is changed. If there was a different dataset with 1.0.1 that you ran locally, you may need to clear out the related directories in your datasets directory as well. It may actually make sense to increment the version if the structure of the dataset is changing.

grobertson-ext: Thanks -- still same error. It's not a different dataset; just an update to the dataset builder file.

davidslater: True, but that causes the processed dataset to be different - so you'll probably need to increment the version number.

grobertson-ext: That worked! Thanks for picking up on that and explaining it!

PR #657, opened 2020-06-25
maybe download weights from s3

### LINKED ISSUE #646: Maybe download from s3 inside config_loading
This will download the file if possible, but will not update pathing.

davidslater:
This is done in model_loading, but should be backwards compatible (as long as their weights are used and in the saved_model_dir) since I do not update to pass in paths.

Fixes #646


davidslater:
Note that I added the possibility of model_weights being a list so that it works with later extensions, but this is unnecessary.

davidslater:
This is just a couple of lines, and needs to wait until we introduce breaking changes. Closing.

PR #660, opened 2020-06-26
Poisoning scenario -- reproducible splits

ng390:
* Uses a single scenario input ("split_id") to produce reproducible poisoned dataset.
* Some additional steps to remove randomness introduced by performance optimizations from run to run.
* Does not yet allow for reproducible training results, even on CPU.


ng390:
Per the Keras docs, the environmental variable [PYTHONHASHSEED](https://keras.io/getting_started/faq/#how-can-i-obtain-reproducible-results-using-keras-during-development) should be set to facilitate reproducibility on the CPU **before** python launches. Is this something we would want to set globally, or as a sysconfig option that is specific to the poisoning scenario?

lcadalzo:
There are two `if attack_type == "preloaded"` blocks in short succession. [Here](https://github.com/ng390/armory/blob/poison-evaluation-repr/armory/scenarios/poisoning_gtsrb_scenario.py#L109) and [here](https://github.com/ng390/armory/blob/poison-evaluation-repr/armory/scenarios/poisoning_gtsrb_scenario.py#L135). I think the code would be a bit more readable if the logic in these blocks were condensed into just one `if attack_type == "preloaded"` block.

lcadalzo:
Is there a reason we ignore the preprocessing fn returned by `load_model` [here](https://github.com/ng390/armory/blob/poison-evaluation-repr/armory/scenarios/poisoning_gtsrb_scenario.py#L193)? Is it assured that a classifer used for defense won't/shouldn't preprocess inputs?

lcadalzo:
I think a comment or two could be helpful to explain some of the flags. For example I'm unsure exactly what [fit_defense_classifier_outside_defense](https://github.com/ng390/armory/blob/poison-evaluation-repr/armory/scenarios/poisoning_gtsrb_scenario.py#L198) means

lcadalzo:
I think it'd be more clear if the metric variable names indicated whether they're poisoned vs. unpoisoned metrics.

For ex: [validation_metric](https://github.com/ng390/armory/blob/poison-evaluation-repr/armory/scenarios/poisoning_gtsrb_scenario.py#L247) is for benign samples, and [test_metric](https://github.com/ng390/armory/blob/poison-evaluation-repr/armory/scenarios/poisoning_gtsrb_scenario.py#L266) is for poisoned samples

ng390:
> Is there a reason we ignore the preprocessing fn returned by `load_model` [here](https://github.com/ng390/armory/blob/poison-evaluation-repr/armory/scenarios/poisoning_gtsrb_scenario.py#L193)? Is it assured that a classifer used for defense won't/shouldn't preprocess inputs?

We assume the classifier_for_defense and classifier use the same preprocessing function. The preprocessing function for the latter is fed into load_dataset. This could be made more general, but fits the evaluation needs right now. At a minimum, I will add a comment to the code documenting this.

ng390:
@lcadalzo PTAL - made the suggested changes.

PR #670, opened 2020-07-09
mongodb instructions updated

DRenardy:
The old MongoDB instructions work for the username / password authentication, provided those are passed with the mongodb url. The instructions are updated, and a regex is included in _save_to_mongo that will strip the username /password from the mongodb host name in the logs.

### DIFF
--- a/armory/scenarios/base.py
+++ b/armory/scenarios/base.py
@@ -24,6 +24,7 @@ from typing import Optional
 import coloredlogs
 import pymongo
 import pymongo.errors
+import re
 
 import armory
 from armory import paths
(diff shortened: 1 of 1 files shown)


armory/scenarios/base.py
```diff
@@ -122,8 +123,9 @@ def _send_to_mongo(self, mongo_host: str, output: dict):
         client = pymongo.MongoClient(mongo_host, MONGO_PORT)
         db = client[MONGO_DATABASE]
         col = db[MONGO_COLLECTION]
+        mongo_ip = re.findall(r"@([^@]*$)", mongo_host)[0]
```
adamj26, line 126: if `mongo_ip` does not have an `@` (eg an unauthed server), will `findall()` return a list with at least one element?

adamj26: if the regex doesn't find any `@`s, the `mongo_ip` can be printed as-is, since that would mean that there's no username or password included

PR #675, opened 2020-07-10
Disable shuffling of data during evaluation

### LINKED ISSUE #671: Make Order of Evaluation Examples Consistent
Currently the [call](https://github.com/twosixlabs/armory/blob/master/armory/data/datasets.py#L235) to `tfds.load()` is by default [set to True](https://github.com/twosixlabs/armory/blob/master/armory/data/datasets.py#L191).

It appears that a seed is only being set for the poisoning scenario, so for other scenarios we have the following issue: when we load in benign examples for evaluation, and then subsequently we load in examples to perturb before evaluating on, the ordering in these cases doesn't align.

lcadalzo:
fixes #671

### DIFF
--- a/armory/data/adversarial_datasets.py
+++ b/armory/data/adversarial_datasets.py
@@ -25,6 +25,7 @@ def imagenet_adversarial(
     clean_key: str = "clean",
     adversarial_key: str = "adversarial",
     targeted: bool = False,
+    shuffle_files: bool = False,
 ) -> datasets.ArmoryDataGenerator:
     """
     ILSVRC12 adversarial image dataset for ResNet50
@@ -49,7 +50,7 @@ def imagenet_adversarial(
         epochs=epochs,
         dataset_dir=dataset_dir,
         preprocessing_fn=preprocessing_fn,
-        shuffle_files=False,
+        shuffle_files=shuffle_files,
         cache_dataset=cache_dataset,
         framework=framework,
         lambda_map=lambda x, y: ((x[clean_key], x[adversarial_key]), y),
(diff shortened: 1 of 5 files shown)


dxoigmn:
FYI: When we have set `shuffle_files=False`, we have noticed memory leaks. This doesn't matter so much if you have swap on, but on my machine with 126GB of memory and swap off, the process gets killed by the kernel due to OOM. It may be worth running this and watching memory usage as the UCF101 scenario progresses.

davidslater:
@dxoigmn This is separate from the other UCF101 memory issues related to pytorch + attacks?

dxoigmn:
@davidslater Yes. This is a CPU memory leak.

davidslater:
We haven't been able to duplicate a memory leak on Ubuntu 18.04 on the UCF101 scenario. We'll keep this in mind though, if it pops up again.

dxoigmn:
> We haven't been able to duplicate a memory leak on Ubuntu 18.04 on the UCF101 scenario. We'll keep this in mind though, if it pops up again.

Interesting! When we upgrade Armory I will try to replicate on our side and report back.

davidslater:
Thanks, I appreciate it!

dxoigmn:
Quick Q: When the memory-leak test was done, was it run in Docker?

I don't see the leak happen in the Armory Docker containers. I do see it happen when running outside of Docker, so it may have something to do with the PyPI version of Tensorflow I install on the host versus the Conda version installed in the Docker container.

davidslater:
Yes, we were testing it in Docker. I think we also use the conda installed version when running on host, so if that is the issue, we wouldn't have found it.

PR #681, opened 2020-07-15
Docker root and command line docs

### LINKED ISSUE #679: Add documentation for command line arguments
Possibly in its own doc. Should include
--check
--num-eval-batches
--use-gpu
--gpus

davidslater:
Fixes #679 
Also enables `--root` command line arg to run the docker container as root instead of the current user, which is needed to run custom configs or containers that write to `/`.

### DIFF
--- a/docs/command_line.md
+++ b/docs/command_line.md
@@ -1,0 +1,73 @@
+# Command Line Usage
+
+## Root
+* `armory <command> --root [...]`
+Applies to `run`, `launch`, and `exec` commands.
+
+This will run the docker container as root instead of the host user.
+NOTE: this is incompatible with `--no-docker` mode.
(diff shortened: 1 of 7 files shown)


docs/command_line.md
```diff
@@ -0,0 +1,73 @@
...
+If a config is being run, the gpus used will be pulled from the config.
+If a config is not being run or that field is not in the config, it will default to all.
+
+NOTE: when running a config, these will overwrite the fields inside the config.
+
+### Example Usage
+
+Examples:
+```
+armory run scenario_configs/mnist_baseline.json --use_gpus
```
lcadalzo, line 47: change `--use_gpus` to `--use-gpu`

davidslater: Fixed in `docker.md` as well.

PR #683, opened 2020-07-16
Confirm whether attack is targeted after loading

### LINKED ISSUE #682: Confirm if ART attack is targeted after loading
Armory's scenario code determines whether an attack is targeted by [parsing](https://github.com/twosixlabs/armory/blob/master/armory/scenarios/image_classification.py#L94) the config and assumes False is there's nothing in the config to specify it's targeted. However, some ART [attacks](https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/1.2.0/art/attacks/evasion/boundary.py#L61) are targeted by default.

This causes an issue when `targeted: false` isn't explicitly set in the attack config for an ART attack that's targeted by default. In this case, ART will [error](https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/1.2.0/art/attacks/evasion/boundary.py#L152) if `use_label: false` or incorrectly use the _true_ labels to measure targeted attack success rate if `use_label: true`. This latter case caused me to see attack success rates of 100%, while accuracy on adversarial examples was also 100%

lcadalzo:
fixes #682

### DIFF
--- a/armory/scenarios/audio_classification.py
+++ b/armory/scenarios/audio_classification.py
@@ -113,6 +113,9 @@ class AudioClassificationTask(Scenario):
             )
         else:
             attack = load_attack(attack_config, classifier)
+            assert (
+                targeted == attack.targeted
+            ), "Error determining whether attack is targeted."
             test_data = load_dataset(
                 config["dataset"],
                 epochs=1,
--- a/armory/scenarios/audio_classification.py
+++ b/armory/scenarios/audio_classification.py
@@ -113,9 +113,10 @@ class AudioClassificationTask(Scenario):
             )
         else:
             attack = load_attack(attack_config, classifier)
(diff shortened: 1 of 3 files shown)


davidslater:
Is this primarily for sanity checking?

I am concerned that adding this will pin us to ART internal implementations, as it is possible to implement the ART EvasionAttack interface (https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/art/attacks/attack.py#L126) without setting that particular parameter. So, I would recommend something a little softer, like a warning (that also permits `attack.targeted` to not exist), like this:
```
if targeted != getattr(attack, "targeted", False):
    logger.warning(f"targeted config {targeted} != attack field {getattr(attack, 'targeted', False)}")
```

PR #686, opened 2020-07-17
enable use_label in audio and video scenarios 

### LINKED ISSUE #654: use_label = True fails for librispeech baseline and ucf101 scenarios
These result in dimension mismatch errors, as the input shape for x is changed but the shape for y is not changed to match.

The fix is relatively straightforward, and can be done in the scenarios; however, let's hold off until the submission period is over for a release.

davidslater:
Fixes #654

### DIFF
--- a/armory/scenarios/audio_classification.py
+++ b/armory/scenarios/audio_classification.py
@@ -5,6 +5,7 @@ General audio classification scenario
 import logging
 from typing import Optional
 
+import numpy as np
 from tqdm import tqdm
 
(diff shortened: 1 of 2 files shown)


davidslater:
To test, I recommend modifying `scenario_configs/librispeech_baseline_sincnet.json` and `scenario_configs/ucf101_baseline_pretrained.json`. Note that batch_size > 1 is not allowed in the audio example, though it works fine in the video example.

PR #692, opened 2020-07-20
Warn user about amory version mismatch

### LINKED ISSUE #661: Warn when launching an armory container with a mismatched version
It's a bit intuitive that upgrading `armory-testbed` requires you to update your config files to the matching docker image version. Instead we should pass in the launching armory version as an environment variable and check it with the installed version within the container. This method would also support custom docker images.

Ideally this would be a rather large warning so it's not overlooked (possibly failure but that may create issues for dev work).

davidslater:
Fixes #661

### DIFF
--- a/armory/environment.py
+++ b/armory/environment.py
@@ -1,0 +1,5 @@
+"""
+Environment parameter names
+"""
+
+ARMORY_VERSION = "ARMORY_VERSION"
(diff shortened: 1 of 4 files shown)


davidslater:
Also, fixes `logging -> logger` in a couple of places that I found while doing this.

adamj26:
LGTM!

PR #693, opened 2020-07-20
patch attacks

### LINKED ISSUE #687: Add functionality for patch attacks in image and video scenarios
ART API needs both `generate` and `apply_patch` methods being called.

Something like 
```
# scale = ???
patch, mask = attack.generate(x, y=y)
x_adv = attack.apply_patch(x, scale, patch)
```

davidslater:
Fixes #687

This should enable adversarialpatch attack. See the added config for an example.

Ideally, it should have `"use_label": false` and `"targeted": true`, but that requires the targeted PR #690 to get merged in first. However, I would like to get a review of this soon.

### DIFF
--- a/armory/art_experimental/attacks/patch.py
+++ b/armory/art_experimental/attacks/patch.py
@@ -1,0 +1,17 @@
+"""
+Support for patch attacks using same interface.
+"""
+
+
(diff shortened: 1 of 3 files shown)


davidslater:
Rebased on #694

PR #697, opened 2020-07-22
Update CI configs to ART 1.3

ng390:


ng390:
To test this, because of Issue #698 it is necessary to look manually at the CI log output for the "Armory run tests..." and Armory test without docker logs.

PR #699, opened 2020-07-22
Computational resource metrics

DRenardy:
Changes to metrics.py for collecting computational resource metrics using a context manager. Two different types of profiling can be selected: "Basic", which returns just basic CPU timing for the profiled portion, or "Deterministic", which does in-depth deterministic profiling with cProfile. This could be expanded to other profilers in the future. Each profiled portion is assigned a name, which is used later for calculating averages. If "Deterministic" profiling is selected, then the profiler results are dumped in the results file. This may not be the most elegant solution, but a user is warned if that is the case. 

I also added a variant of the image classification scenario that shows how to hook in the profiler. This maybe should be moved somewhere other than where I placed it.


armory/utils/metrics.py
```diff
@@ -270,6 +314,8 @@ def _generate_counters(self, names):
             )
         return [MetricList(x) for x in names]
 
+    from contextlib import contextmanager
+
```
davidslater, line 318: This looks unnecessary.

armory/scenarios/base.py
```diff
@@ -208,7 +208,7 @@ def run_config(config_json, from_file=False, check=False, mongo_host=None):
         "--mongo",
         dest="mongo_host",
         default=None,
-        help="Send scenario results to a MongoDB instance at the given host (eg 'localhost', '1.2.3.4', 'mongodb://USER:PASS@5.6.7.8')",
+        help="Send scenario results to a MongoDB instance at the given host (eg mongodb://USER:PASS@5.6.7.8')",
```
davidslater, line 211: Why are we removing the other types of instances?

Also, need to add a "'":
`        help="Send scenario results to a MongoDB instance at the given host (eg 'mongodb://USER:PASS@5.6.7.8')",`

armory/scenarios/base.py
```diff
@@ -119,12 +120,11 @@ def _send_to_mongo(self, mongo_host: str, output: dict):
...
         Send results to a Mongo database at mongo_host
         """
-        client = pymongo.MongoClient(mongo_host, MONGO_PORT)
+        client=pymongo.MongoClient(mongo_host,MONGO_PORT)
         db = client[MONGO_DATABASE]
         col = db[MONGO_COLLECTION]
-        logger.info(
-            f"Sending evaluation results to MongoDB instance {mongo_host}:{MONGO_PORT}"
-        )
+        mongo_ip=re.findall(r"@([^@]*$)", mongo_host)[0]
```
davidslater, line 126: Formatting looks off. Make sure that you have added the git commit pre hooks for black, flake8, etc.

armory/utils/metrics.py
DRenardy: It was indeed. Removed and tested.

armory/utils/metrics.py
```diff
@@ -151,6 +157,44 @@ def _snr_spectrogram(x_i, x_adv_i):
...
+        pr.enable()
+    startTime = time.perf_counter()
+    yield
+    elapsedTime = time.perf_counter() - startTime
+    if profiler == "Deterministic":
+        pr.disable()
+        s = io.StringIO()
+        sortby = "cumulative"
+        ps = pstats.Stats(pr, stream=s).sort_stats(sortby)
+        ps.print_stats()
```
davidslater, line 179: I would recommend removing the print line. It outputs a lot to the screen, which makes it difficult to see the other logs.

armory/scenarios/base.py
adamj26: Those were removed when we added authentication to the Mongo server, as the name must be passed with the schema in order for the username and password to be supplied

armory/utils/metrics.py
```diff
@@ -326,5 +370,20 @@ def results(self):
                         raise ZeroDivisionError(
                             f"No values to calculate mean in {prefix}_{metric.name}"
                         )
-
+        if computational_resource_dict:
```
adamj26, line 373: This `if` is unnecessary; the dict always exists and `for` looping over an empty dict is valid

DRenardy: Yup, removed.

docs/configuration_files.md
```diff
@@ -40,6 +40,7 @@ All configuration files are verified against the jsonschema definition at run ti
     perturbation: [String] Perturbation metric to calculate for adversarial examples
     record_metric_per_sample: [Bool] Boolean to record metric for every sample in save in output
     task: [List[String]] List of task metrics to record (e.g. categorical_accuracy)
+    profiler_type: [String] Type of computational resource profiling desired for scenario profiling
```
adamj26, line 43: Like `defense.type` and `dataset.framework`, the description should also list the acceptable values, eg

`…desired for scenario profiling. One of <Basic, Deterministic>`

adamj26: The type should also be `[Optional String]`, like eg `sysconfig.output_dir`

DRenardy:
The scenarios should be current to master after my last push. Should I still submit a new PR?

davidslater:
It is probably helpful to rebase and submit a new PR. It may be current, but due to when you branched from master, it has merge conflicts.

Issue #669, opened 2020-07-09
Patch Attack Size - Perturbation Metric


davidslater:
# pixel differences?

lcadalzo:
This isn't as straightforward to compute as expected because of preprocessing. Specifically, this [function](https://github.com/twosixlabs/armory/blob/master/armory/baseline_models/keras/densenet121_resisc45.py#L44) scales images to a max of 255. `max(x_adv)` typically equals 255, so adversarial images don't get scaled. `max(x)` is usually less than 255, so the benign examples do get scaled. At least, this is the case for the [resisc adversarial patch dataset.](https://github.com/twosixlabs/armory/blob/master/armory/data/adversarial/resisc45_densenet121_univpatch_and_univperturbation_adversarial_224x224.py)

Consequently, simply finding the indices where `x != x_adv` won't work. You can get a rough idea where the patch is located by finding where `|x - x_adv| > some_threshold` but this is imprecise.

PR #701, opened 2020-07-22
image circle patch area 

lcadalzo:
Addresses #669

### DIFF
--- a/armory/utils/metrics.py
+++ b/armory/utils/metrics.py
@@ -182,32 +182,48 @@ def _image_circle_patch_diameter(x_i, x_adv_i):
         raise ValueError(f"Expected image with 3 dimensions. x_i has shape {x_i.shape}")
     if (x_i == x_adv_i).mean() < 0.5:
         logging.warning(
-            f"x_i and x_adv_i differ at {int(100*(x_i != x_adv_i).mean())} percent of indices. image_circle_patch_area may not be accurate"
+            f"x_i and x_adv_i differ at {int(100*(x_i != x_adv_i).mean())} percent of "
+            "indices. image_circle_patch_area may not be accurate"
         )
-    channel_dim = img_shape.index(min(img_shape))
-    spat_ind = 1 if channel_dim != 1 else 0
+    # Identify which axes of input array are spatial vs. depth dimensions
+    depth_dim = img_shape.index(min(img_shape))
+    spat_ind = 1 if depth_dim != 1 else 0
+
+    # Determine which indices (along the spatial dimension) are perturbed
     pert_spatial_indices = set(np.where(x_i != x_adv_i)[spat_ind])
     if len(pert_spatial_indices) == 0:
         logging.warning("x_i == x_adv_i. image_circle_patch_area is 0")
         return 0
-    right_ind = max(pert_spatial_indices)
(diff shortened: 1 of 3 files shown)


armory/utils/metrics.py
```diff
@@ -174,6 +174,39 @@ def snr_spectrogram_db(x, x_adv):
     return [float(i) for i in 10 * np.log10(snr_spectrogram(x, x_adv))]
 
 
+def _image_circle_patch_area(x_i, x_adv_i):
+    if x_i.shape != x_adv_i.shape:
+        raise ValueError(f"x_i.shape {x_i.shape} != x_adv_i.shape {x_adv_i.shape}")
+    img_shape = x_i.shape
```
davidslater, line 180: Should return 0 if `x_i == x_adv_i`. Currently, it raises ValueError.

armory/utils/metrics.py
```diff
@@ -174,6 +174,39 @@ def snr_spectrogram_db(x, x_adv):
...
+    spat_ind = 1 if channel_dim != 1 else 0
+    pert_spatial_indices = set(np.where(x_i != x_adv_i)[spat_ind])
+    right_ind = max(pert_spatial_indices)
+    unpert_ind_left_of_patch = [
+        i for i in range(right_ind) if i not in pert_spatial_indices
+    ]
+    left_ind = max(unpert_ind_left_of_patch) + 1 if unpert_ind_left_of_patch else 0
+    if min(pert_spatial_indices) < left_ind:
+        logging.warning("Multiple regions of the image have been perturbed")
+    radius = (right_ind - left_ind) / 2.0
```
davidslater, line 197: Should be
```
radius = (right_ind - left_ind + 1) / 2.0
```

A patch size of 1 pixel (where right_ind would equal left_ind) should have non-zero area.

armory/utils/metrics.py
```diff
@@ -174,6 +174,39 @@ def snr_spectrogram_db(x, x_adv):
...
+    patch_area = np.pi * (radius * radius)
+    image_area = np.prod([dim for i, dim in enumerate(img_shape) if i != channel_dim])
+    return patch_area / image_area
+
+
+def image_circle_patch_area(x, x_adv):
+    """
+    Return what fraction of the image is covered by a circular patch
+    """
+    return [_image_circle_patch_area(x_i, x_adv_i) for (x_i, x_adv_i) in zip(x, x_adv)]
```
davidslater, line 207: Area and radius are a little bit confusing, actually, to think through, as a circular patch that covers the entire image has an area of 0.785. From the ART patch attacks, they either use a bounding box (DPatch) or a diameter (relative to overall image diameter).

I recommend returning `patch_diameter`, scaled to the largest circle that would fit in the image (the min of the two spatial dimensions). That will give us a value from 0 to 1 for all valid circles that fit in an image.

armory/utils/metrics.py
lcadalzo: addressed in ead1986

armory/utils/metrics.py
lcadalzo: addressed in db239ec

armory/utils/metrics.py
lcadalzo: addressed in db239ec

PR #703, opened 2020-07-23
Computational resource usage.

DRenardy:
Third times the charm. 

Fixed the scenarios to record computational resource usage. Profiler type is set in the config file.


armory/scenarios/video_ucf101_scenario.py
```diff
@@ -145,34 +148,41 @@ def _evaluate(self, config: dict, num_eval_batches: Optional[int]) -> dict:
...
+                    y_pred_adv = np.mean(
+                        classifier.predict(x_adv, batch_size=1), axis=0
+                    )
+                    if targeted:
+                        metrics_logger.update_task(
+                            y_target, y_pred_adv, adversarial=True
+                        )
+                    else:
+                        metrics_logger.update_task(y, y_pred_adv, adversarial=True)
+                    metrics_logger.update_perturbation([x], [x_adv])
```
davidslater, line 186: This should not be part of the attack calculation.

armory/scenarios/video_ucf101_scenario.py
```diff
@@ -145,34 +148,41 @@ def _evaluate(self, config: dict, num_eval_batches: Optional[int]) -> dict:
...
                 label_targeter = load_label_targeter(attack_config["targeted_labels"])
         for x_batch, y_batch in tqdm(test_data, desc="Attack"):
-            if attack_type == "preloaded":
-                x_batch = list(zip(*x_batch))
-                if targeted:
-                    y_batch = list(zip(*y_batch))
-            for x, y in zip(x_batch, y_batch):
+            with metrics.resource_context(
+                name="Attack", profiler=config["metric"].get("profiler_type")
+            ):
```
davidslater, line 153: I think you need to put this inside the inner loop.

armory/utils/metrics.py
```diff
@@ -151,6 +157,44 @@ def _snr_spectrogram(x_i, x_adv_i):
     return signal_power / noise_power
 
 
+@contextmanager
+def resource_context(name="Name", profiler=None):
+    profiler_types = ["Basic", "Deterministic"]
+    if profiler is not None and profiler not in profiler_types:
+        raise ValueError(f"Profiler {profiler} is not one of {profiler_types}.")
```
davidslater, line 164: There are a lot of "if" statements in the function, which make it harder to parse. I would recommend starting with something like:
```
if profiler is None:
    yield
    return 0
profiler_types = ["Basic", "Deterministic"]
...
```
as that will remove a lot of the ifs.

armory/utils/metrics.py
```diff
@@ -151,6 +157,44 @@ def _snr_spectrogram(x_i, x_adv_i):
...
+            computational_resource_dict[name]["total_time"] += elapsedTime
+            if profiler == "Deterministic":
+                computational_resource_dict[name]["stats"] += stats
+
+        else:
+            computational_resource_dict[name] = {}
+            computational_resource_dict[name]["execution_count"] = 1
+            computational_resource_dict[name]["total_time"] = elapsedTime
+            if profiler == "Deterministic":
+                computational_resource_dict[name]["stats"] = stats
```
davidslater, line 194: There's a lot of duplicate code here that could be collapsed. A defaultdict would work well for that, I think.

Something like:
```
from collections import defaultdict
...
if name not in computational_resource_dict:
    computational_resource_dict[name] = defaultdict(lambda: 0)  # uninitialized values default to 0
comp = computational_resource_dict[name]
comp["execution_count"] += 1
comp["total_time"] += elapsedTime
if profile == "Deterministic":
    comp["stats"] += stats
```

armory/utils/metrics.py
```diff
@@ -8,8 +8,14 @@
...
 import logging
 import numpy as np
+import time
+from contextlib import contextmanager
+import cProfile
+import pstats
+import io
 
 logger = logging.getLogger(__name__)
+computational_resource_dict = {}
```
davidslater, line 18: I'm not a fan of having these module-level mutable containers. Can you look into making the computational_resource_dict a part of the MetricsLogger class and have it be passed in to the resource context?

Alternatively, the resource_context method could be a method of the MetricsLogger class.

armory/utils/metrics.py
```diff
@@ -168,30 +172,23 @@ def resource_context(name="Name", profiler=None):
...
+    elapsedTime = time.perf_counter() - startTime
+    if profiler == "Deterministic":
+        pr.disable()
+        s = io.StringIO()
+        sortby = "cumulative"
+        ps = pstats.Stats(pr, stream=s).sort_stats(sortby)
+        ps.print_stats()
+        stats = s.getvalue()
+    if name not in computational_resource_dict:
+        computational_resource_dict[name] = defaultdict(lambda: 0)
```
ng390, line 186: It looks like if the profiler is "Deterministic", the value for the key "stats" is a string, so that the defaultdict needs to be initialized accordingly

armory/utils/metrics.py
```diff
@@ -327,4 +375,18 @@ def results(self):
...
+                    "Computational resource dictionary entry corrupted, missing data."
+                )
+            total_time = entry["total_time"]
+            execution_count = entry["execution_count"]
+            average_time = total_time / execution_count
+            results[
+                f"Avg. CPU time (s) for {execution_count} executions of {name}"
+            ] = average_time
+            if "stats" in entry:
+                results["{name} profiler stats"] = entry["stats"]
```
ng390, line 391:
```suggestion
                results[f"{name} profiler stats"] = entry["stats"]
```

ng390:
@DRenardy Two minor changes from my review, but otherwise good to me.

Issue #704, opened 2020-07-24
Inaccurate error message when requirements from `host-requirements.txt` are not installed


adamj26:
If one runs armory without ART (for example) installed in the environment, one gets the following error message:

```
$ python -m armory --help
ERROR: cannot import 'art' module
    Please run: $ pip install -r requirements.txt
```

But ART is in `host-requirements.txt`, not `requirements.txt`.

The error message should be changed to mention `host-requirements.txt`

davidslater:
This only occurs on the PR branch #706 due to some import changes there. Will fix on that branch and close this issue.

PR #706, opened 2020-07-24
Exit with a non-zero status code if the scenario in armory run errors

### LINKED ISSUE #698: Identifying errors when running scenarios in CI testing
Currently, when calling `armory run` to run scenarios from json configs in the CI testing, the return code is not captured. It is necessary to manually look at the logs to see if a scenario succeeds or fails in running. This can be corrected by finding a way to capture the return codes of the scenario run (passed through docker) or by parsing the log for "error" statements.

adamj26:
Modifies the base scenario to print a sentinel value before exiting cleanly. The ArmoryInstance that manages the container then checks for this value

As part of checking for the sentinel value, remove duplicate newlines from container log output

Closes #698

### DIFF
--- a/armory/docker/management.py
+++ b/armory/docker/management.py
@@ -6,7 +6,7 @@ import logging
 
 import docker
 
-from armory import paths
+from armory import paths, scenarios
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


adamj26:
The test failures appear to possibly be related to the dependency issues mentioned in #704

armory/docker/management.py
```diff
@@ -6,7 +6,7 @@
 
 import docker
 
-from armory import paths
+from armory import paths, scenarios
```
davidslater, line 9: This import is causing the host tests to fail (in conjunction with the updated init in scenarios).

Let's move the `END_SENTINEL` to somewhere like "armory/common.py"

PR #707, opened 2020-07-24
PyTorch data loading via Tensorflow-GPU to NumPy to Torch

### LINKED ISSUE #626: Make PyTorch Dataset Loaders for all datasets
Follow up to #455 which added experimental support for mnist, cifar10, and resisc45. 

Once we've thoroughly tested the experimental method it's implemented in 0.9 we need to functionality for variable amount of `num_workers`. Ensuring that multiple workers not return duplicate results.

## Also need support for the following datasets:
- [x] cifar10
- [x] mnist
- [ ] digit
- [ ] imagenette
- [ ] german_traffic_sign
- [ ] ucf101
- [x] resisc45
- [ ] librispeech_dev_clean
- [ ] imagenet_adversarial
- [ ] librispeech_adversarial
- [ ] resisc45_adversarial_224x224
- [ ] ucf101_adversarial_112x112

kevinmerchant:
Wraps a Tensorflow dataset with a simple PyTorch data loader.  I ran some simple benchmarks and this was found to be comparable in performance to a native PyTorch loader, provided that the tensorflow-gpu package is used.  Both frameworks internally store tensors in the same format as NumPy, so conversions are essentially free as long as the tensors are already stored on the CPU.

Closes #626


armory/data/pytorch_loader.py
```diff
@@ -0,0 +1,18 @@
...
+
+class TFToTorchGenerator(torch.utils.data.IterableDataset):
+    def __init__(self, tf_dataset):
+        super().__init__()
+        self.tf_dataset = tf_dataset
+
+    def __iter__(self):
+        for ex in self.tf_dataset.take(-1):
+            x, y = ex
+            if isinstance(x, tuple):
```
lcadalzo, line 12: is this logic to handle adversarial datasets? Maybe a comment would make this a bit more clear

lcadalzo: referring to `if isinstance(x, tuple):` in my previous comment

PR #709, opened 2020-07-29
Revert "Exit with a non-zero status code if the scenario in armory ru…

ng390:
…n errors (#706)"

This reverts commit 7fa809ac7e46bd1758554e5e19c6831fd4de4524.


ng390:
There are intermittent CI issues with this previous pull request. Specifically, it appears the loop: https://github.com/twosixlabs/armory/blob/7fa809ac7e46bd1758554e5e19c6831fd4de4524/armory/docker/management.py#L81 is somehow not capturing all the output of a scenario. See the failing test on the following [CI run](https://github.com/twosixlabs/armory/runs/923560131?check_suite_focus=true) - the sentinel is outputted in the log, but only after the check fails.

PR #713, opened 2020-08-12
Option to skip benign classification

### LINKED ISSUE #712: Create flag in scenarios to run benign inference
Create flag in scenarios to run benign inference

ng390:
Fixes #712. Also includes original command-line arguments in string passed to user in interactive mode.


lcadalzo:
Perhaps we could add to the logs indicating that benign inference is being skipped? Otherwise LGTM

PR #717, opened 2020-08-17
Video metrics

### LINKED ISSUE #715: Norm metric calculation incorrect for UCF101 video inputs
The `norm` function defined in https://github.com/twosixlabs/armory/blob/91f60420cdb7113e4cfddd7127c79e7280aac7e0/armory/utils/metrics.py#L74 assumes argument `x` has batch_size as the first dimension and the rest of the dimensions are associated with the model input (e.g., for image, `x` has shape `(nb, h, w, c)`, and each model input has shape `(h, w, c)`).

However, currently the way the video scenario is structured, whole videos are batched and passed into the norm function, but each video actually represents multiple model inputs. So `x` has shape `(nb, nf, 3, 16, 112, 112)`, and each model input has shape `(3, 16, 112, 112)`. When `norm` is called on `x`, it actually calculates one norm value over multiple inputs. This is only a problem since perturbation norms are based on a single input and not over multiple inputs. In order to correctly log the correct norm calculated by an attack algorithm, the `norm` function needs to treat the first two dimensions as batches.

davidslater:
Fixes #715

### DIFF
--- a/armory/utils/metrics.py
+++ b/armory/utils/metrics.py
@@ -287,6 +287,38 @@ SUPPORTED_METRICS = {
     "image_circle_patch_diameter": image_circle_patch_diameter,
 }
 
+# Image-based metrics applied to video
+
+
+def video_metric(metric, frame_average="mean"):
+    mapping = {
+        "mean": np.mean,
+        "max": np.max,
+        "min": np.min,
+    }
+    if frame_average not in mapping:
+        raise ValueError(f"frame_average {frame_average} not in {tuple(mapping)}")
+    frame_average_func = mapping[frame_average]
+
+    def func(x, x_adv):
+        results = []
+        for x_sample, x_adv_sample in zip(x, x_adv):
+            frames = metric(x_sample, x_adv_sample)
+            results.append(frame_average_func(frames))
+        return results
+
+    return func
+
+
+for metric_name in "l0", "l1", "l2", "linf", "image_circle_patch_diameter":
+    metric = SUPPORTED_METRICS[metric_name]
+    for prefix in "mean", "max":
+        new_metric_name = prefix + "_" + metric_name
+        if new_metric_name in SUPPORTED_METRICS:
+            raise ValueError(f"Duplicate metric {new_metric_name} in SUPPORTED_METRICS")
+        new_metric = video_metric(metric, frame_average=prefix)
+        SUPPORTED_METRICS[new_metric_name] = new_metric
+
 
 class MetricList:
     """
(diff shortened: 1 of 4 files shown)


davidslater:
Updated mars metric "mars_mean_l2". Should wait on Mike Tan's input before merging.

armory/utils/metrics.py
```diff
@@ -159,6 +159,50 @@ def _snr_spectrogram(x_i, x_adv_i):
...
+# Metrics specific to MARS model preprocessing in video UCF101 scenario
+
+
+def verify_mars(x, x_adv):
+    if len(x) != len(x_adv):
+        raise ValueError(f"len(x) {len(x)} != {len(x_adv)} len(x_adv)")
+    for x_i, x_adv_i in zip(x, x_adv):
+        if x_i.shape[1:] != x_adv_i.shape[1:]:
+            raise ValueError(f"Shape {x_i.shape[1:]} != {x_adv_i.shape[1:]}")
+        if x_i.shape[1:] != (16, 3, 112, 112):
```
ng390: Per Mike Tan's [comment](https://github.com/twosixlabs/armory/issues/715#issuecomment-675743273), it looks like the [Intel](https://github.com/twosixlabs/gard-submissions/blob/a94d66815c170b078827e6d06fea0b43512cd82b/repos/IntelLabs__GARD__gard-phase1-eval1/GARD/oscar/classifiers/ucf101_mars.py#L25) submission and [JHUM's custom classifier](https://github.com/twosixlabs/gard-submissions/blob/a94d66815c170b078827e6d06fea0b43512cd82b/repos/shaoyuanlo__GARD_eval1__master/GARD_eval1/ucf101_3bn.py#L364) don't directly use the MARS preprocessing. So a less strict check may make sense.

armory/utils/metrics.py
```diff
@@ -159,6 +159,50 @@ def _snr_spectrogram(x_i, x_adv_i):
...
+
+
+def verify_mars(x, x_adv):
+    if len(x) != len(x_adv):
+        raise ValueError(f"len(x) {len(x)} != {len(x_adv)} len(x_adv)")
+    for x_i, x_adv_i in zip(x, x_adv):
+        if x_i.shape[1:] != x_adv_i.shape[1:]:
+            raise ValueError(f"Shape {x_i.shape[1:]} != {x_adv_i.shape[1:]}")
+        if x_i.shape[1:] != (3, 16, 112, 112):
+            raise ValueError(f"Shape {x_i.shape[1:]} != (16, 3, 112, 112)")
```
ng390, line 172:
```suggestion
            raise ValueError(f"Shape {x_i.shape[1:]} != (3, 16, 112, 112)")
```

PR #723, opened 2020-08-18
check for the sentinel value indicating clean container shutdown more robustly

### LINKED ISSUE #720: Warnings issued during clean container shutdown confuse manager
When the container shuts down cleanly it prints a sentinel value that the manager looks for to recognize the clean exit. If, however, a warning is issued (eg by pytorch) during shutdown, the container will shutdown cleanly but the sentinel will not be the last value printed from the container, so the manager will incorrectly think that the container did not exit cleanly.

One possible solution is to have the manager look for the sentinel in any position in the log output from the container, not just the last position.

adamj26:
Instead of checking if the last line from the container was the sentinel, check if the container emitted the sentinel ever. This provides robustness to warning messages being printed during container shutdown, after the sentinel is emitted, which has been observed in some situations

Closes #720

### DIFF
--- a/armory/docker/management.py
+++ b/armory/docker/management.py
@@ -82,14 +82,14 @@ class ArmoryInstance(object):
         #  ie after the sentinel
         sentinel_found = False
         for out in log.output:
-            output = out.decode().rstrip().strip()
+            output = out.decode().strip()
             if not output:  # skip empty lines
                 continue
             # this looks absurd, but in some circumstances log.output will combine
             #  outputs from the container into a single string
             # eg, print(a); print(b) is delivered as 'a\r\nb'
             for inner_line in output.splitlines():
-                inner_output = inner_line.rstrip().strip()
+                inner_output = inner_line.strip()
                 if not inner_output:
                     continue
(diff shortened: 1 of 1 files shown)


armory/docker/management.py
```diff
@@ -77,14 +77,26 @@ def exec_cmd(self, cmd: str, user=""):
...
-
-        if last_output == scenarios.END_SENTINEL:
+            output = out.decode().rstrip().strip()
+            if not output:  # skip empty lines
+                continue
+            # this looks absurd, but in some circumstances log.output will combine
+            #  outputs from the container into a single string
+            # eg, print(a); print(b) is delivered as 'a\r\nb'
+            for inner_line in output.splitlines():
+                inner_output = inner_line.rstrip().strip()
```
davidslater, line 92: Is there a reason you're using `.rstrip().strip()` instead of `.strip()`

adamj26: I thought `strip()` only stripped trailing whitespace, but it strips both leading and trailing whitespace, so I've removed the calls to `rstrip()`

PR #725, opened 2020-08-21
set classifer inputs to be immutable

### LINKED ISSUE #719: Make Inputs Immutable
Some custom classifiers modify the values of inputs themselves in their `predict()` method, e.g. as is done [here](https://github.com/twosixlabs/gard-submissions/blob/july1-2020/repos/ryan-feng__armory_submission__gard-phase1-eval1/armory_submission/resisc45/xform_atta/arch/xform_classifier.py#L167-L168).

This is causing unexpected behavior in cases where the classifier's `predict()` is invoked; for example this call to ART's `compute_success()` modifies the value of `adv_x` itself which is then returned in the `generate()` function.

If the arrays are set to be immutable, this will raise an error during the `predict()` of any custom classifier that modifies inputs directly. This error can prompt the user to edit the classifier's `predict()`, e.g. make a copy of the array that can be modified, and avoid the unexpected behavior that's currently occurring.

lcadalzo:
Fixes #719

### DIFF
--- a/armory/scenarios/audio_classification.py
+++ b/armory/scenarios/audio_classification.py
@@ -100,6 +100,8 @@ class AudioClassificationTask(Scenario):
             logger.info("Running inference on benign examples...")
 
             for x, y in tqdm(test_data, desc="Benign"):
+                # Ensure that input sample isn't overwritten by classifier
+                x.flags.writeable = False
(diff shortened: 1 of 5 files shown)


ng390 (approved):
LGTM

PR #726, opened 2020-08-24
Enable --check runs of video scenarios with finetuning

### LINKED ISSUE #721: Video finetuning bug when --check is set
When finetuning for the video scenario, an AttributeError is raised when `--check` is passed since the EvalGenerator object (see [here](https://github.com/twosixlabs/armory/blob/master/armory/utils/config_loading.py#L62)) doesn't contain a `batches_per_epoch` attribute

The error can be reproduced by running:
`python -m armory run scenario_configs/ucf101_baseline_finetune.json --check`

lcadalzo:
Fixes #721.


ng390 (approved):
LGTM

PR #733, opened 2020-09-11
Add a `--no-use-gpu` flag

### LINKED ISSUE #731: Add a `--no-use-gpu` flag to force a scenario not to use a gpu
As an inverse of the `--use-gpu` flag, for use when running scenarios that specify `use_gpu` but don't actually require the GPU on a host without GPU support

ng390:
Fixes #731

### DIFF
--- a/armory/__main__.py
+++ b/armory/__main__.py
@@ -499,7 +499,6 @@ def launch(command_args, prog, description):
     _jupyter(parser)
     _port(parser)
     _use_gpu(parser)
-    _no_use_gpu(parser)
     _gpus(parser)
     _root(parser)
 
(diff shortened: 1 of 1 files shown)


davidslater (changes requested):
1) no-gpus should also function in the launch and exec commands
2) I think that it would be easier and more intuitive if you just modified the --use-gpu flag to be able to take a False value, such as `--use-gpu=false`. Alternatively, maybe just shorten the flag name to `--no-gpu`.

armory/__main__.py
```diff
@@ -170,18 +176,24 @@ def _root(parser):
...
         if not use_gpu:
             logger.info("--gpus field specified. Setting --use-gpu to True")
             use_gpu = True
         config["sysconfig"]["gpus"] = gpus
+
     if use_gpu or "use_gpu" not in config["sysconfig"]:
         # Override if use_gpu, otherwise if config exists, leave config setting in place
         config["sysconfig"]["use_gpu"] = use_gpu
+    elif no_gpu:
+        config["sysconfig"]["use_gpu"] = not no_gpu
```
davidslater, line 196: Let's just set this to `False`. It's a bit confusing as it is. The previous line utilizes `use_gpu` for the case that it's not specified in the sysconfig, which means it can be set to True or False, but here the only valid value is False.

davidslater (changes requested):
One nit, otherwise good.

davidslater (approved):
LGTM

PR #734, opened 2020-09-12
WIP: updated ucf scenario

davidslater:


armory/baseline_models/pytorch/ucf101_mars_update.py
```diff
@@ -0,0 +1,269 @@
...
+        video = torch.nn.functional.interpolate(
+            video,
+            size=(sample_size, sample_size),
+            mode="bilinear",
+            align_corners=align_corners,
+        )
+
+    if video.max() > 1.0:
+        raise ValueError("Video exceeded max after interpolation")
+    if video.min() > 1.0:
```
ng390, line 117: Should this be < -1.0?

davidslater: It should be < 0.0, will change.

armory/scenarios/video_update.py
```diff
@@ -0,0 +1,197 @@
...
+    batch = batch.astype(context.default_float) / context.quantization  # 255
+    assert batch.dtype == context.default_float
+    assert batch.max() <= 1.0
+    assert batch.min() >= 0.0
+
+    return batch
+
+
+class Ucf101(Scenario):
+    def _evaluate(
```
mzweilin, line 61: Can we decouple some loading logic as standalone callable methods? It would make us easier to tweak/visualize a scenario in Jupyter.

FYI, we have an internal implementation that has
```
_load_defenses()
_load_attacks()
_load_test_data()
_load_classifier_with_defense()
```

davidslater: Can you explain in more detail (or provide an example) of how these differ from the imported methods:
```
from armory.utils.config_loading import (
    load_dataset,
    load_model,
    load_attack,
    load_adversarial_dataset,
    load_label_targeter,
)
```

davidslater: That's a great idea! I'll include it as a separate issue and PR, though.

davidslater: Added https://github.com/twosixlabs/armory/issues/741

davidslater:
Rebasing and then will map to look more like image/audio updates.

PR #736, opened 2020-09-16
Allow configure to modify existing configuration

### LINKED ISSUE #664: Allow armory configure to reuse prior configuration
This is helpful when you only want to change a single line of the configuration.

kevinmerchant:
Closes #664


ng390 (changes requested):
One minor change with the user input, otherwise LGTM

ng390 (approved):

PR #738, opened 2020-09-17
added art_experimental modules for use as baseline defense

yusong-tan:


armory/art_experimental/defences/mp3_compression_channelized.py
```diff
@@ -0,0 +1,25 @@
...
+            channels_first=channels_first,
+            apply_fit=apply_fit,
+            apply_predict=apply_predict,
+        )
+
+    def __call__(self, x, y=None):
+        # add a channel axis
+        x = np.expand_dims(x, axis=-1)
+        x, _ = super().__call__(x)
+        x = np.squeeze(x)
```
ng390, line 24: @yusong-tan I believe if the batch size is 1, the squeeze operation will collapse the batch dimension which will cause errors during inference. Perhaps we could use an axis in the call to squeeze, or raise an error if the batch size is 1 for this defense?

ng390 (approved):
LGTM

PR #752, opened 2020-10-05
Cifar10 and MNIST preprocessing canonicalization

### LINKED ISSUE #749: Canonicalize CIFAR10 preprocessing
Need to also update Pytorch / Keras CIFAR models

ng390:
Fixes #749


armory/data/datasets.py
```diff
@@ -279,12 +279,44 @@ def _generator_from_tfds(
...
+    if batch.ndim != len(mnist_context.x_dimensions):
+        raise ValueError(
+            f"input batch dim {batch.ndim} != {len(mnist_context.x_dimensions)}"
+        )
+    for dim, (source, target) in enumerate(
+        zip(batch.shape, mnist_context.x_dimensions)
+    ):
+        pass
+    assert batch.dtype == np.uint8
+    assert batch.shape[2:] == mnist_context.x_dimensions[2:]
```
lcadalzo, line 302: should this be 
```suggestion
    assert batch.shape[1:] == mnist_context.x_dimensions[1:]
```

ng390: Good catch! Updated the same bug on CIFAR as well.

lcadalzo:
The `preprocessing_fn` arg that gets passed to the `cifar10` dataset loading fn [isn't actually used](https://github.com/ng390/armory/blob/cifar-mnist-canonicalization/armory/data/datasets.py#L346). So `cifar10` will always be loaded with canonical preprocessing.

For mnist, on the other hand, the `preprocessing_fn` arg is set to canonical by default but [allows the user to specify a different function.](https://github.com/ng390/armory/blob/cifar-mnist-canonicalization/armory/data/datasets.py#L334)

Which of these is the desired behavior? In my opinion the former, as (1) it enforces consistency and (2) preprocessing is now supposed to be performed inside the model. Regardless, we should be consistent across different datasets

ng390:
Agreed that always loading datasets with canonical preprocessing is preferred. I pushed the change for MNIST.

PR #754, opened 2020-10-07
APRICOT TF1 Faster RCNN baseline model

yusong-tan:


lcadalzo:
Closing because of #783

PR #756, opened 2020-10-09
Adding xview baseline model

yusong-tan:


lcadalzo:
this PR can be closed now because of #762

PR #762, opened 2020-10-13
Canonicalize Xview dataset to floats in [0,1]

### LINKED ISSUE #761: Canonicalize XView with floats in [0,1]
All other image datasets in Armory are canonicalized to lie in [0,1]. Additionally, ART classifiers need to receive floats or else the following error is raised:

`TypeError: The data type of input data `x` is uint8 and cannot represent negative values. Consider changing the data type of the input data `x` to a type that supports negative values e.g. np.float32.
`

lcadalzo:
Fixes #761


lcadalzo:
[This line ](https://github.com/pytorch/vision/blob/fef2c8a5d68cee482dd811d9ffbdf1418de15359/torchvision/models/detection/faster_rcnn.py#L361)tries to download the pretrained weights, writing them to $TORCH_HOME directory which by default cannot be written to without root access. There are two ways around this issue: either add the `--root` flag when running `armory run` or set the TORCH_HOME environment variable.

See [link](https://github.com/pytorch/vision/issues/616) for related discussion in torchvision repo.

ng390 (approved):
LGTM

PR #764, opened 2020-10-14
Create So2Sat dataset

ng390:
Add So2Sat dataset.


ng390:
Includes an update to _generator_from_tfds x_key loading to allow combining sentinel1 (first 4 channels) and sentinel2 (all 10 channels) into a single tensor.

kevinmerchant (approved):
LGTM

PR #765, opened 2020-10-14
Update pytorch to `1.6.0` and torchvision to `0.7.0`

adamj26:
Should `cudatoolkit` be updated as well? We're pinned to `10.1`, but `11.0.221` is available on Conda

Note that this PR may cause issues when we update ART to `1.4`, if they're pinned to a different version of `pytorch`


kevinmerchant (approved):
LGTM

PR #766, opened 2020-10-14
Librispeech dataset

ng390:
Librispeech full dataset integration


ng390:
To use same preprocessing method as librispeech_dev_clean

davidslater (approved):
LGTM.

PR #770, opened 2020-10-15
XView Object Detection Integration

### LINKED ISSUE #760: Add XView Object Detection Scenario

lcadalzo:
Fixes #760


davidslater (approved):
This works. I'm not a big fan of the precision and recall (mainly due to the hard-coded requirements on keys that differ between y and y_pred), or the differences between the xview dataset output and the estimator's output, but I think that can go in a follow-up PR. 

I will make an issue for it.

PR #771, opened 2020-10-15
Use maybe_download_weights_from_s3() when loading models

### LINKED ISSUE #646: Maybe download from s3 inside config_loading

adamj26:
Update all baseline models to expect to be given a path as produced by `maybe_download_weights_from_s3()` instead of a file name

Based on #657 

Fixes #646

### DIFF
--- a/armory/baseline_models/pytorch/ucf101_mars.py
+++ b/armory/baseline_models/pytorch/ucf101_mars.py
@@ -19,8 +19,6 @@ from MARS.opts import parse_opts
 from MARS.models.model import generate_model
 from MARS.dataset import preprocess_data
 
-from armory.data.utils import maybe_download_weights_from_s3
-
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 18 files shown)


armory/baseline_models/pytorch/ucf101_mars.py
```diff
@@ -54,13 +49,13 @@ def make_model(model_status="ucf101_trained", weights_file=None):
         opt.batch_size = 32
         opt.ft_begin_index = 4
 
-        opt.pretrain_path = filepath
+        opt.pretrain_path = weights_path
```
adamj26, line 52: This is a minor change in behavior — previously, if `weights_file` was `None`, this would be an error, as `filepath` would be undefined; now if `weights_path` is `None` it will also be `None`

armory/utils/config_loading.py
```diff
@@ -75,8 +76,18 @@ def load_model(model_config):
...
     weights_file = model_config.get("weights_file", None)
+    if isinstance(weights_file, str):
+        weights_path = maybe_download_weights_from_s3(weights_file)
+    elif isinstance(weights_file, list):
+        # TODO: if there are multiple weights files,
+        #  should model_fn() be called for each
+        #  or called once on the list?
+        weights_path = [maybe_download_weights_from_s3(w) for w in weights_file]
+    else:
+        weights_path = None
```
davidslater, line 87: If weights_file is a dict, mapping from key to filenames
```
{"model1": "path1.pth", ...}
```
list logic over dict.values()

davidslater: Also ensure it is one of None, str, list, dict.

armory/utils/config_loading.py
```diff
@@ -79,10 +79,9 @@ def load_model(model_config):
     if isinstance(weights_file, str):
         weights_path = maybe_download_weights_from_s3(weights_file)
     elif isinstance(weights_file, list):
-        # TODO: if there are multiple weights files,
-        #  should model_fn() be called for each
-        #  or called once on the list?
         weights_path = [maybe_download_weights_from_s3(w) for w in weights_file]
+    elif isinstance(weights_file, dict):
+        weights_path = [maybe_download_weights_from_s3(w) for w in weights_file.keys()]
```
davidslater, line 84: should be
```
weights_path = {k:maybe_download_weights_from_s3(v) for (k,v) in weights_file.items()}
```

davidslater (approved):
LGTM

PR #783, opened 2020-10-16
APRICOT object detection scenario

### LINKED ISSUE #772: Add APRICOT Object Detection Scenario
Use ART's ShapeShifter attack on the TF Faster-RCNN model

lcadalzo:
Fixes #772, #784 and #799.

The APRICOT dataset/scenario differs from the others in that there is no benign dataset, as the images contain physical adversarial patches. The current scenario configuration schema requires specifying certain fields that don't apply to this scenario: a (benign) dataset as well as perturbation metric (e.g. "linf").

Currently, this PR follows this schema and uses the already-existing [scenario code](https://github.com/lcadalzo/armory/blob/772-apricot-od-scenario/armory/scenarios/image_classification.py). It sets both the `dataset` and `attacks` config to load the same dataset. This means that, unless the user runs with `--skip-benign`, the scenario repeats evaluation of the same dataset. The perturbation metric is also set arbitrarily to "linf", as any metric will trivially return 0 since "`x`" [refers to](https://github.com/lcadalzo/armory/blob/772-apricot-od-scenario/armory/scenarios/image_classification.py#L159) `x_adv` (since, unlike other adversarial datasets in Armory, an iteration of APRICOT returns only one image).  [Conditional logic](https://github.com/lcadalzo/armory/blob/772-apricot-od-scenario/armory/scenarios/image_classification.py#L156-L159) is needed to parse this difference with the APRICOT dataset (unless we were to modify its preprocessing). Beyond this change, though, the scenario code remains the same.

Another option would be to add new scenario code and/or modify the configuration schema to handle the differences of the APRICOT dataset/scenario, which comes with the obvious downside of extra code/complexity that's only necessary for one scenario (and a scenario that's more limited in comparison to the XView object detection one).


lcadalzo:
Note: the model outputs 0-indexed class labels, so the predicted labels need to increased by one. This is done in ART only when the `standarsise_output` kwarg of the detector's `.predict()` is set to `True`. It is currently `False` by default, so there's a label offset mismatch between the APRICOT datasets and TF Faster-RCNN model. 

See [here](https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/dev_1.4.2/art/estimators/object_detection/tensorflow_faster_rcnn.py#L434).

kevinmerchant (approved):
LGTM

PR #792, opened 2020-10-16
Configurable magic number for Apricot dataset

ng390:
The apricot dataset uses 12 as the category_id for adversarial patches, which may be used for meaningful categories for other datasets. Add a lambda_map to make this category_id configurable.


lcadalzo (approved):
LGTM

PR #794, opened 2020-10-17
Unify OD dataset keys and fix OD metrics to filter out adv patch id from groundtruth labels

lcadalzo:
Fixes #785 and #786

PR #795, opened 2020-10-18
Add mAP metric for object detection

### LINKED ISSUE #759: Add mAP Metric for Object Detection
IOU threshold should default to 0.5 but can be configurable

lcadalzo:
Fixes #759

### DIFF
--- a/armory/utils/metrics.py
+++ b/armory/utils/metrics.py
@@ -574,12 +574,19 @@ def object_detection_AP_per_class(list_of_ys, list_of_y_preds):
 
             if highest_iou > IOU_THRESHOLD:
                 # If the gt box has not yet been covered
-                if img_idx_to_gtboxismatched_array[pred_box["img_idx"]][highest_iou_gt_idx] == 0:
+                if (
+                    img_idx_to_gtboxismatched_array[pred_box["img_idx"]][
+                        highest_iou_gt_idx
+                    ]
+                    == 0
+                ):
                     true_positives[pred_idx] = 1
 
                     # Record that we've now covered this gt box. Any subsequent
                     # pred boxes that overlap with it are considered false positives
-                    img_idx_to_gtboxismatched_array[pred_box["img_idx"]][highest_iou_gt_idx] = 1
+                    img_idx_to_gtboxismatched_array[pred_box["img_idx"]][
+                        highest_iou_gt_idx
+                    ] = 1
                 else:
                     # This gt box was already covered previously (i.e a different predicted
                     # box was deemed a true positive after overlapping with this gt box)
@@ -612,7 +619,9 @@ def object_detection_AP_per_class(list_of_ys, list_of_y_preds):
 
         # Compute mean precision across the different recall levels
         average_precision = interpolated_precisions.mean()
-        average_precisions_by_class[int(class_id)] = np.around(average_precision, decimals=2)
+        average_precisions_by_class[int(class_id)] = np.around(
+            average_precision, decimals=2
+        )
 
     return average_precisions_by_class
 
--- a/armory/utils/metrics.py
+++ b/armory/utils/metrics.py
@@ -433,7 +433,7 @@ def _check_object_detection_input(y, y_pred):
 
 def _intersection_over_union(box_1, box_2):
     """
-    Assumes format of [y1, x1, y2, x2]
+    Assumes format of [y1, x1, y2, x2] or [x1, y1, x2, y2]
     """
     assert box_1[2] >= box_1[0]
     assert box_2[2] >= box_2[0]
--- a/armory/utils/metrics.py
+++ b/armory/utils/metrics.py
@@ -603,7 +603,7 @@ def object_detection_mAP(list_of_ys, list_of_y_preds):
         # the corresponding recall exceeds the recall point
         # See http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.157.5766&rep=rep1&type=pdf
         for i, recall_point in enumerate(RECALL_POINTS):
-            precisions_points = precisions[np.where(recalls > recall_point)]
+            precisions_points = precisions[np.where(recalls >= recall_point)]
             # If there's no cutoff at which the recall > recall_point
             if len(precisions_points) == 0:
                 interpolated_precisions[i] = 0
--- a/armory/utils/metrics.py
+++ b/armory/utils/metrics.py
@@ -726,6 +726,12 @@ class MetricList:
         else:
             raise ValueError("total_wer() only for WER metric")
 
+    def mAP(self):
+        # Computed at once across all samples
+        y_s = [i[0] for i in self._inputs]
+        y_preds = [i[1] for i in self._inputs]
+        return object_detection_mAP(y_s, y_preds)
+
 
 class MetricsLogger:
     """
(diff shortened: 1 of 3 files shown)


lcadalzo:
For the [xView scenario](https://github.com/lcadalzo/armory/blob/759-mAP-metric/scenario_configs/xview_frcnn.json), mAP will return 0% because the ground-truth boxes are normalized while the predicted boxes are not. This needs to be fixed as part of a separate issue.

For the [APRICOT scenario](https://github.com/lcadalzo/armory/blob/772-apricot-od-scenario/scenario_configs/apricot_frcnn.json), maP will also return 0%. This is due to a label offset issue that also needs to be fixed as part of a separate issue (predicted classes are 0-indexed, while ground truth classes are 1-indexed). To test the mAP implementation, temporarily change [this line](https://github.com/lcadalzo/armory/blob/759-mAP-metric/armory/utils/metrics.py#L488) to `label = y_pred["labels"][pred_box_idx] + 1`, and run the [APRICOT scenario](https://github.com/lcadalzo/armory/blob/772-apricot-od-scenario/scenario_configs/apricot_frcnn.json)

A small change will need to made to the mAP function once #794 is merged in to account for the updated key swtich from `category_id` and `bbox` to `labels` and `boxes` respectively.

armory/utils/metrics.py
```diff
@@ -431,6 +431,192 @@ def _check_object_detection_input(y, y_pred):
...
+
+    # Union of (1) the set of all true classes and (2) the set of all predicted classes
+    set_of_class_ids = set([i["label"] for i in gt_boxes_list]) | set(
+        [i["label"] for i in pred_boxes_list]
+    )
+
+    # Remove the class ID that corresponds to a physical adversarial patch in APRICOT
+    # dataset, if present
+    # TODO: retrieve this ID from datasets.py rather than hardcoding it
+    set_of_class_ids.discard(-10)
```
kevinmerchant, line 514: Saw this was left as a TODO.  May want to replace this before merging to prevent confusion down the line.

armory/utils/metrics.py
```diff
@@ -431,6 +431,192 @@ def _check_object_detection_input(y, y_pred):
         )
 
 
+def _intersection_over_union(box_1, box_2):
+    """
+    Assumes format of [y1, x1, y2, x2]
```
kevinmerchant, line 436: I believe this will still work as intended if the input format is instead (x1, y1, x2, y2).  Is that correct?

armory/utils/metrics.py
```diff
@@ -431,6 +431,192 @@ def _check_object_detection_input(y, y_pred):
...
+
+        # Determine how many gt boxes (of class_id) there are in each image
+        num_gt_boxes_per_img = Counter([gt["img_idx"] for gt in class_gt_boxes])
+
+        # Initialize array where we'll keep track of whether a gt box has been matched to a
+        # prediction yet. This is necessary because if multiple predicted boxes of class_id
+        # overlap with a single gt box, only one of the predicted boxes can be considered a
+        # true positive
+        for img_idx, num_gt_boxes in num_gt_boxes_per_img.items():
+            num_gt_boxes_per_img[img_idx] = np.zeros(num_gt_boxes)
```
kevinmerchant, line 541: num_gt_boxes_per_img is type Counter, but what it stores is an array rather than a count.  It works, but I wonder if a regular dict is more appropriate.

armory/utils/metrics.py
```diff
@@ -431,6 +431,192 @@ def _check_object_detection_input(y, y_pred):
...
+
+        recalls = tp_cumulative_sum / (total_gt_boxes + 1e-6)
+        precisions = tp_cumulative_sum / (tp_cumulative_sum + fp_cumulative_sum + 1e-6)
+
+        interpolated_precisions = np.zeros(len(RECALL_POINTS))
+        # Interpolate the precision at each recall level by taking the max precision for which
+        # the corresponding recall exceeds the recall point
+        # See http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.157.5766&rep=rep1&type=pdf
+        for i, recall_point in enumerate(RECALL_POINTS):
+            precisions_points = precisions[np.where(recalls > recall_point)]
```
kevinmerchant, line 606: Per the link, I think this should be `recalls >= recall_point`

armory/utils/metrics.py
```diff
@@ -431,6 +431,192 @@ def _check_object_detection_input(y, y_pred):
...
+    box_1_area = (box_1[3] - box_1[1]) * (box_1[2] - box_1[0])
+    box_2_area = (box_2[3] - box_2[1]) * (box_2[2] - box_2[0])
+
+    iou = intersect_area / (box_1_area + box_2_area - intersect_area)
+    assert iou >= 0
+    assert iou <= 1
+    return iou
+
+
+def object_detection_mAP(list_of_ys, list_of_y_preds):
```
kevinmerchant, line 467: I wonder if mAP is a confusing name since it returns AP per class rather than mAP?

armory/utils/metrics.py
```diff
@@ -523,6 +711,9 @@ def values(self):
     def mean(self):
         return sum(float(x) for x in self._values) / len(self._values)
 
+    def append_inputs(self, *args):
```
kevinmerchant, line 714: Was this left in for debug purposes?

armory/utils/metrics.py
lcadalzo: yea, I wavered on this. I was considering just returning the actual mAP scalar value, but I thought it'd be useful to have the AP's by class logged to the user by default. 

I held off on naming it something like `AP_per_class` only because mAP is a ubiquitous term for object detection metrics. Any opinion @davidslater ?

davidslater: I would rather name it specifically what it is. In the scenario configs we can show how you can generate mAP from that.

armory/utils/metrics.py
lcadalzo: this method is used to store up all of the `y`'s and `y_pred`'s so that maP can be computed across all samples at once after inference has finished. `self._inputs` is used [here](https://github.com/lcadalzo/armory/blob/759-mAP-metric/armory/utils/metrics.py#L729) to compute mAP

armory/utils/metrics.py
lcadalzo: agreed, will update

armory/utils/metrics.py
lcadalzo: correct; updating documentation

armory/utils/metrics.py
lcadalzo: done

armory/utils/metrics.py
lcadalzo: It initially stores a count, but I agree that once the keys are mapped to np arrays the variable name no longer makes sense. Updating

armory/utils/metrics.py
lcadalzo: Renaming to `object_detection_AP_per_class()`. Currently, only the mAP scalar value is saved to the results dict that's written to the output json. I'm modifying this so that both the `average_precision_by_class` dict and the mAP scalar value are saved (so you don't have to manually compute mean from the dict values). They will both also be logged (as is currently implemented).

kevinmerchant (approved):
LGTM

PR #797, opened 2020-10-19
Load xView with PyTorch box format by default

### LINKED ISSUE #796: Load xView dataset with PyTorch bbox format by default
current (TF) format: [y1/height, x1/width, y2/height, x2/width]    
PyTorch format: [x1, y1, x2, y2] (unnormalized)

Users should have the ability to load with either format, but given that the xView baseline model is an ART PyTorch Faster-RCNN, load with PyTorch format as default. This will match the box format outputted by ART's PyTorch Faster-RCNN.

lcadalzo:
Fixes #796


kevinmerchant (approved):
LGTM

PR #812, opened 2020-10-20
Fix pytorch loader

### LINKED ISSUE #807: Fix PyTorch/TF Data Loading Issues
Currently [this line](https://github.com/twosixlabs/armory/blob/9a657c60af279d383ff42ce31339578d3dc1fc90/armory/data/pytorch_loader.py#L19) causes data loading to crash if framework is set to pytorch for object detection datasets.  `y` is assumed to be a NumPy array, but is actually a dictionary.

This should be solvable by iterating over all items in the dictionary and converting all TF Tensors to NumPy, then from NumPy to a PyTorch Tensor.

Additionally, data loading with framework of either TF or PyTorch causes `preprocessing_fn` and `label_preprocessing_fn` to be ignored [here](https://github.com/twosixlabs/armory/blob/9a657c60af279d383ff42ce31339578d3dc1fc90/armory/data/datasets.py#L271), but does not notify the user. As a result, the generator will return values that are not normalized, potentially causing scale differences between training and inference datasets.

kevinmerchant:
Closes #807 by 

1.  Throwing an error if the user selects pytorch or tf framework and applies a data or label preprocessing_fn
2.  For pytorch datasets, check if the labels are in dictionary format and, if so, convert each stored value that is a tensorflow tensor.

### DIFF
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -268,6 +268,9 @@ def _generator_from_tfds(
         ds = ds.batch(batch_size, drop_remainder=False)
     ds = ds.prefetch(tf.data.experimental.AUTOTUNE)
 
+    if framework != "numpy" and (preprocessing_fn is not None or label_preprocessing_fn is not None):
+        raise ValueError(f"Data/label preprocessing functions only supported for numpy framework.  Selected {framework} framework")
+
     if framework == "numpy":
         ds = tfds.as_numpy(ds, graph=default_graph)
         generator = ArmoryDataGenerator(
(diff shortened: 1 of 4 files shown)


lcadalzo:
Did you see a case where [this line](https://github.com/kevinmerchant/armory/blob/fix_pytorch_loader/armory/data/pytorch_loader.py#L29) is actually reached? I tested on xView and APRICOT and did not. If no, should we catch/handle the unexpected behavior rather than silently passing it along?

kevinmerchant:
I did not see a case where this was reached.  I added that in case non-tensor metadata was included in the labels in a future dataset.  I can raise an error here instead so we know to take another look if that ever happens.

lcadalzo:
Waiting on CI tests but LGTM.

lcadalzo (approved):

PR #814, opened 2020-10-20
Add multichannel Jpeg compression defense to art_experimental

ng390:
Add Jpeg compression defense to art_experimental for number of channels other than 1, 3.


ng390:
Likely want to test with updated scaling for so2sat dataset. With a single scaling factor for both the SAR and optical inputs, the benign accuracy is just 10.94%, however this may be expected. The range of the optical inputs is a totally different scale than the range of the SAR inputs. The quantization before the Jpeg compression thus loses a lot of information.

davidslater (approved):

PR #826, opened 2020-10-21
Update datasets, licensing, scenarios, baseline models, and metrics

kevinmerchant:


docs/metrics.md
```diff
@@ -17,13 +17,25 @@ is a JSON-able dict.
 | Name | Type | Description |
 |:-------: |:-------: |:-------: |
 | categorical_accuracy | Task | Categorical Accuracy |
-| top_5_categorical_accuracy | Task | Top-5 Categorical Accuracy |
 | object_detection_AP_per_class | Task | Average Precision @ IOU=0.5 |
+| object_detection_class_precision | Task | Per-class precision @ IOU=0.5 |
+| object_detection_class_recall | Task | Per-class recall @ IOU=0.5 |
```
lcadalzo, line 22: I think we discussed omitting `object_detection_class_recall` and `object_detection_class_precision` since they were placeholders until mAP was implemented. If we are going to include them, we might want to indicate that they're computed at a score_threshold of 0.5

kevinmerchant: Agreed - removing these from the documentation since AP is the metric we will typically care about.

docs/datasets.md
```diff
@@ -17,36 +17,45 @@ These tfrecord files will be pulled from S3 if not available on your
...
+| [so2sat](https://mediatum.ub.tum.de/1454690) | Co-registered synthetic aperture radar and multispectral optical images | (N, 32, 32, 14) | float32 | (N,) | int64 | train, validation |
 
 <br>
 
 ### Preprocessing
 
-Input-modifying preprocessing of datasets occurs as part of a model used within Armory. The cached
-datasets are preprocessed into tfrecords, however this preprocessing primarily consists of changing the
-representation of inputs, e.g. running pydub on flac audio files.
+Armory applies preprocessing to each convert each dataset to canonical form (e.g. normalize the range of values, set the data type).
```
lcadalzo, line 55: there's an extra "each" here

kevinmerchant: Fixed

davidslater (approved):
LGTM

PR #832, opened 2020-10-22
fast untarring when possible

### LINKED ISSUE #802: Speed up cache dataset extraction
Python implementation is really slow compared to command line usage. (About 10x slower.)

https://github.com/twosixlabs/armory/blob/master/armory/data/utils.py#L271-L272

If tar is not available on the command line, falling back to python is okay, but my preference would be try the linux version first.

davidslater:
Fixes #802

### DIFF
--- a/armory/data/utils.py
+++ b/armory/data/utils.py
@@ -10,6 +10,7 @@ import shutil
 import random
 import string
 import json
+import subprocess
 
 import boto3
 from botocore import UNSIGNED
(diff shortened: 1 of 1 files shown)


davidslater:
@mwartell I can't get the piping to work properly (it doesn't output files extracted, which would be nice, especially for very large files)

mwartell (approved):
LGTM.

absurdly trivial nitpick: verbose mode for tar is unneeded if we're just throwing the output away; mode "zxf" might save a few microseconds.

davidslater:
I would ideally like to print the output, which I thought the stdout and stderr piping would do. Can you figure out why that's not happening?

mwartell:
> I would ideally like to print the output, which I thought the stdout and stderr piping would do. Can you figure out why that's not happening?

It's because you didn't print it. The official docs often times get very cryptic, especially for subprocess. I made a quick stop at PYMOTW to confirm my recollection https://pymotw.com/3/subprocess/#capturing-output

You needed a `print(completedprocess.stdout.decode('utf-8'))` which I would add for you but I don't want to guess at your specific intention.

davidslater:
Ready for re-review. Didn't realize removing those pipes would give me what I was looking for.

mwartell (approved):
LGTM. Less _is_ more.

PR #842, opened 2020-10-24
Piping from stdin

### LINKED ISSUE #561: Overriding configuration options in command line
I wish something like this could be supported in Armory, so that we don't have to create many configuration files in the development phase.

```bash
$ armory run scenario_configs/baseline.json dataset.batch_size=32
```

davidslater:
Fixes #561

To test, try:
```
more scenario_configs/cifar10_baseline.json | armory run - --check
```

Should get a reasonable error message from:
```
echo "hi there" | armory run -
```

### DIFF
--- a/armory/__main__.py
+++ b/armory/__main__.py
@@ -262,9 +262,12 @@ def run(command_args, prog, description):
         )
         sys.exit(1)
     except json.decoder.JSONDecodeError:
-        logger.exception(f"Could not decode {args.filepath} as a json file.")
-        if not args.filepath.lower().endswith(".json"):
-            logger.warning(f"{args.filepath} is not a '*.json' file")
+        if args.filepath == "-":
+            logger.exception("'stdin' did not provide a json-parsable input")
+        else:
+            logger.exception(f"Could not decode '{args.filepath}' as a json file.")
+            if not args.filepath.lower().endswith(".json"):
+                logger.warning(f"{args.filepath} is not a '*.json' file")
         sys.exit(1)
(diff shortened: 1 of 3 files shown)


armory/__main__.py
```diff
@@ -248,16 +251,23 @@ def run(command_args, prog, description):
...
         logger.error(
             f"Could not validate config: {e.message} @ {'.'.join(e.absolute_path)}"
         )
         sys.exit(1)
     except json.decoder.JSONDecodeError:
-        logger.exception(f"Could not decode {args.filepath} as a json file.")
-        if not args.filepath.lower().endswith(".json"):
-            logger.warning(f"{args.filepath} is not a '*.json' file")
+        if args.filepath == "-":
+            logger.exception("'stdin' did not provide a json-parsable input")
```
adamj26, line 266: The traceback from this `logging.exception()` (and the other on line 268 below) obscures the error message and does not provide much useful feedback to the user.

Could this (and L268) be instead `logging.error()`

davidslater:
Added @mwartell in case @adamj26 is waylaid by other program.

davidslater:
One item that could use improvement:

If you do the following
```
armory run -
```
You get this:
```
2020-10-26 15:00:29 mac0831.local armory.__main__[24034] INFO Reading config from stdin...

```
and it hangs.

It would be nice if there was a way for it to catch that condition, as I am pretty sure that no one is actually going to type the config.

davidslater:
That update looks good, thanks!

adamj26 (approved):
LGTM

PR #845, opened 2020-10-26
Update attack params

davidslater:
This parameter has to be set to the max length of the test and dev dataset splits.


lcadalzo (approved):
LGTM

PR #859, opened 2020-10-27
working FGSM config for ASR + lint

### LINKED ISSUE #847: Provide sample config file for FGSM on ASR scenario
Currently, they run into type error on the input.

davidslater:
Fixes #847 

Config for ASR using FGSM.

The other file is some weird lint thing. It passes black formatting, so I don't know.


ng390 (approved):
LGTM

PR #865, opened 2020-10-28
scaled epsilon

### LINKED ISSUE #862: UCF101 scenario epsilons are too large
Given the c14n of datasets in Armory 0.12, the attack `eps` value in the UCF101 scenarios is too large now (51), which means the attacker is always successful. These values should be normalized to `[0, 1]` range. I imagine a similar analysis should be done for the other scenarios to make sure the `eps` values are in the correct range.

davidslater:
Fixes #862

Other scenarios appear appropriately scaled.


kevinmerchant (approved):
LGTM

Issue #805, opened 2020-10-20
Fix ART directory pathing in container


davidslater:
ART-related content is saved in the container in `/root/.art/` if the user is root and `/tmp/.art/` when the user is not root:
https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/bb7c1773902e4d2f3d43a2418136973eac55cf8d/art/config.py#L35-L38

However, these do not survive the container, which causes re-downloading of models every time the following scenario is run:
```
armory run scenario_configs/deepspeech_baseline.json --check
2020-10-19 16:48:25 mac0831.local armory.docker.management[59953] INFO ARMORY Instance db6949d356 created.
2020-10-19 16:48:25 mac0831.local armory.eval.evaluator[59953] INFO Running evaluation script
2020-10-19 23:48:28 db6949d3565b __main__[6] WARNING Outputs will be written to /armory/outputs/2020-10-19T234824.513615
2020-10-19 23:48:29 db6949d3565b armory.utils.external_repo[6] INFO Downloading external repo: hkakitani/deepspeech.pytorch
2020-10-19 23:48:29 db6949d3565b art.utils[6] INFO Downloading data from https://github.com/SeanNaren/deepspeech.pytorch/releases/download/v2.0/librispeech_pretrained_v2.pth
```

Ideally, these would survive the container and either be mapped to the host user's art path or to something in armory, similar to the TORCH_HOME issue.

davidslater:
Actually, this is a duplicate of #773 

I've left both up for now, as they have complementary info.

davidslater:
Blocked on ART issue here: Trusted-AI/adversarial-robustness-toolbox#693

This issue will make the data path easily configurable to "/armory/saved_models/art".

PR #870, opened 2020-10-28
Save deep speech model weights in armory directory

### LINKED ISSUE #856: Move DeepSpeech weights into armory saved_model_dir
https://github.com/SeanNaren/deepspeech.pytorch/releases/download/v2.0/librispeech_pretrained_v2.pth

davidslater:
Fixes #856 

There is a more permanent solution in #805 , but that requires a 1.5 update in ART.


kevinmerchant (approved):
LGTM

PR #873, opened 2020-10-30
Update host-requirements.txt

### LINKED ISSUE #872: Issue while running APRICOT Scenario in --no-docker mode
The baremetal installation (host-requirements.txt) seems outdated since it installs an older version of ART. APRICOT scenario is compatible only with ART version 1.4.2 - which is specified in the Dockerfile (this version is not yet available with pypi, and can be installed only from the git repo)

Please modify the host-requirements.txt with recent versions.

ng390:
Fixes #872


davidslater (approved):
LGTM

PR #875, opened 2020-10-30
Sanity check model configurations

### LINKED ISSUE #753: Sanity Check Models
We should add a few basic tests to ensure that models are configured correctly before running evaluation.  Checks could include verifying the model returns a gradient, assuring that the clip values are the right type and shape, etc.

Note that verifying model input shapes will likely be performed in the new canonical data preprocessing.

kevinmerchant:
Closes #753 

To run/test, run the following command using one of the example configurations in the scenario_configs folder:
`python -m armory run <path_to_scenario_config_json> --validate-config`


davidslater:
`armory/utils` seems like an odd place to put the `test_config` folder. How about creating a higher level folder like `armory/validation` or `armory/validate` or `armory/verify` or `armory/checks` and putting it there?

kevinmerchant:
Moved to `armory/validation`

lcadalzo (changes requested):
Generate arbitrary ground truth for object detectors, rather than using the predicted output as ground truth

armory/validation/test_config/test_model.py
```diff
@@ -0,0 +1,135 @@
...
+        np.broadcast_to(clip_max, input_shape)
+
+        # verify model forward pass
+        test_input = self._get_test_input(input_shape, clip_min, clip_max)
+        copy_input = test_input.copy()
+        test_output = model.predict(test_input)
+        assert np.all(
+            test_input == copy_input
+        ), "Model prediction overwrites the input x value"
+        if hasattr(model, "nb_classes"):
```
lcadalzo, line 97: This is minor, but `test_output.shape` will fail when `test_output` is not a numpy array, which is the case for object detection. I suppose the ART object detectors don't have a `nb_classes` attribute, so this conditional block likely won't be reached, but there would be nothing stopping a user from adding a `nb_classes` attribute in their custom detector

kevinmerchant:
Modified the test ground truth generation to create a set of bounding boxes for object detector gradient testing.

lcadalzo (approved):

lcadalzo:
LGTM but I think it'd be useful to include the use/purpose of `--validate-config` in the documentation

PR #877, opened 2020-11-02
Add --skip-attack flag

### LINKED ISSUE #849: Add `--skip-attack` argument to just get benign output

lcadalzo:
Fixes #849.


armory/__main__.py
```diff
@@ -268,10 +268,19 @@ def run(command_args, prog, description):
...
+        help="Skip attack generation and metric calculations",
+    )
 
     args = parser.parse_args(command_args)
     coloredlogs.install(level=args.log_level)
 
+    if args.skip_benign and args.skip_attack:
+        logger.error("--skip-benign and --skip-attack cannot both be passed. ")
+        sys.exit(1)
+
```
davidslater, line 283: I would probably allow this to happen, particularly for the case if someone just wanted to train a model and ensure that piece worked, but it could be helpful as a test no-op.

davidslater (changes requested):
One minor thing. Doing some test runs currently.

armory/scenarios/base.py
```diff
@@ -311,12 +322,16 @@ def init_interactive(config_json, from_file=True):
             "--num_eval_batches will be overwritten and set to 1 since --check was passed"
         )
 
+    if args.skip_benign and args.skip_attack:
+        raise ValueError("--skip-benign and --skip-attack cannot both be passed")
+
```
davidslater, line 327: Ditto on this. When I comment this and the above similar lines in `__main__`, it runs smoothly and has `{}` for results, which is expected.

davidslater:
My tests looked good. I would just have you remove those lines I mentioned and we're good to go.

davidslater (approved):
LGTM.

PR #878, opened 2020-11-02
Dataset index and slicing

### LINKED ISSUE #871: Add "skip" feature to dataset kwargs and dataset generator to start at a specific index
This can be very useful for testing. For instance, sample 3313 in the UCF101 test set is the largest sample, so it would be helpful to test directly on that without processing through the rest.

davidslater:
Fixes #871 

This PR does a few related things:
1) Renames "split_type" to "split" to make it clear that it is verbatim what is given to TFDS load (same kwarg name)
https://www.tensorflow.org/datasets/api_docs/python/tfds/load

2) Modifies scenarios so that configs can override the train and eval splits. This enables TFDS slicing operations on them such as `"test[:10]"` or `"train_clean100+train_clean360+train_other500"`:
https://www.tensorflow.org/datasets/splits#examples

3) Adds additional parsing of splits to enable individual indexing for a specific data point, such as "test[355]", and indexing based off of a list, "test[[1, 4, 5, 7, 9]]". Under the hood, these are translated to single-element slices. Advanced slicing operations involving a `step` parameter, such as "test[10:20:2]" are not permitted, and raise a NotImplementedError.

### DIFF
--- a/armory/data/adversarial_datasets.py
+++ b/armory/data/adversarial_datasets.py
@@ -56,7 +56,7 @@ def apricot_canonical_preprocessing(batch):
 
 
 def imagenet_adversarial(
-    split_type: str = "adversarial",
+    split: str = "adversarial",
     epochs: int = 1,
     batch_size: int = 1,
     dataset_dir: str = None,
@@ -86,7 +86,7 @@ def imagenet_adversarial(
 
     return datasets._generator_from_tfds(
         "imagenet_adversarial:1.1.0",
-        split_type=split_type,
+        split=split,
         batch_size=batch_size,
         epochs=epochs,
         dataset_dir=dataset_dir,
@@ -99,7 +99,7 @@ def imagenet_adversarial(
 
 
 def librispeech_adversarial(
-    split_type: str = "adversarial",
+    split: str = "adversarial",
     epochs: int = 1,
     batch_size: int = 1,
     dataset_dir: str = None,
@@ -115,7 +115,7 @@ def librispeech_adversarial(
     Adversarial dataset based on Librispeech-dev-clean including clean,
     Universal Perturbation using PGD, and PGD.
 
-    split_type - one of ("adversarial")
+    split - one of ("adversarial")
 
     returns:
         Generator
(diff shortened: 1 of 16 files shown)


davidslater:
Biggest code changes are in `datasets.py` and its associated test file.

lcadalzo:
What's the intended behavior when overlapping splits are provided? E.g. `"train[1:3] + train[2]"`. The current behavior is to repeat elements, so in this case `train[2]` is returned in two consecutive batches.

lcadalzo:
Is it of interest to allow for specifying negative indices when (1) specifying a single index and/or (2) when providing a list of indices?

(1) E.g, if I only want to evaluate on the penultimate sample, specifying `"train[-2]"` will cause an error. `"train[[-2:-1]]"`, however, will yield the one sample I want.

(2) E.g., if I want to evaluate on the first and penultimate sample, `"train[[0, -2]]"` will cause an error since only positive indices are supported when passing a list. And `"train[0] + train[[-2:-1]]"` would work.

davidslater:
> What's the intended behavior when overlapping splits are provided? E.g. `"train[1:3] + train[2]"`. The current behavior is to repeat elements, so in this case `train[2]` is returned in two consecutive batches.

The intended behavior is simply to match the underlying TFDS behavior, which allows duplicate and out-of-order splits. I don't expect people to make much use of it, but it is much simpler on our end than trying to determine uniqueness.

davidslater:
> Is it of interest to allow for specifying negative indices when (1) specifying a single index and/or (2) when providing a list of indices?
> 
> (1) E.g, if I only want to evaluate on the penultimate sample, specifying `"train[-2]"` will cause an error. `"train[-2:-1]"`, however, will yield the one sample I want.
> 
> (2) E.g., if I want to evaluate on the first and penultimate sample, `"train[[0, -2]]"` will cause an error since only positive indices are supported when passing a list. And `"train[0] + train[-2:-1]"` would work.

I had not considered using negative indices. I didn't realize they would work. I can make an update to allow them if you think that makes sense.

lcadalzo:
I think, since it's unlikely they would be commonly used, we can hold off and only add later if it's something that a user is specifically looking for.

lcadalzo (approved):
LGTM

PR #880, opened 2020-11-03
Optionally truncate very long videos in MARS model

### LINKED ISSUE #869: UCF101 baseline scenario OOM error during attack phase
I am seeing out of memory errors during the `ucf101_baseline_pretrained.json` scenario on our Titan Xp 12GB GPUs. I see this happening after the 2760th example (of 3783 total) during the attack phase of the scenario. My hunch is the change to processing whole videos is the cause since now ART needs to compute gradients with respect to the whole video at once. This may be due to memory not be freed appropriately (i.e., there needs to be a `del` placed somewhere) or the 2760th example (+- 1 example, perhaps) may be just too large to fit on the GPU.

davidslater:
Fixes #869


ng390 (approved):
LGTM

PR #884, opened 2020-11-04
Split metrics when targeted attack is present

### LINKED ISSUE #848: Split adversarial and task metrics
This is particularly important for targeted attacks, such as WER in ASR scenarios.

kevinmerchant:
Closes #848 

For targeted attacks, this will now compute the selected metrics two different ways:

1. Relative to the ground truth labels (i.e. a measure of the effectiveness of the defense).
2. Relative to the target labels (i.e. a measure of the effectiveness of the attack).

For untargeted attacks, only the first metric is computed.  We could compute the second metric in some cases, such as `1-accuracy` for classification problems, but the desired behavior is less obvious for other problem types (e.g. object detection).

### DIFF
--- a/armory/scenarios/audio_asr.py
+++ b/armory/scenarios/audio_asr.py
@@ -85,9 +85,14 @@ class AutomaticSpeechRecognition(Scenario):
                 "this is not yet supported for speech recognition models."
             )
 
+        attack_config = config["attack"]
+        attack_type = attack_config.get("type")
+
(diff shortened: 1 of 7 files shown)


davidslater (changes requested):
These seem to work well for me. Can you provide some documentation on how this works?

kevinmerchant:
Added a short paragraph to `docs/metrics.md` to explain the new behavior.

davidslater (approved):

Issue #852, opened 2020-10-27
Move ART to the end of the docker container installs


davidslater:

lcadalzo:
This was addressed in #887.

PR #887, opened 2020-11-04
Pin to ART 1.4.2 and move to end of container installs

### LINKED ISSUE #886: Pin to ART 1.4.2
Currently pinned to the dev branch

lcadalzo:
Fixes #886 and #852


docker/Dockerfile
```diff
@@ -34,4 +33,6 @@ RUN /opt/conda/bin/pip install --no-cache-dir \
 RUN /opt/conda/bin/conda install -c conda-forge ffmpeg==4.2.3 && \
     /opt/conda/bin/conda clean --all
 
+RUN /opt/conda/bin/pip install --no-cache-dir adversarial-robustness-toolbox==1.4.2
+
```
davidslater, line 37: How does this work if it is moved to the pytorch/tf1/tf2 dockerfiles after the installation of those frameworks?

davidslater (changes requested):
I think you'll need to add something like this to {framework}-dev Dockerfiles. For those, let's still just do a normal pip install of art.

Have you tried building all of these?

lcadalzo:
I'd forgotten to include the dev Dockerfiles but address that in the latest commit. After making this change, I tried pytorch-dev which built fine. Didn't test tf1 and tf2 locally

davidslater (approved):

davidslater:
The CI testing will check builds, so I think we're good.

PR #888, opened 2020-11-04
GTSRB canonical preprocessing

### LINKED ISSUE #830: German Traffic Sign Preprocessing
Add canonical preprocessing to this dataset and the adversarial counterpart. Since we primarily use this for poisoning, we may consider having the canonical version also map it to a single shape so that it's easy to use in the scenario.

ng390:
Fixes #830


armory/data/datasets.py
```diff
@@ -730,11 +731,34 @@ def imagenette(
     )
 
 
+def gtsrb_canonical_preprocessing(img):
+    img_size = 48
+    img_out = []
+    quantization = 255.0
+    for im in img:
+        img_eq = ImageOps.equalize(Image.fromarray(im))
```
davidslater, line 739: I'm not sure about doing equalization inside of the canonical preprocessing step - that can significantly change an image and is not a simple thing to reverse.

davidslater (changes requested):
Instead of making that preprocessing function the "canonical" preprocessing, let's make the canonical preprocessing function the standard image one without resizing, eq, or cropping.

Instead, let's keep the scenario-specific preprocessing_fn in the scenario itself, and then override the preprocessing_fn in the data calls in the scenario. Then document this.

davidslater (approved):
LGTM

PR #897, opened 2020-11-05
Remove adversarial configs

### LINKED ISSUE #894: Check adversarial configs
Remove if they fail.

davidslater:
Fixes #894 

The librispeech one works, technically, but the scale is completely wrong. Also renamed adversarialpatch to patch to make it clear that it wasn't an adversarial dataset.


ng390 (approved):
LGTM

PR #903, opened 2020-11-05
exclude microseconds from output directory timestamp name

### LINKED ISSUE #902: evaluator output directory does not need microsecond accuracy
The name filename `outputs/2020-11-05T000907.224346/ImageClassificationTask_1604535968.json` requires explanation in the tutorial.

"If the implementation is hard to explain, it's a bad idea." — PEP-20

mwartell:
Fixes #902 

There are more trivial pull requests in the world, but not many.


adamj26:
LGTM, assuming the failing pytorch test is unrelated, which I believe it is

davidslater (commented):
My main concern with this is that we're using the timestamp to ensure different runs don't write to the same tmp and output directories. With this change, if you were to launch several runs simultaneously, would this cause them both to write to the same place?

adamj26:
If multiple runs were launched within the same second, they would potentially end up colliding, yes

davidslater:
I would like to avoid those collisions, especially for scripts that launch several runs in parallel.

What's the benefit to doing this? Less messy names for file folders?

One possibility is using folder creation as a locking mechanism, so that if the current second is taken, you increment by 1 and try to take the next one, etc.

mwartell:
It was an aesthetic issue. I am withdrawing it because I didn't appreciate the circumstance in which it would run.

PR #911, opened 2020-11-06
Return non-zero exit code if validate-config fails

### LINKED ISSUE #910: Return non-zero exit code when --validate-config fails
We should return a non-zero exit code when `--validate-config` fails a test so that this can be automatically detected externally.

kevinmerchant:
Closes #910


adamj26 (approved):
LGTM

PR #912, opened 2020-11-09
Log output dir at end of run

### LINKED ISSUE #908: Log the full output path upon closing scenario
Here's an example of what gets logged following a scenario:

`2020-11-05 19:47:55 0be046fb6e0a armory.scenarios.base[266] INFO Saving evaluation results saved to <output_dir>/ObjectDetectionTask_1604605675.json`

The user has to scroll up to the beginning of the logs to see what `<output_dir>` actually is

lcadalzo:
Fixes #908


ng390:
This logs the DockerPaths version of the output directory. Would it be clearer to log the HostPaths version, unless in interactive mode? Or we could just emphasize that it is a path in the docker container.

ng390 (changes requested):
Minor changes w.r.t. using DockerPaths/HostPaths

lcadalzo:
Currently we log DockerPaths regardless, see this example log that occurs at start of run:
`2020-11-12 17:20:51 6a9c55b655dd __main__[4290] WARNING Outputs will be written to /armory/outputs/2020-11-12T150059.839212`

This class doesn't know whether `--interactive` was set or not. Perhaps we could modify to: 

"Saving evaluation results to {scenario_output_dir}/{filename} path inside container"

ng390 (approved):

PR #915, opened 2020-11-09
Update So2Sat scenario

yusong-tan:
- Modified art_experimental pgd_patch attack to mask arbitrary channels from config files
- Corrected art_experimental jpeg_compression_multichannel_image so inverting input normalization is done in the correct order
- Updated so2sat baseline model to convert channelized clip_values list to ndarray
- Update so2sat configs to reflect baseline evaluation threat model


armory/baseline_models/keras/so2sat.py
```diff
@@ -66,5 +68,10 @@ def get_art_model(
     if weights_path:
         model.load_weights(weights_path)
 
-    wrapped_model = KerasClassifier(model=model, **wrapper_kwargs)
+    if "clip_values" in wrapper_kwargs:
+        if isinstance(wrapper_kwargs["clip_values"], list):
+            wrapper_kwargs2 = deepcopy(wrapper_kwargs)
```
kevinmerchant: This should move outside the `if` statement, otherwise this will crash if `clip_values` is not set.  I can make this change.

Otherwise, LGTM

kevinmerchant (approved):
LGTM

PR #916, opened 2020-11-10
Kenansville DFT attack

yusong-tan:
- Implements a blackbox signal processing attack called Kenansville
- Fixes art_experimental MP3 compression
- Added baseline evaluations configs for this attack


armory/art_experimental/defences/mp3_compression_channelized.py
```diff
@@ -25,7 +25,9 @@ def __call__(self, x, y=None):
             )
 
         # add a channel axis
-        x = np.expand_dims(x, axis=-1)
+        for i in range(len(x)):
+            x[i] = np.expand_dims(x[i], axis=-1)
```
kevinmerchant, line 29: Can we assign this to another variable so that the input array is not overwritten?  Also, can the `axis` parameter be set based on whether `channels_first` is True or False? Otherwise, we should assert that `channels_first` is False

armory/art_experimental/attacks/kenansville_dft.py
```diff
@@ -0,0 +1,93 @@
...
+        return np.real(x_ifft).astype(np.float32)
+
+    def generate(self, x):
+        x_out = np.empty((len(x),), dtype=object)
+        for i, x_example in enumerate(x):
+            if self.partial_attack:
+                # split input into multiple segments and attack each with some probability
+                x_adv = np.zeros_like(x_example)
+                seg_len = self.attack_len
+                for j in range(len(x_example) // seg_len + 1):
```
kevinmerchant, line 84: We should use a `ceil` method here.  Otherwise, this fails in the (rare) case where `len(x_example)` is an integer multiple of `seg_len`.

armory/art_experimental/attacks/kenansville_dft.py
```diff
@@ -0,0 +1,93 @@
...
+class KenansvilleDFT:
+    def __init__(
+        self,
+        estimator,
+        sample_rate=16000,
+        snr_db=100,
+        partial_attack=False,
+        attack_len=500,
+        attack_prob=0.5,
+        targeted=False,
```
kevinmerchant, line 13: So that it is more clear to the user that `targeted` is not used, we should add an assert statement to verify it is False.

armory/art_experimental/attacks/kenansville_dft.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+        The goal of the following search is to find all the
+        low power frequencies that can be discarded while
+        maintaining a minimum SNR.
+
+        If desired, the following coarse and fine search could
+        be replaced with binary search for faster convergence.
+        """
+        # coarse search
+        id = 2  # real signals have symmetry in positive and negative frequencies
```
kevinmerchant, line 54: If the length of `x` is odd, this method of grouping bins will fail.  Though every positive frequency has a corresponding negative frequency that may be grouped, the 0 index does not have a matched index for odd-length sequences.

armory/art_experimental/attacks/kenansville_dft.py
```diff
@@ -0,0 +1,93 @@
...
+        noise = np.sum(x_psd[x_psd_ind[:id]])
+        noise_db = 10 * np.log10(noise)
+        while signal_db - noise_db > self.snr_db:
```
kevinmerchant, line 57: If this loop is never entered (such as if the user requests too high of a value for `self.snr_db`), `id` will be divided by 2 and become 1 immediately after this, breaking the symmetry.

armory/art_experimental/attacks/kenansville_dft.py
```diff
@@ -0,0 +1,93 @@
...
+            id *= 2
+            noise = np.sum(x_psd[x_psd_ind[:id]])
+            noise_db = 10 * np.log10(noise)
+
+        # fine search
+        id = int(id / 2)
+        noise = np.sum(x_psd[x_psd_ind[:id]])
+        noise_db = 10 * np.log10(noise)
+        while signal_db - noise_db > self.snr_db:
```
kevinmerchant, line 66: Similarly, if this loop is never entered,  we should not subtract 2 at the next step.  As a simple example, assume the case where neither the coarse or fine searches are entered because `self.snr_db` is too high.  Then, `id` will be 1 before this loop, and `id-2` will become negative, causing most values to be zeroed out.

kevinmerchant (changes requested):
I added a few comments to address before this is ready to merge in.

Also, when running the provided configurations with `num-eval-batches=10`, I noticed that the WER on adversarial examples was actually higher for the defended version than the baseline model.  Do we know why this happens? Does this change if I evaluate for more batches?

armory/art_experimental/defences/mp3_compression_channelized.py
yusong-tan: @kevinmerchant do you want to go ahead and make the changes as needed to be backward compatible and future proof?

As for WER increasing in defended model, this made sense because the DFT attack is not additive (only subtractive), and compression shouldn't improve performance.

armory/art_experimental/attacks/kenansville_dft.py
yusong-tan: Sounds good. Again, do you want to change it?

armory/art_experimental/attacks/kenansville_dft.py
yusong-tan: Good catch. I'll fix it.

armory/art_experimental/defences/mp3_compression_channelized.py
yusong-tan: Actually, as of ART 1.4.2, this art_experimental is no longer needed.

armory/art_experimental/attacks/kenansville_dft.py
yusong-tan: Pushed a new update to address high snr case

armory/art_experimental/defences/mp3_compression_channelized.py
yusong-tan: Deleted this file in the updated submission

kevinmerchant (approved):

PR #920, opened 2020-11-11
enable targeted object detection attacks

### LINKED ISSUE #919: Enable targeted attacks on object detectors
The label targeter classes in [labels.py](https://github.com/twosixlabs/armory/blob/master/armory/utils/labels.py) are not equipped to handle the fact that object detection labels are dictionaries, as opposed to numpy arrays.

lcadalzo:
Closes #919.


lcadalzo:
In the recently added `ObjectDetectionFixedLabelTargeter` class, the output is returned as a list of length 1, with the one element being a dictionary. This was done because if the returned `y_target` is a dictionary in the same form as original `y`, then the ART attack will crash [here](https://github.com/lcadalzo/armory/blob/919-targeted-od/armory/utils/labels.py#L107).

The issue with this workaround is that the original `y` and the generated `y_target` are now of a different format (the former being a dict, latter a list). This causes errors during the metric computation.

Potential solutions would be to modify the format of the original `y` to a list of dictionaries, modify ART to handle the case when `y` is a dictionary, or add conditional logic to metric computation to handle inputted `y`'s of different formats (both dict and list of dicts).

lcadalzo:
I've successfully tried, separately, options (2) and (4). Waiting to hear from ART r.e. option (3)

lcadalzo:
ART attacks expect object detection labels to be a list of dictionaries, as this is consistent with the format outputted by ART object detectors. I think it then makes sense to load object detection datasets with the labels in this format, which I'll create a separate issue for

kevinmerchant (approved):
LGTM

PR #924, opened 2020-11-11
OD metrics handle arbitrary batch size 

### LINKED ISSUE #923: Update OD Metrics for batch_size > 1 case
[This line ](https://github.com/twosixlabs/armory/blob/master/armory/utils/metrics.py#L1024) in metrics.py assumes a batch size of 1 for object detection datasets, a constraint that's no longer being enforced for xView. This will cause incorrect metric calculation as it only considers the first element in the batch.

lcadalzo:
Closes #923


ng390 (approved):
LGTM

PR #927, opened 2020-11-12
variable length bug when batch elements are same shape

### LINKED ISSUE #926: Variable length dataset loading bug
When loading a variable length dataset with batch size > 1, an error occurs when all the fetched elements are in fact of the same shape. In the xView dataset, the first two images are 200x200x3, while the third image is 300x300x3. If batch_size is set to 3 or more, the elements are correctly stored as a np 1D object array. However when batch_size is set specifically to 2, each of the 2 elements in the batch are np objects, rather than floats. Hence, [this assertion ](https://github.com/twosixlabs/armory/blob/master/armory/data/datasets.py#L508) fails.

To reproduce error, set batch_size equal to 2 in the [xView config](https://github.com/twosixlabs/armory/blob/master/scenario_configs/xview_frcnn.json#L27).

lcadalzo:
Closes #926


ng390 (approved):
LGTM

Issue #336, opened 2020-04-01
Add dataset metadata to own module


davidslater:

seanpmorgan:
Enable contextual information (as possible) to scenario datasets

joemathai:
it would be nice to have this as a parameter inside _generator_from_tfds()

davidslater:
In particular, provide label strings and any metadata about dataset statistics (mean and std for z-norm).

PR #928, opened 2020-11-12
Save adversarial examples for listening/viewing

### LINKED ISSUE #876: Saving attack audio waveforms / images / videos for listening / viewing

kevinmerchant:
Closes #876


kevinmerchant:
Currently, when saving audio or video samples, we assume a sampling rate or frame rate.  For video datasets, this is less of an immediate issue since we currently only support UCF101 with a frame rate of 25 Hz.  For audio, we currently have several datasets with different sampling frequencies, so the assumption of 16 kHz is not always correct.

Probably the easiest solution is to address #336 so that dataset metadata is more accessible. We can then pass this metadata to the `SampleExporter` class to save audio/video at the correct rate.

ng390 (approved):
LGTM

PR #937, opened 2020-11-16
auto-expand tarballs submitted as weights files

### LINKED ISSUE #921: Enable tar files passed in as model_weights to be automatically expanded

adamj26:
Closes #921

### DIFF
--- a/armory/data/utils.py
+++ b/armory/data/utils.py
@@ -92,11 +92,8 @@ def maybe_download_weights_from_s3(
                 # check if the tarfile contains a directory containing all its members
                 # ie if the tarfile expands out entirely into a subdirectory
                 dirs = [fi.name for fi in tar.getmembers() if fi.isdir()]
-                for d in dirs:
-                    if all(name.startswith(d) for name in tar.getnames()):
-                        subdir = d
-                        break
-                else:
+                commonpath = os.path.commonpath(tar.getnames())
+                if not commonpath or commonpath not in dirs:
                     raise PermissionError(
                         (
                             f"{weights_file} does not expand into a subdirectory."
--- a/armory/data/utils.py
+++ b/armory/data/utils.py
@@ -33,10 +33,13 @@ def add_checksums_dir(dir):
     CHECKSUMS_DIRS.append(dir)
 
 
-def maybe_download_weights_from_s3(weights_file: str) -> str:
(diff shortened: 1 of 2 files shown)


armory/data/utils.py
```diff
@@ -81,6 +84,29 @@ def maybe_download_weights_from_s3(weights_file: str) -> str:
...
+
+    if auto_expand_tars:
+        if tarfile.is_tarfile(filepath):
+            logger.debug(f"Detected model weights file {weights_file} as a tar archive")
+            with tarfile.open(filepath) as tar:
+                # check if the tarfile contains a directory containing all its members
+                # ie if the tarfile expands out entirely into a subdirectory
+                dirs = [fi.name for fi in tar.getmembers() if fi.isdir()]
+                for d in dirs:
+                    if all(name.startswith(d) for name in tar.getnames()):
```
davidslater, line 96: If two subdirs share a common prefix, like "test" and "test2", this will return "test", even though there are two subdirectories. This is because `startswith` is just a string comparison. I propose doing this with `os.path`:

```
dirs = [fi.name for fi in tar.getmembers() if fi.isdir()]
commonpath = os.path.commonpath(tar.getnames())
if not commonpath or commonpath not in dirs:
    raise PermissionError(...)

```

armory/data/utils.py
```diff
@@ -81,6 +84,29 @@ def maybe_download_weights_from_s3(weights_file: str) -> str:
...
+                        break
+                else:
+                    raise PermissionError(
+                        (
+                            f"{weights_file} does not expand into a subdirectory."
+                            f" Weights files submitted as tarballs must expand into a subdirectory."
+                        )
+                    )
+                logger.info(f"Auto-unpacking model weights from {weights_file}")
+                tar.extractall(path=saved_model_dir)
```
davidslater, line 107: Some of these tar files are bound to be rather large - it would be rather painful, I think, to have to expand them each time you run armory. If running locally, you will repeatedly extract/overwrite those files.

I would rather check to see if the subdir exists and then log a warning like `f"Subdir already exists. Not expanding tar. Must remove {os.path.join(saved_model_dir, subdir)} manually."` and then skip extraction.

davidslater (changes requested):
Two minor changes with comments.

davidslater (approved):
LGTM

PR #941, opened 2020-11-17
object detection label format

### LINKED ISSUE #932: Make object detection label format consistent
Currently, object detection labels are loaded as a dictionary with keys that map to np object arrays when batch_size > 1, or np.float32 arrays when batch_size=1. However, the labels outputted by ART's PyTorchFasterRCNN are a list of dictionaries, one dictionary for each image in the batch. This is also the format that ART attacks expect.

Loading the labels in this format would enable targeted attacks as well as `use_label: true`. In other words, it'd enable passing a `y` value to `generate()` besides `None`. It also simplifies things because the current (loaded) labels take on a different format depending on whether batch_size > 1 or not. There would be no need for np object arrays if using a list of dictionaries

lcadalzo:
Closes #932


kevinmerchant (approved):
LGTM

PR #942, opened 2020-11-17
frame saliency

davidslater:


lcadalzo:
After briefly looking through the [attack paper](https://arxiv.org/pdf/1811.11875.pdf), it seems like applying this to the Ucf101 scenario might be more fitting?

lcadalzo:
added a config for this. If I'm missing context though and there's a reason resisc was chosen, I'll delete. If not, and the purpose is simply to show an example usage of the attack, I think we can delete the resisc config

davidslater:
Yeah, that makes sense. I was just doing a simple test and didn't want to incur video compute costs.

davidslater:
If you're good, go ahead and merge.

lcadalzo:
I'm going to remove the resisc frame saliency config since I assume we don't need two examples, and the attack is more relevant for ucf

lcadalzo (approved):
LGTM

PR #944, opened 2020-11-18
Fix TORCH_HOME in --no-docker mode

### LINKED ISSUE #938: TORCH_HOME set incorrectly for --no_docker
In `__main__.run()`, the Evaluator object is instantiated before `paths.set_mode()` is ever run. This results in `rig.extra_env_vars["TORCH_HOME"]` being set to "/armory", rather than "~/.armory", in the --no_docker case, causing permission errors. This should be a simple one-line fix, by adding the line:

`paths.set_mode("host" if args.no_docker else "docker")`

before 
`rig = Evaluator(config, no_docker=args.no_docker, root=args.root)`

davidslater:
Fixes #938

### DIFF
--- a/armory/eval/evaluator.py
+++ b/armory/eval/evaluator.py
@@ -125,7 +125,12 @@ class Evaluator(object):
 
         # Because we may want to allow specification of ARMORY_TORCH_HOME
         # this constant path is placed here among the other imports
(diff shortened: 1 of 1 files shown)


davidslater:
runtime_paths are always set to docker mode as noted in the bug report

lcadalzo (approved):
LGTM

PR #945, opened 2020-11-18
Update HOME directory in container for jupyter to work in non-root mode

### LINKED ISSUE #680: Allow Jupyter to run as non-root user in container
Currently, there are permissions problems, as it attempts to write to `/.local`, which is not allowed when in a non-root account. 

This is likely better fixed by an overhaul of how the docker container and user accounts are set up on it. 

One approach would be to create a specific user account inside the container, and then if a different user needs to be mapped in, use these sort of mappings: https://seravo.fi/2019/align-user-ids-inside-and-outside-docker-with-subuser-mapping

### LINKED ISSUE #933: Editing files in Armory/JupyterLab instance saves them as root even when --root flag unused
When I create or modify files in a jupyterlab instance launched with `armory launch pytorch --jupyter` they are being saved with `root` permissions, but I don't have root permissions on my workstation (which causes problems when I need to subsequently manipulate the files outside of docker). 

Note that I am __not__ using the `--root` flag, so this seems to be a bug.

Armory version 0.12.1

davidslater:
Fixes #933
Fixes #680

### DIFF
--- a/armory/eval/evaluator.py
+++ b/armory/eval/evaluator.py
@@ -126,6 +126,8 @@ class Evaluator(object):
         # Because we may want to allow specification of ARMORY_TORCH_HOME
         # this constant path is placed here among the other imports
         self.extra_env_vars["TORCH_HOME"] = paths.runtime_paths().pytorch_dir
+        if not self.no_docker:
+            self.extra_env_vars["HOME"] = "/tmp"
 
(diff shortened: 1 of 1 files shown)


lcadalzo:
@davidslater if my previous commit addressing the merge conflict looks fine, we can merge

davidslater:
Yeah, that looks good.

lcadalzo (approved):
LGTM

PR #946, opened 2020-11-18
SNR: remove constraint on 1D inputs

### LINKED ISSUE #934: Extend SNR Metric To Images
Currently, the SNR metric assumes 1D (audio) data.  This should be extended to 2D data, in particular so it can be computed for perturbed SAR channels on the so2sat dataset.

davidslater:
Fixes #934

Let me know if this is what you're thinking. To test, try:
```
>>> from armory.utils import metrics
>>> import numpy as np
>>> metrics.snr(*[np.random.random((1, 400)) for i in range(2)]). # 1D inputs
[2.1268499851463973]
>>> metrics.snr(*[np.random.random((1, 20, 20, 3)) for i in range(2)]) # 3-channel image inputs
[2.0411371110126844]
```

It should work for any shape so long as the values are comparable.


kevinmerchant (approved):
LGTM

PR #948, opened 2020-11-18
Parameterize so2sat Scenario

### LINKED ISSUE #835: so2sat different scenarios
We should make the masking of different sensors more explicit, so that we can verify that an attack only worked on SAR or optical.

### LINKED ISSUE #935: Parameterize so2sat Scenario
The so2sat scenario currently allows a mask to be passed which allows control over which channels/modalities are perturbed.  However, this does not enable separate perturbation metrics for SAR and optical channels.  We should parameterize the so2sat scenario to allow control over which modality (or both) are attacked and metric computation per modality.

kevinmerchant:
Closes #935, closes #835


armory/scenarios/multimodal_so2sat_scenario.py
```diff
@@ -0,0 +1,296 @@
...
+from armory.scenarios.base import Scenario
+
+logger = logging.getLogger(__name__)
+
+
+class So2SatClassification(Scenario):
+    def __init__(self, **kwargs):
+        if "attack_modality" not in kwargs.keys():
+            raise ValueError("`attack_modality` must be defined for So2Sat scenario")
+        if kwargs["attack_modality"] is None or kwargs["attack_modality"] not in (
```
lcadalzo, line 31:
```suggestion
        if kwargs["attack_modality"] is None or kwargs["attack_modality"].lower() not in (
```

lcadalzo: Without `.lower()`, passing "SAR", "EO", or "Both", which is what the error message says is allowed, will actually raise an error

lcadalzo (approved):
LGTM

PR #949, opened 2020-11-18
Clean label backdoor attack scenario

### LINKED ISSUE #950: Clean label poisoning

ng390:
Fixes #950 
Clean label backdoor attack scenario
- [x] Clean up docker build process / incorporate all builds into release script
- [x] Further testing with filtering defenses
Separate issue:
-Hyperparameter tuning so targeted attack success rate is higher.


armory/scenarios/poisoning_gtsrb_clbd.py
```diff
@@ -0,0 +1,284 @@
...
+            src_img = src_imgs[idx]
+            p_img, p_label = attack.poison(src_img, [tgt])
+            poison_x.append(p_img)
+            poison_y.append(p_label)
+        else:
+            poison_x.append(src_imgs[idx])
+            poison_y.append(src_lbls[idx])
+    poison_x, poison_y = np.array(poison_x), np.array(poison_y)
+
+    return poison_x, poison_y
```
ebubae, line 71: Take note that for this attack, we are only poisoning images in the target class. The `attack.poison` method takes the entire dataset and selects the poisoned data within the dataset, and returns the new set containing poisoned data points.

ng390: Understood. The poison_dataset wrapper is only used with the backdoor object of type PoisoningAttackBackdoor, not the PoisoningAttackCleanLabelBackdoor object.

ng390: I.e. it is used at eval time, not to poison the training data.

kevinmerchant (approved):
LGTM

PR #953, opened 2020-11-18
cascading attack

### LINKED ISSUE #867: Cascading attacks
How to produce a sequence of attacks and return the best result.

lcadalzo:
Closes #867


kevinmerchant:
I wonder if we want to make this more general.  `AutoAttack` seems to support any combination of attacks from `art.attacks.evasion`, so we could load in a list of attack names, use the methods [here](https://github.com/twosixlabs/armory/blob/master/armory/utils/config_loading.py) to load them, and then use a list of dictionaries of kwargs to parameterize them.  This would require the user to define individual kwargs for each (and identify issues like `BoundaryAttack` not having a `batch_size` parameter), but would also be more flexible.

Otherwise, LGTM

kevinmerchant (approved):
LGTM

PR #963, opened 2020-11-23
added GTSRB clean label poisoning dataset

yusong-tan:
Data to be sent separately.


ng390:
Going forward, poison data will be generated on-the-fly using appropriate ART attacks.

PR #968, opened 2020-12-01
Fixes so2sat mask bug

### LINKED ISSUE #961: Changing attack_modality in so2sat classification is not effective.
The official so2sat's scenario config specifies the modality in `scenario/kwargs/attack_modality`, but changing it to "sar", "eo", and "both" has the same adversarial accuracy. Unless we specify the `attack/generate_kwargs/channels` accordingly.

https://github.com/twosixlabs/armory-example/blob/master/official_scenario_configs/so2sat_baseline.json

My distill is as follows.

The scenario `So2SatClassification` creates the `attack_channels` from `attack/generate_kwargs/channels`, and then from `scenario/kwargs/attack_modality` if the previous one is not specified.

https://github.com/twosixlabs/armory/blob/master/armory/scenarios/multimodal_so2sat_scenario.py#L62-L74

This `attack_channels` is then passed to the attack class via `generate_kwargs["mask"]`.

https://github.com/twosixlabs/armory/blob/master/armory/scenarios/multimodal_so2sat_scenario.py#L255-L261

However, this mask is ignored in `PGDPatch`, which re-creates the channels only from `attack/generate_kwargs/channels`.

https://github.com/twosixlabs/armory/blob/master/armory/art_experimental/attacks/pgd_patch.py#L24

kevinmerchant:
Fixes #961 

I changed the `channels` argument back to the binary `mask` argument to allow compatibility with other attack types.

### DIFF
--- a/armory/art_experimental/attacks/pgd_patch.py
+++ b/armory/art_experimental/attacks/pgd_patch.py
@@ -21,8 +21,10 @@ class PGDPatch(ProjectedGradientDescent):
             raise ValueError("generate_kwargs did not define 'xmin'")
 
         assert x.ndim == 4, "This attack is designed for images (i.e. 4 dimensions)"
-        channels = generate_kwargs.get("mask", np.ones(x.shape[3], dtype=np.float32))
-        channels = np.where(channels)[0]
+        channels_mask = generate_kwargs.get(
+            "mask", np.ones(x.shape[3], dtype=np.float32)
+        )
+        channels = np.where(channels_mask)[0]
 
         mask = np.zeros(shape=x.shape[1:], dtype=np.float32)
         if "patch_ratio" in generate_kwargs:
--- a/armory/art_experimental/attacks/pgd_patch.py
+++ b/armory/art_experimental/attacks/pgd_patch.py
@@ -21,18 +21,19 @@ class PGDPatch(ProjectedGradientDescent):
             raise ValueError("generate_kwargs did not define 'xmin'")
 
         assert x.ndim == 4, "This attack is designed for images (i.e. 4 dimensions)"
-        channels = generate_kwargs.get("channels", range(x.shape[3]))
+        channels = generate_kwargs.get("mask", np.ones(x.shape[3], dtype=np.float32))
+        channels = np.where(channels)[0]
 
-        mask = np.zeros(shape=x.shape[1:])
+        mask = np.zeros(shape=x.shape[1:], dtype=np.float32)
         if "patch_ratio" in generate_kwargs:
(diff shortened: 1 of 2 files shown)


armory/scenarios/multimodal_so2sat_scenario.py
```diff
@@ -60,35 +60,42 @@ def _evaluate(
...
+                    axis=0,
+                )
             elif self.attack_modality == "both":
                 logger.info("No mask configured. Attacking all SAR and EO channels")
-                attack_channels = range(14)
+                attack_channels = np.ones(14, dtype=np.float32)
 
         else:
             assert isinstance(
                 attack_channels, list
```
lcadalzo, line 84: I think it would make sense here to also assert that the list is (1) of length 14 and (2) only contains 0/1. This would catch when a user incorrectly tries to specify channels e.g. [0,1,2,3] rather than a mask [1, 1, 1, 1, 0, ..., 0]

armory/scenarios/multimodal_so2sat_scenario.py
```diff
@@ -60,35 +60,42 @@ def _evaluate(
             estimator = load_defense_internal(config["defense"], estimator)
 
         attack_config = config["attack"]
-        attack_channels = attack_config.get("generate_kwargs", {}).get("channels")
+        attack_channels = attack_config.get("generate_kwargs", {}).get("mask")
```
lcadalzo, line 63: minor comment, but I think it'd be a little clearer if this variable name was something like `attack_channels_mask` rather than `attack_channels`. The former name I would expect implies a binary array (of length 14) while the latter name to me implies values ranging from (0,13) of length equal to the number of channels being attacked

armory/art_experimental/attacks/pgd_patch.py
```diff
@@ -21,18 +21,19 @@ def generate(self, x, y=None, **generate_kwargs):
             raise ValueError("generate_kwargs did not define 'xmin'")
 
         assert x.ndim == 4, "This attack is designed for images (i.e. 4 dimensions)"
-        channels = generate_kwargs.get("channels", range(x.shape[3]))
+        channels = generate_kwargs.get("mask", np.ones(x.shape[3], dtype=np.float32))
+        channels = np.where(channels)[0]
```
lcadalzo, line 25:
```suggestion
        channels_mask = generate_kwargs.get("mask", np.ones(x.shape[3], dtype=np.float32))
        channels = np.where(channels_mask)[0]
```

lcadalzo (approved):
LGTM

PR #973, opened 2020-12-03
Masked PGD attack for UCF101

yusong-tan:


armory/art_experimental/attacks/pgd_patch.py
```diff
@@ -3,13 +3,21 @@
...
+    Apply Masked PGD to image and video inputs,
+    where images are assumed to have shape (NHWC)
+    and video are assumed to have shape (NFHWC)
+    """
+
     def __init__(self, estimator, **kwargs):
         super().__init__(estimator=estimator, **kwargs)
         assert (
             not self.estimator.channels_first
         ), "This attack currently only supports channels last format"
```
lcadalzo, line 16: Only tangentially related but while we're here: I believe we discussed this assertion and determined that it's unnecessary, since inputs are assumed to be in channels-last format (regardless of the estimator being used). Do you agree/am I remembering correctly? If so, we should remove the assertion

yusong-tan: Yes, that's correct. Please remove the assertion.

lcadalzo (approved):
LGTM

PR #979, opened 2020-12-09
mAP (minor) fix

lcadalzo:
Because of the [epsilon in the denominator](https://github.com/twosixlabs/armory/blob/master/armory/utils/metrics.py#L627), the values of the `recalls` variable could never equal 1. The final AP is computed by measuring precision at 11 different recall points (0, 0.1, 0.2, ..., 1). Notice that precision is computed at a recall point of 1. Because of the issue mentioned in the first sentence, there will never be a recall value equal to 1 and thus [these two lines](https://github.com/twosixlabs/armory/blob/master/armory/utils/metrics.py#L637-L638) will always be reached.

Still todo:
1. update results in scenario docs
2. Add a test for mAP


lcadalzo:
xView mAP on 1000 benign samples (on `mAP-fix` branch): 27.61%
xView mAP on 1000 benign samples (on master acaa39d): 26.69%

lcadalzo:
APRICOT mean `apricot_patch_targeted_AP_per_class` (on mAP-fix branch): 5.70%
APRICOT mean `apricot_patch_targeted_AP_per_class` (on master acaa39d): 5.52%

lcadalzo:
Removing WIP and creating a separate issue #983 for updating mAP values in the docs since it will take some time to run experiments.

kevinmerchant (approved):
LGTM

PR #981, opened 2020-12-16
SNR PGD attack

davidslater:


davidslater:
I get a strange error when I run this, and I can't figure it out:
```
2020-12-16 02:48:26 c1bfcec1a360 absl[8] INFO Constructing tf.data.Dataset for split test, from /armory/datasets/librispeech_dev_clean_split/plain_text/1.1.0
Attack:   0% 0/1 [00:00<?, ?it/s]
7.6229554481487565 3.8114777240743782 [[0.         0.         0.         ... 0.00039673 0.00045776 0.00045776]] None {}
2020-12-16 02:48:26 c1bfcec1a360 art.attacks.evasion.projected_gradient_descent.projected_gradient_descent[8] INFO Creating adversarial samples.
/workspace/art/estimators/classification/pytorch.py:525: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
x_grad = torch.tensor(x).to(self._device)
/workspace/art/estimators/classification/pytorch.py:526: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
y_grad = torch.tensor(y).to(self._device)
Attack:   0% 0/1 [00:01<?, ?it/s]
2020-12-16 02:48:27 c1bfcec1a360 armory.scenarios.base[8] ERROR Encountered error during scenario evaluation.
File "/workspace/armory/scenarios/base.py", line 77, in evaluate
results = self._evaluate(config, num_eval_batches, skip_benign, skip_attack)
File "/workspace/armory/scenarios/audio_classification.py", line 181, in _evaluate
x_adv = attack.generate(x=x)
File "/workspace/armory/art_experimental/attacks/snr_pgd.py", line 63, in generate
return self._attack.generate(x, y=y, **kwargs)
File "/workspace/art/attacks/attack.py", line 74, in replacement_function
return fdict[func_name](self, *args, **kwargs)
File "/workspace/art/attacks/evasion/projected_gradient_descent/projected_gradient_descent.py", line 181, in generate
return self._attack.generate(x=x, y=y, **kwargs)
File "/workspace/art/attacks/attack.py", line 74, in replacement_function
return fdict[func_name](self, *args, **kwargs)
File "/workspace/art/attacks/evasion/projected_gradient_descent/projected_gradient_descent_pytorch.py", line 196, in generate
x=batch, targets=batch_labels, mask=mask_batch, eps=batch_eps, eps_step=batch_eps_step
File "/workspace/art/attacks/evasion/projected_gradient_descent/projected_gradient_descent_pytorch.py", line 250, in _generate_batch
adv_x, inputs, targets, mask, eps, eps_step, self.num_random_init > 0 and i_max_iter == 0,
File "/workspace/art/attacks/evasion/projected_gradient_descent/projected_gradient_descent_pytorch.py", line 377, in _compute_torch
perturbation = self._compute_perturbation(x_adv, y, mask)
File "/workspace/art/attacks/evasion/projected_gradient_descent/projected_gradient_descent_pytorch.py", line 277, in _compute_perturbation
grad = self.estimator.loss_gradient(x=x, y=y) * (1 - 2 * int(self.targeted))
File "/workspace/art/estimators/classification/pytorch.py", line 560, in loss_gradient
loss.backward()
File "/opt/conda/lib/python3.7/site-packages/torch/tensor.py", line 185, in backward
torch.autograd.backward(self, gradient, retain_graph, create_graph)
File "/opt/conda/lib/python3.7/site-packages/torch/autograd/__init__.py", line 127, in backward
allow_unreachable=True)  # allow_unreachable flag
RuntimeError: select(): index 0 out of range for tensor of size [0, 1] at dimension 0
Exception raised from select at /opt/conda/conda-bld/pytorch_1595629403081/work/aten/src/ATen/native/TensorShape.cpp:889 (most recent call first):
frame #0: c10::Error::Error(c10::SourceLocation, std::string) + 0x4d (0x7ff670a4077d in /opt/conda/lib/python3.7/site-packages/torch/lib/libc10.so)
frame #1: at::native::select(at::Tensor const&, long, long) + 0x347 (0x7ff6a3ab8ff7 in /opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_cpu.so)
frame #2: <unknown function> + 0xfe3789 (0x7ff6a3e9d789 in /opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_cpu.so)
frame #3: <unknown function> + 0xfd6a83 (0x7ff6a3e90a83 in /opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_cpu.so)
frame #4: at::select(at::Tensor const&, long, long) + 0xe0 (0x7ff6a3dc30f0 in /opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_cpu.so)
frame #5: <unknown function> + 0x2b62186 (0x7ff6a5a1c186 in /opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_cpu.so)
frame #6: <unknown function> + 0xfd6a83 (0x7ff6a3e90a83 in /opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_cpu.so)
frame #7: at::Tensor::select(long, long) const + 0xe0 (0x7ff6a3f4e240 in /opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_cpu.so)
frame #8: <unknown function> + 0x2a6d69d (0x7ff6a592769d in /opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_cpu.so)
frame #9: torch::autograd::generated::MaxBackward1::apply(std::vector<at::Tensor, std::allocator<at::Tensor> >&&) + 0x188 (0x7ff6a59410d8 in /opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_cpu.so)
frame #10: <unknown function> + 0x30d1017 (0x7ff6a5f8b017 in /opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_cpu.so)
frame #11: torch::autograd::Engine::evaluate_function(std::shared_ptr<torch::autograd::GraphTask>&, torch::autograd::Node*, torch::autograd::InputBuffer&, std::shared_ptr<torch::autograd::ReadyQueue> const&) + 0x1400 (0x7ff6a5f86860 in /opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_cpu.so)
frame #12: torch::autograd::Engine::thread_main(std::shared_ptr<torch::autograd::GraphTask> const&) + 0x451 (0x7ff6a5f87401 in /opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_cpu.so)
frame #13: torch::autograd::Engine::execute_with_graph_task(std::shared_ptr<torch::autograd::GraphTask> const&, std::shared_ptr<torch::autograd::Node>) + 0x37c (0x7ff6a5f84b1c in /opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_cpu.so)
frame #14: torch::autograd::python::PythonEngine::execute_with_graph_task(std::shared_ptr<torch::autograd::GraphTask> const&, std::shared_ptr<torch::autograd::Node>) + 0x3c (0x7ff6aa2aedcc in /opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_python.so)
frame #15: torch::autograd::Engine::execute(std::vector<torch::autograd::Edge, std::allocator<torch::autograd::Edge> > const&, std::vector<at::Tensor, std::allocator<at::Tensor> > const&, bool, bool, std::vector<torch::autograd::Edge, std::allocator<torch::autograd::Edge> > const&) + 0x803 (0x7ff6a5f83e53 in /opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_cpu.so)
frame #16: torch::autograd::python::PythonEngine::execute(std::vector<torch::autograd::Edge, std::allocator<torch::autograd::Edge> > const&, std::vector<at::Tensor, std::allocator<at::Tensor> > const&, bool, bool, std::vector<torch::autograd::Edge, std::allocator<torch::autograd::Edge> > const&) + 0x4e (0x7ff6aa2aebbe in /opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_python.so)
frame #17: THPEngine_run_backward(THPEngine*, _object*, _object*) + 0xa29 (0x7ff6aa2af889 in /opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_python.so)
<omitting python frames>
2020-12-15 18:48:28 mac0831.local armory.docker.management[36937] ERROR Command did not finish cleanly
2020-12-15 18:48:28 mac0831.local armory.eval.evaluator[36937] WARNING Shutting down container
2020-12-15 18:48:28 mac0831.local armory.docker.management[36937] INFO Stopping instance: c1bfcec1a3
2020-12-15 18:48:28 mac0831.local armory.eval.evaluator[36937] INFO Deleting tmp_dir /Users/david.slater/.armory/tmp/2020-12-16T024817.049203
2020-12-15 18:48:28 mac0831.local armory.eval.evaluator[36937] INFO Removing output_dir /Users/david.slater/.armory/outputs/2020-12-16T024817.049203 if empty
```

davidslater:
The standard pipeline seems to work fine, and I don't see myself doing anything weird other than setting eps and eps_step before using the underlying attack.

(NOTE: this may require ART 1.5)

davidslater:
That is a long way of saying I could use some help debugging.

kevinmerchant:
Digging into this a bit, this is a result of NaNs appearing during the gradient calculation [here](https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/a9464796aead77132b45211a4fbd4e637ab53da4/art/attacks/evasion/projected_gradient_descent/projected_gradient_descent_pytorch.py#L249l) in the first ~6000 samples, after which gradients are fine.
(Note: I'm running ART 1.4.2, but the behavior is the same)

I'm guessing this didn't cause issues before because if `self.norm == np.inf`, the `grad.sign` converts NaN to 0, but it's popping up now when `self.norm == 2` because the subsequent normalization forces the gradient to be NaN everywhere.

It would seem that the best solution would be to change the gradient behavior of this model so we do not get any NaNs.

kevinmerchant:
I looked into it a bit more. The SincNet model splits the signal into windows of 3000 samples [here](https://github.com/twosixlabs/armory/blob/bac56afd8f88b8876eb185251a0fe6e38fb3f16c/armory/baseline_models/pytorch/sincnet.py#L27)

If every sample in a given window is 0, the gradients for the samples within that window will be NaN.  The first input sample when I run this scenario is padded with ~6400 zeros, causing the first 6000 gradient values to be NaN.

davidslater:
It looks like it's an issue on the ART side, then?

I think it's probably also an issue if I try to set epsilon to 0 due to signal power being 0.

kevinmerchant:
ART is being handed NaN gradients from the model itself, so while something could be done there to not propagate the NaNs forward, I think the true issue lies with the SincNet model.  The easiest solution, though not necessarily the best, might be to override `loss_gradient` of the `PyTorchClassifier` object [here](https://github.com/twosixlabs/armory/blob/bac56afd8f88b8876eb185251a0fe6e38fb3f16c/armory/baseline_models/pytorch/sincnet.py#L280), and in the new `loss_gradient` call `super().loss_gradient(x, y)`, and then replace gradient elements corresponding to all-zero windows with zeros.  This would have the effect of limiting the attack to the non-zero-padded portion of the signal, which is probably fair.

davidslater:
I see the issue, thanks. I disagree with the solution, though, as I think that having NaN values in gradients is hard to completely avoid. It would be good for attacks to handle this more gracefully.

I think that I can patch the `_apply_perturbation` function in PGD so that it more gracefully handles NaN and Inf values. This error pops up on the second pass. (After the NaNs have been added to the input)

davidslater:
@kevinmerchant should be ready for another review

armory/art_experimental/attacks/snr_pgd.py
```diff
@@ -0,0 +1,147 @@
...
+    def __init__(
+        self, estimator, norm="snr", eps=10, eps_step=0.5, batch_size=1, **kwargs
+    ):
+        if batch_size != 1:
+            raise NotImplementedError("only batch_size 1 supported")
+        super().__init__(estimator, norm=2, batch_size=1, **kwargs)
+
+        # Map to SNR domain
+        eps = float(eps)
+        if norm == "snr":
```
kevinmerchant, line 38: I think the logic here might be reversed. If `norm == "snr"`, we're already on a linear scale so I think we want `snr = eps`, and in the case where `norm == "snr_db"`, we want to do the conversion `snr = 10 ** (eps / 10)` to go from dB to a linear scale.

armory/art_experimental/attacks/snr_pgd.py
```diff
@@ -0,0 +1,147 @@
...
+        else:
+            if not torch.isfinite(normalization).all():
+                logger.warning(
+                    "Some gradient values are infinite. Perturbation will be 0!"
+                )
+            perturbation = grad * eps_step / normalization
+        perturbation[torch.isnan(perturbation)] = 0.0
+
+        x = x + perturbation
+        x[torch.isnan(perturbation)] = 0.0
```
kevinmerchant, line 138: This should be extraneous due to zeroing out NaNs in `perturbation` a few lines earlier.

armory/art_experimental/attacks/snr_pgd.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def generate(self, x, y=None, **kwargs):
+        x_l2 = np.linalg.norm(x, ord=2)
+        if x_l2 == 0:
+            logger.warning("Input all 0. Not making any change.")
+            return x
+        elif self.snr_sqrt_reciprocal == 0:
+            return x
+
+        eps = x_l2 * self.snr_sqrt_reciprocal
```
kevinmerchant, line 68: I think the l2 `eps` bound should be the ratio of the l2 norm of x to the desired snr (i.e. `x_l2 / snr` without the `sqrt`)

armory/art_experimental/attacks/snr_pgd.py
```diff
@@ -0,0 +1,147 @@
...
+        elif snr == 0:
+            self.snr_sqrt_reciprocal = np.inf
+        elif snr == np.inf:
+            self.snr_sqrt_reciprocal = 0
+        else:
+            self.snr_sqrt_reciprocal = 1 / np.sqrt(snr)
+
+        eps_step = float(eps_step)
+
+        if not (0 < eps_step <= 2):
```
kevinmerchant, line 56: I'm not sure I understand when `eps_step = 2` is valid, since then it will exceed `eps` (since `eps_step` is later updated by multiplying by `eps`). It also seems to be throwing internal ART errors when I set this to 2.  Should this be limited to 1?

kevinmerchant (changes requested):
A few comments that could use another look, but otherwise LGTM.  As is, I am observing adversarial samples with SNRs below the bound.  With the below suggested changes, the bound does not appear to be exceeded.

armory/art_experimental/attacks/snr_pgd.py
davidslater: Limited to 1 now, may bring it up with ART.

armory/art_experimental/attacks/snr_pgd.py
davidslater: Done.

davidslater:
Output values with snr and snr_db metrics look good now, ready for another review @kevinmerchant

kevinmerchant (approved):
LGTM

PR #987, opened 2021-01-08
snr range attack

davidslater:
Attacks that search for the optimal attack snr. Could use better naming conventions.

### DIFF
--- a/armory/art_experimental/attacks/snr_pgd.py
+++ b/armory/art_experimental/attacks/snr_pgd.py
@@ -6,6 +6,172 @@ import logging
 logger = logging.getLogger(__name__)
 
 
+class SNR_PGDRange:
+    """
+    Finds the maximum SNR for each sample (if possible), via binary search.
+
+    Example attack config:
+        "attack": {
+            "kwargs": {
+                "batch_size": 1,
+                "eps_range": [
+                    10,
+                    50
+                ],
+                "eps_step": 0.5,
+                "max_iter": 10,
+                "norm": "snr_db",
+                "num_random_init": 2,
+                "targeted": false,
+                "verbose": false
+            },
+            "module": "armory.art_experimental.attacks.snr_pgd",
+            "name": "SNR_PGDRange",
+            "use_label": true
+        }
+    """
+
+    def __init__(self, estimator, eps_range=(0, 10, 20, 30, 40, 50), **kwargs):
+        if "eps" in kwargs:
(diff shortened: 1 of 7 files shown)


davidslater:
I also wonder if we just want to make generic versions of these classes.

armory/art_experimental/attacks/snr_pgd.py
```diff
@@ -1,11 +1,235 @@
...
+
+        x_adv = self.min_attack.generate(x, y, **kwargs)
+        if self.estimator.predict(x_adv).argmax() != y:
+            logger.info(f"Success at lower boundary eps = {self.eps_min}")
+            lower_eps = self.eps_min
+            x_best = x_adv
+            eps_best = lower_eps
+        else:
+            logger.info(f"Failure at lower boundary eps = {self.eps_min}")
+            return x_adv
```
kevinmerchant, line 165: Wondering if we want to return `x` instead of `x_adv` to match the behavior of `SNR_PGDRange`

armory/art_experimental/attacks/snr_pgd.py
```diff
@@ -1,11 +1,235 @@
...
+        self.estimator = estimator
+        self.eps_range = sorted(eps_range)
+        if len(eps_range) < 2:
+            raise ValueError("Please select multiple values for eps_range")
+        self.attacks = []
+        for eps in eps_range:
+            self.attacks.append(SNR_PGD(estimator, eps=eps, **kwargs))
+
+    def generate(self, x, y=None, **kwargs):
+        if y is None:
```
kevinmerchant, line 49: Me may want to assert that `kwargs["targeted"] == False` both here and in the `SNR_PGDRange2` variant. Otherwise, the logic of detecting attack success/failure by comparing to `y` will be backwards.

kevinmerchant (approved):
Two very minor comments, but otherwise LGTM. Do we want to add a Numpy/Linf variant of the attack?

davidslater:
@kevinmerchant ready for another quick review. The code is messy, but needed for the evaluation. We can refactor in the next cycle.

davidslater:
I can't seem to expand the reasons for failure on the pytest TF1. It seems like it ran for a super long time.

kevinmerchant:
I relaunched the tests, and while it started giving some output and looked to be moving along just fine, it did eventually fail without an error and now won't show any output.  I'll try running the TF1 tests locally to see what happens.

kevinmerchant:
The tests finish cleanly locally

PR #1003, opened 2021-03-04
add indexing to datasets

davidslater:
Currently, you can add a numeric index to loading a dataset, like this:
```
import logging
import coloredlogs
coloredlogs.install(logging.DEBUG)
import numpy as np
from armory.data import datasets
ds = datasets.mnist("test", shuffle_files=False, index=None)
print(np.hstack([next(ds)[1] for i in range(10)]))
# [2 0 4 8 7 6 0 6 3 1]
ds = datasets.mnist("test", shuffle_files=False, index=[1, 3, 5, 7, 9])
print(np.hstack([next(ds)[1] for i in range(5)]))
# [0 8 6 6 1]
```

I need to check to make sure that:
- [x] this works with config files
- [x] enable handling of simple slices
- [x] update documentation
- [x] update ArmoryDataGenerator size
- [x] Add tests

I'd be happy for any thoughts at this point.

### DIFF
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -288,7 +288,7 @@ def filter_by_index(dataset: "tf.data.Dataset", index: list, dataset_size: int):
     def enum_index(i, x):
         i = tf.expand_dims(i, 0)
         out, _ = tf.raw_ops.ListDiff(x=i, y=index_tensor, out_idx=tf.int64)
-        return tf.size(out) == 0
+        return tf.equal(tf.size(out), 0)
 
     return dataset.enumerate().filter(enum_index).map(lambda i, x: x), indexed_size
 
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -287,6 +287,7 @@ def _generator_from_tfds(
     framework: str = "numpy",
     lambda_map: Callable = None,
     context=None,
+    index=None,
 ) -> Union[ArmoryDataGenerator, tf.data.Dataset]:
     """
     If as_supervised=False, must designate keys as a tuple in supervised_xy_keys:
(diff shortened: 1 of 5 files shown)


davidslater:
@ng390 I think it's ready for another review.

armory/data/datasets.py
```diff
@@ -269,6 +269,72 @@ def parse_split_index(split: str):
     return "+".join(output_tokens)
 
 
+def filter_by_index(dataset: "tf.data.Dataset", index: list, dataset_size: int):
```
ng390, line 272: Very minor point - if dataset_size is expected to be an int, why is it cast to an int later? Should it be a Union of int and another type?

armory/data/datasets.py
```diff
@@ -269,6 +269,72 @@ def parse_split_index(split: str):
...
+        raise ValueError("index must have at least one value")
+    elif index[0] < 0:
+        raise ValueError("index values must be nonnegative")
+
+    index_tensor = tf.constant(index, dtype=tf.int64)
+
+    def enum_index(i, x):
+        i = tf.expand_dims(i, 0)
+        out, _ = tf.raw_ops.ListDiff(x=i, y=index_tensor, out_idx=tf.int64)
+        return tf.size(out) == 0
```
ng390, line 291:
```suggestion
        return tf.equal(tf.size(out), 0)
```

ng390: I think this is the compatibility issue that is causing tf1 pytest to fail -- checking the tf1 and tf2 tests with the change locally now.

ng390: With the change, the tests in tests/test_docker/test_dataset.py pass locally on the tf1, tf2, and pytorch containers

davidslater: Updated.

armory/data/datasets.py
davidslater: It's primarily so that if the input is not an int, it will fail fast.

davidslater:
Okay, pushed change. Any further review?

ng390:
LGTM -- good to drop the WIP label?

ng390 (approved):

PR #1019, opened 2021-03-11
Filter by class

### LINKED ISSUE #1013: Filter datasets by class for universal patch attacks?
Unclear what is needed for object detection datasets. (Probably don't want OD for this release)

lcadalzo:
Closes #1013


armory/data/datasets.py
```diff
@@ -290,7 +301,39 @@ def enum_index(i, x):
...
+        )
+
+    def _filter_by_class(x, y, classes_to_keep=tf.constant(class_ids, dtype=tf.int64)):
+        isallowed_array = tf.equal(classes_to_keep, tf.cast(y, tf.int64))
+        isallowed = tf.reduce_sum(tf.cast(isallowed_array, tf.int64))
+        return tf.greater(isallowed, tf.constant(0, dtype=tf.int64))
+
+    filtered_ds = dataset.filter(_filter_by_class)
+
+    if tf.executing_eagerly():
```
lcadalzo, line 326: Filtering by class is already slow since it has to iterate over the whole dataset when `dataset.filter(_filter_by_class) is called. This process below of computing the filtered dataset size slows things down even further by iterating over the whole (filtered) dataset. It's also a bit messy that the approach differs based on the value of `tf.executing_eagerly()`. 

Ideally there would be a way to keep track of the filtered dataset size during the filtering of the dataset. Leaving this PR as WIP until I'm more confident in whether this is doable

lcadalzo: removing the WIP since I can't find a way around this. 

For most datasets it's relatively quick to filter the test set to a single class, < 3 seconds. However, it's taking ~20sec for So2Sat and ~2min for UCF
So2Sat: ~20sec
UCF: ~140sec

davidslater: This seems pretty reasonable for now. It took me 15 sec on so2sat and 44 sec on ucf101.

However, I see a big problem if this is done on a train split. Should we put a warning?

davidslater (changes requested):
A couple of minor changes below.

Also, if the "class_ids" key is set in dataset in the config, AND fit is True, then it will filter both the training dataset and the evaluation dataset. I think that for now we should add a check in scenario.evaluate to check to see if that condition is true and raise an error.

lcadalzo:
a warning if class-filtering is enabled while using train split? I think that's a good idea

tests/test_docker/test_dataset.py
```diff
@@ -98,6 +98,50 @@ def test_filter_by_index():
...
 
 
+def test_filter_by_class():
+    with pytest.raises(ValueError):
+        datasets.cifar10("test", shuffle_files=False, class_ids=[])
+
+    ds_filtered = datasets.cifar10("test", shuffle_files=False, class_ids=[3])
+    assert len(list(ds_filtered)) == 1000
+    for x, y in ds_filtered:
+        assert int(y) == 3
```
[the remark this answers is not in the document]
lcadalzo: good catch; updating

davidslater (approved):
LGTM

PR #1021, opened 2021-03-15
DAPRICOT integration

yusong-tan:
This is an initial draft of the DAPRICOT dataset builder. The associated draft (and incomplete) dataset and a tutorial notebook will be emailed to Two Six separately. This allows the integration with Armory to begin as soon as possible. The final version of all the relevant files will be submitted in the next few days, and they should not impact integration.


lcadalzo:
Dapricot test is failing for TF2 and PyTorch, but not TF1, because in D-APRICOT label preprocessing the type of `y_object["id"].flat_values` (see [here](https://github.com/yusong-tan/armory/blob/dapricot_dev/armory/data/adversarial_datasets.py#L339)) differs based on the docker image used. When using TF1 image, it's a numpy array, whereas for PyTorch/TF2 it's a TF eager tensor.

scenario_configs/dapricot_frcnn.json
```diff
@@ -0,0 +1,61 @@
...
+        "weights_file": null,
+        "wrapper_kwargs": {}
+    },
+    "scenario": {
+        "kwargs": {},
+        "module": "armory.scenarios.dapricot_scenario",
+        "name": "ObjectDetectionTask"
+    },
+    "sysconfig": {
+        "docker_image": "twosixarmory/tf1:0.13.0-dev",
```
ng390, line 54: Should this be tf1:0.13.0?

lcadalzo: yes it should, thanks

armory/scenarios/dapricot_scenario.py
```diff
@@ -0,0 +1,174 @@
...
+            for img_idx in range(len(y_object)):
+                y_i_target = y_target[img_idx]
+                y_i_pred = y_pred_adv[img_idx]
+                metrics_logger.update_task(
+                    [y_i_target], [y_i_pred], adversarial=True, targeted=True
+                )
+
+            metrics_logger.update_perturbation(x, x_adv)
+            if sample_exporter is not None:
+                sample_exporter.export(x, x_adv, y, y_pred_adv)
```
ng390, line 171: sample_exporter.export(x[0], x_adv, y, y_pred_adv)

ng390: Not sure about the cleanest logic to save the correct number of samples since sample_exporter.export treats each input as a batch of 3 for the dapricot_dev_adversarial dataset. Perhaps a small note in docs/scenarios.md would suffice?

ng390 (approved):
Minor update on export_samples to avoid a shape mismatch between x and x_adv, but otherwise LGTM.

davidslater:
If the shape is mismatched on `export_samples`, why does it work on `metrics_logger.update_perturbation(x, x_adv)`?

lcadalzo:
> If the shape is mismatched on `export_samples`, why does it work on `metrics_logger.update_perturbation(x, x_adv)`?

I'm looking into this now but at a quick glance think that both sample export and perturbation update can be moved into same loop as where `update_task()` is called

lcadalzo:
> If the shape is mismatched on `export_samples`, why does it work on `metrics_logger.update_perturbation(x, x_adv)`?

to answer your question: numpy will allow you to subtract a (1, 3, H, W, 3) array by a (3, H, W, 3) array, which is what was happening. But this will be a moot point after my next commit

PR #1023, opened 2021-03-18
Micronnet model using pytorch sequential api

### LINKED ISSUE #1022: Integrate pytorch version of MicronNet classifier
Needed to complete attack for #1012 -- different from #1011 because only a single model is needed.

ng390:
Closes #1022


mwartell (approved):
LGTM. Three of three test configurations passed with > 0.8 accuracy.

PR #1026, opened 2021-03-23
WIP Bullseye polytope scenario

### LINKED ISSUE #1012: Updated scenario on the bullseye polytope attack?

ng390:
Fixes #1012


davidslater:
[ ] Allow attack to skip poisoning
[ ] More natural loading of model
[ ] Include gt/confusion matrix
[ ] Metrics improvements?

davidslater:
Blocked on ART.

davidslater:
Looking at the PR diff now, this is going to be really hard to dig into what this update is supposed to do. I'm considering closing it. Thoughts, @ng390 ?

ng390:
Agreed that the diff of this PR is a mess now that it is so dated. Some guts of the code could be ripped out for any scenarios using the bullseye polytope attack (or any clean label triggerless attack).

A lot of the structure was in place for an end-to-end attack. Since we decided end-to-end performance is not strong enough to evaluate attacks and a transfer learning attack is needed, most of the guts of this will need to change. 

There may be some convenience functions that are useful, especially for pulling out subsets of the data. For a (triggerless) clean label poisoning attack, the dataset evaluated really needs 3 splits - train, test and target images. Since most datasets have just train/test, there is logic to pull out a subset of the training dataset for use as candidate target images.

ng390:
Superseded by #1121 / contains code that is specific to an end-to-end attack rather than a fine-tuning attack.

PR #1035, opened 2021-03-25
Build tag script

davidslater:
Added command echoes and ensured that dev building also locally tagged the latest. (Otherwise, this needs to be done manually by the user.)


mwartell (approved):
LGTM

PR #1037, opened 2021-04-02
channels_first pytorch models

### LINKED ISSUE #988: channel_first for MARS model is not set correctly
MARS model has channel as its last dimension in `input_shape`, so `channel_first` should set to `False` in https://github.com/twosixlabs/armory/blob/d4d3e781cdcf01b83ce21b708f3357e538b2026c/armory/baseline_models/pytorch/ucf101_mars.py#L306 . Otherwise `channel_first` will be set to `True` by default.

### LINKED ISSUE #1036: Set `channels_first=False` in MARS classifier
In ART, the default value of `channels_first` in `PyTorchClassifier` is `True`. However, the baseline MARS model specifies the input shape to have the channels axis as the last axis:
https://github.com/twosixlabs/armory/blob/2b72bea09cdcf6e1d4d812d1a4bc68d96d2cd665/armory/baseline_models/pytorch/ucf101_mars.py#L310

This is really only a problem when one tries to apply attacks in ART that work with estimators that require `channels_first=False`.

I haven't check any other classifiers in Armory; it may be worth review this default value of `channels_first`.

lcadalzo:
Closes #988

### DIFF
--- a/armory/baseline_models/pytorch/cifar.py
+++ b/armory/baseline_models/pytorch/cifar.py
@@ -59,6 +59,7 @@ def get_art_model(
         loss=nn.CrossEntropyLoss(),
         optimizer=torch.optim.Adam(model.parameters(), lr=0.003),
         input_shape=(32, 32, 3),
+        channels_first=False,
         nb_classes=10,
         clip_values=(0.0, 1.0),
         **wrapper_kwargs,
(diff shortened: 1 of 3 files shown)


ng390 (approved):
LGTM

PR #1038, opened 2021-04-12
Resisc10 poison dataset

yusong-tan:
I've taken a first cut at integrating the Resisc dataset (10 classes, 700 images/class, 64x64x3 per image) with Armory. There may be other things that are missing, but hopefully this PR has the majority of the integration pieces in place.

Since we want to experiment with this dataset to see if it's appropriate to include for poison scenario, the merge may be better done on a dev branch, but I don't know which one is appropriate.


ng390 (approved):
LGTM

PR #1040, opened 2021-04-22
dapricot updates

lcadalzo:
(1) add targeted RobustDPatch attack, which is not yet supported in an official ART version

(2) add `dapricot_patch_target_success()` metric

(3) fix label offset issue for targeted attack by updating baseline model's `loss_gradient()`

(4) added checks to dapricot scenario (e.g. force scenario to be targeted)


ng390 (approved):
LGTM. Needs some code moved to armory-private before it is merged in.

Issue #1041, opened 2021-04-26
refactor evaluator for stepwise execution and probe-points


mwartell:
Separate the multiple stages in the execution of a configuration so:
1. phases are segregated (for example, benign and adversarial execution)
2. inter-phase results are stored for debugging or re-run
3. make clear documentation of the internal process

mwartell:
I have refactored `Scenario` subclasses to (optionally) split out the functional groups in `_evaluate`. The setup stages

1. config 
2. model
3. defense
4. fitting
5. attack
6. dataset

Have defined interfaces and do not communicate through local variables. This allows "probe-points" (pdb or jupyter) to tap in before or after these setup functions. I'm not sure interjection in those points is that compelling; it is easier to affect setup with
`config` than to monkeypatch in the middle.

The "optional" refactor allows not-yet-converted `scenarios/*.py` to work without modification. It will be non-optional once those
are converted.

Next step: refactoring metric/attack/predict block.

mwartell:
I've subtasked this issue

- [ ] #1089 
- [ ] #1090
- [ ] #1091

mwartell:
Issues #1089 and #1090 have likely been obsoleted by https://github.com/twosixlabs/armory/pull/1093/files

davidslater:
Closing, as only #1091 remains, and that is tracked on a separate milestone.

PR #1052, opened 2021-05-07
WIP: label targeter refactor

### LINKED ISSUE #1050: Refactor Label Targeters
[Current approach](https://github.com/twosixlabs/armory/blob/master/armory/utils/config_loading.py#L222-L253) relies on a growing list of hardcoded strings that map to `Targeter` classes. Configs could, instead of specifying one of these strings, point directly to the Targeter class. This would simplify things and allow users to point to their own Targeter classes, if perhaps their desired label targeting approach isn't supported in Armory

lcadalzo:
Closes #1050


lcadalzo:
One downside of the way label targeters are loaded in scenarios is that they do not fail fast. I think it will be simpler to address this once scenarios are refactored per #1041

armory/utils/config_loading.py
```diff
@@ -30,7 +30,7 @@
 from armory.art_experimental.attacks import patch
 from armory.data.datasets import ArmoryDataGenerator, EvalGenerator
 from armory.data.utils import maybe_download_weights_from_s3
-from armory.utils import labels
+from armory.utils import label_targeters
```
davidslater, line 33: I would prefer to keep the original "labels" module name.

scenario_configs/asr_deepspeech_baseline_fgsm.json
```diff
@@ -17,8 +17,11 @@
         "name": "FastGradientMethod",
         "targeted": true,
         "targeted_labels": {
-            "scheme": "string",
-            "value": "TEST STRING"
+            "args": {
+                "value": "TEST STRING"
+            },
```
davidslater, line 22: Two issues here:
1) args here is dictionary, which is confusing. I think we should probably either have this be a list so it can be handled like args, or more helpfully, turned into a kwargs dictionary (which would require modifying the label targeters to take kwargs).
2) The "args" field is never actually used by the `load_label_targeter` function. It doesn't look for config["args"]["value"]. It looks directly for config["value"]

armory/utils/config_loading.py
```diff
@@ -220,34 +220,45 @@ def load_defense_internal(defense_config, classifier):
...
+            score = config.get("score", 1.0)
+            return label_targeters.ObjectDetectionFixedLabelTargeteer(value, score)
+        else:
+            raise ValueError(
+                f'scheme {scheme} not in ("fixed", "random", "round-robin", "manual", "identity", "matched length", "object_detection_fixed")'
+            )
+    label_targeter_module = import_module(config["module"])
+    label_targeter_class = getattr(label_targeter_module, config["name"])
+    label_targeter_args = config["args"]
+    label_targeter = label_targeter_class(**label_targeter_args)
```
davidslater, line 258: This should be kwargs, not args. Otherwise it is confusing.

armory/utils/config_loading.py
```diff
@@ -220,34 +220,45 @@ def load_defense_internal(defense_config, classifier):
...
-        return labels.MatchedTranscriptLengthTargeter(transcripts)
-    elif scheme == "object_detection_fixed":
-        value = config.get("value")
-        score = config.get("score", 1.0)
-        return labels.ObjectDetectionFixedLabelTargeteer(value, score)
-    else:
-        raise ValueError(
-            f'scheme {scheme} not in ("fixed", "random", "round-robin", "manual", "identity", "matched length")'
+    if config.get("scheme"):
+        scheme = config["scheme"].lower()
```
davidslater, line 224: Can you add a "deprecation" warning on the logger?

armory/utils/label_targeters.py
```diff
@@ -111,6 +111,10 @@ def generate(self, y):
         return targeted_y
 
 
+class ObjectDetectionFixedLabelTargeteer(ObjectDetectionFixedLabelTargeter):
+    pass
+
+
```
davidslater, line 117: Why does this exist?

davidslater (changes requested):
See the comments. Also, I recommend changing the LabelTargeter classes to do keyword-only arguments following PEP3102 (https://legacy.python.org/dev/peps/pep-3102/). This would basically require adding a `*,` after `self,` in the `__init__` method.

armory/utils/config_loading.py
lcadalzo: switched it back

armory/utils/config_loading.py
lcadalzo: agreed

armory/utils/config_loading.py
lcadalzo: that's a good idea, added

scenario_configs/asr_deepspeech_baseline_fgsm.json
lcadalzo: I like the kwargs dictionary better than a list of args. R.e. (2), I did confirm that "TEST STRING" was being used. Perhaps this is a moot point after my latest commits

PR #1065, opened 2021-05-12
Resisc10 poison

yusong-tan:
Created:

1. new Resnet18 Pytorch model
2. custom affine transform function for optional data augmentation
3. new dirty-label backdoor poison scenario for RESISC-10 dataset
4. new scenario config
5. new trigger for RESISC-10


ng390 (approved):
LGTM

PR #1071, opened 2021-05-21
sweep attacks

### LINKED ISSUE #1066: Build interface for "range of epsilon" attacks
The [SNR_PGDRange](https://github.com/twosixlabs/armory/blob/master/armory/art_experimental/attacks/snr_pgd.py#L12-L80) attack iterates over a range of epsilons. It would be useful to have a convenient interface for applying this kind of approach to any attack, in order to use a sole config to generate an epsilon vs. robustness curve. Plotting utility functions should also be provided to easily convert a results JSON into a useful visualization

lcadalzo:
Closes #1066.

Two other items that this PR includes:

1. Documenting the supported values for attack config `"type"` field and throwing a warning if an unexpected value is received
2. Adding `object_detection_mAP()` metric function. This simply calls the `object_detection_AP_per_class()` function and then returns the mean across each class. SweepAttacks enforce that metrics used to measure attack success return scalars, hence the reason this was included in this PR

### DIFF
--- a/armory/art_experimental/attacks/sweep.py
+++ b/armory/art_experimental/attacks/sweep.py
@@ -125,11 +125,14 @@ class SweepAttack(EvasionAttack):
                 )
 
         if x_best is None:
-            logger.info("Attack failed. Returning original x.")
+            logger.info(
+                "Sweep attack concluded. Returning original x since attack failed at all sweep points."
+            )
             return x
         if x_best is not None:
             logger.info(
-                f"Returning best attack with kwargs {best_attack_kwargs} and generate_kwargs {best_generate_kwargs}"
+                f"Sweep attack concluded. Returning the weakest-strength successful attack with "
+                f"kwargs {best_attack_kwargs} and generate_kwargs {best_generate_kwargs}"
             )
             return x_best
 
--- a/armory/art_experimental/attacks/sweep.py
+++ b/armory/art_experimental/attacks/sweep.py
@@ -1,0 +1,209 @@
+import logging
+from importlib import import_module
+
+import numpy as np
+from art.attacks import EvasionAttack
+
+
+logger = logging.getLogger(__name__)
+
+
+class SweepAttack(EvasionAttack):
+    def __init__(self, estimator, attack_fn, sweep_params, **kwargs):
+        if not isinstance(sweep_params, dict):
+            raise ValueError(
+                "If attack_config['type'] == 'sweep', then "
+                "attack_config['sweep_params'] must be specified with "
+                "'kwargs' and/or 'generate_kwargs' field(s)."
+            )
+        for key in sweep_params:
+            if key not in ["kwargs", "generate_kwargs", "metric"]:
+                raise ValueError(
+                    f"Received an unexpected field '{key}' in attack_config['sweep_params']. "
+                    f"The list of allowable fields is as follows: "
+                    f"['kwargs', 'generate_kwargs', 'metric']."
+                )
+        if (
+            "kwargs" not in sweep_params.keys()
+            and "generate_kwargs" not in sweep_params.keys()
+        ):
+            raise ValueError(
+                "attack_config['sweep_params'] must contain at least one "
+                "of the following fields: ['kwargs', 'generate_kwargs']."
+            )
+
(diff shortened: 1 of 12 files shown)


lcadalzo:
TODO: try with an ASR config, add docs, tests.

lcadalzo:
Necessary for this PR:

- works for targeted and non-targeted attacks
- works for non-classification scenarios where measuring attack success is less black-and-white
- allows sweeping over `"kwargs"` passed to attack instantiation as well as `"generate_kwargs"` passed to `generate()` method
- must be able to specify an arbitrary number of kwargs/generate_kwargs. E.g. if I vary `"eps"` I should also be able to vary `"eps_step"` accordingly.
- exception handling
- docs

Assumptions being made:
- for an untargeted attack, an attack is successful if the metric value is below a given threshold. Opposite for targeted attacks
- the parameter lists are sorted in ascending order of attack strength. I.e. if an attack succeeds, the search will only then try values corresponding to lower indices in the parameter lists. A warning is thrown if the parameters aren't sorted in ascending order


Not (currently) supported in this PR:
- batch_size > 1
- this feature can't be used in conjunction with setting `attack_config["type"]` to `"patch"`. This is necessary for attacks whose `generate()` returns a patch rather than the input with patch.
- taking into account stochasticity of models when measuring attack success (i.e. it will only query the model one time).
- plotting utilities. Added a separate issue #1076

docs/sweep_attacks.md
```diff
@@ -0,0 +1,87 @@
...
+    }
+}
+```
+
+Each range of sweep parameters must be specified with a list of length `N`, where `N > 1` is 
+the number of desired search points. All parameters specified inside of 
+`attack_config["sweep_params"]` must correspond to lists of length `N`, and the `ith` element of
+a given kwarg list will be used in conjunction with the `ith` element of all other kwarg 
+lists (i.e. in the first example, when `"eps"` is `0.01`, `"eps_step"` is `0.005` and so on). The
+search algorithm assumes that parameters lists are in ascending order of attack strength.
```
ng390, line 58: For completeness, maybe we should mention in the docs that the attack returns the weakest strength adversarial example that is a successful attack based on the metrics, or the original input if the attack fails.

armory/utils/config_loading.py
```diff
@@ -136,10 +138,26 @@ def load_attack(attack_config, classifier):
         if targeted:
             logger.warning("Patch attack generation may ignore 'targeted' set to True")
         attack_config["kwargs"] = kwargs
+    else:
+        if attack_config.get("type") not in SUPPORTED_TYPES:
+            logger.warning(
```
ng390, line 143: To avoid a warning for users with the default configs (in the future), we could also pull out the ```config["attack"]["type"]``` field from scenario_configs/gtsrb_scenario_clbd.json and from scenario_configs/gtsrb_scenario_clbd_defended.json.

armory/art_experimental/attacks/sweep.py
```diff
@@ -0,0 +1,228 @@
...
+                i_max = i_mid
+            else:
+                # attack failure
+                i_min = i_mid + 1
+                logger.info(
+                    f"Failure with kwargs {attack_kwargs} and generate_kwargs {generate_kwargs}"
+                )
+
+        if x_best is None:
+            logger.info("Attack failed. Returning original x.")
```
ng390, line 128: It may be useful to be able to extract the actual attack_kwargs and generate_kwargs at a per sample level -- it would be slightly easer if there were a common prefix for successful and failing attacks, so that a string could be grep'ed for. This might be useful to keep in the results json as well, but that would likely break the abstraction of the attack...

docs/sweep_attacks.md
lcadalzo: added

armory/utils/config_loading.py
lcadalzo: agreed, done

armory/art_experimental/attacks/sweep.py
lcadalzo: now using common prefix `"Sweep attack concluded."`

ng390 (approved):
LGTM

PR #1080, opened 2021-05-27
Hotfix: Docker tf1 fix to allow tensorflow.keras to load h5 weights (fixes CI testing)

ng390:
(Same as #1079, but with hotfix branch name.) The tf1 container has a dated version of Tensorflow (1.15.0), which still has the following [bug](https://github.com/h5py/h5py/issues/1732) when h5py is >=3.0.0. The conda version of tensorflow still does not support tensorflow 1 >1.15.0. As a workaround can pin the h5py version to be <3.0.0.


lcadalzo (approved):
LGTM

PR #1085, opened 2021-06-01
WIP: coco integration

lcadalzo:
Closes #1069.

Beyond integrating the COCO dataset, this PR also ensures that (1) the COCO dataset, (2) the APRICOT dataset, and (3) the COCO baseline model (which is used for APRICOT too) are all 0-indexed, meaning object labels start at 0 rather than 1. 

The tfds COCO dataset comes 0-indexed out-of-the-box, as does the [baseline model](https://github.com/lcadalzo/armory/blob/1069-coco-integration/armory/baseline_models/tf_graph/mscoco_frcnn.py#L22). The APRICOT dataset came 1-indexed, though, and so previously (i.e. currently in Armory 0.13.1), the baseline model was [modified](https://github.com/twosixlabs/armory/blob/v0.13.1/armory/baseline_models/tf_graph/mscoco_frcnn.py#L54-L61) to output 1-indexed classes to avoid a label mismatch. 

This PR undoes this modification and makes everything 0-indexed. For the APRICOT dataset, this means adding a label_preprocessing function to decrement labels by 1 so that they're 0-indexed and thus align with the baseline model (and COCO dataset).

Note: this baseline model is also used in the D-APRICOT scenario, although the D-APRICOT dataset doesn't contain labels for COCO objects so no update to the D-APRICOT dataset is needed

PR #1097, opened 2021-06-08
Coco

lcadalzo:
See discussion in #1085


lcadalzo:
[Link](https://github.com/lcadalzo/armory/actions/runs/919564226) to CI tests (doesn't include a coco test, since adding a test for coco is causing CI to fail due to limited resources on the host)

ng390:
Would it be feasible to add back the test with pytest.skip logic, similar to for [Librispeech](https://github.com/twosixlabs/armory/blob/deb7a469bf4a7497d14fdd87eba6417b5e44589f/tests/test_docker/test_dataset.py#L341-L343). Then the test could be run locally even if the resources are too high for a Github runner.

lcadalzo:
[Link](https://github.com/lcadalzo/armory/actions/runs/921908098) to CI tests after adding commit 93ac02c

PR #1099, opened 2021-06-08
gardproject.org initial jekyll site

mwartell:
@davidslater This is one way get the website into easy version control.  Merge this if and only if you want it to be part of the
main armory repository from now on.

Pros:
1. it is easy to find
2. familiar armory git process can be used for changes to the website

Cons:
1. needs to be excluded from armory tool distribution (not too hard)
2. repository permissions match armory (because it is armory)
3. clutters the repo

Perhaps the better way to launch this would be to create repository twosixlabs/gardproject and localize the work to there. If so, it will take me 25 seconds to recreate this pull request there.


mwartell:
obsoleted by twosixlabs/gardproject repository

PR #1101, opened 2021-06-09
Bugfix: Integrate TFDS script

ng390:
Bugfix in the TFDS integration script - now correctly creates a tarfile with the directory structure expected by armory.


lcadalzo (approved):
LGTM

PR #1104, opened 2021-06-10
WIP: eval-update smoke test

mwartell:
Contains fixes for typo level errors in davidslater/eval-update (b9ae00c474b351fcec056ff268c04dbe316cb01e) found through scenario testing.

Incorporates davidslater/eval-update since that is the branch base.


host-requirements.txt
```diff
@@ -26,8 +26,15 @@ tensorflow-gpu==1.15.0
 
 ### Begin Pytorch required libraries
 pytorch==1.7.0
-torchvision==0.8.0
+torchvision==0.7.0
```
lcadalzo, line 29: happened to glance at this PR and was curious why the change to torchvision version? For reference, corresponding PyTorch/torchvision versions are listed in this [table](https://pypi.org/project/torchvision/)

mwartell: I was curious, too; thanks for the reference document. There was a pip version error when running `--no-docker` which may not be the proper consumer of `host-requirements.txt`. I am reverting that version regression and will better explore what the issue was.

davidslater:
@mwartell can you update for black and flake8?

host-requirements.txt
lcadalzo: I believe the packages added in 4e21bcd are only needed for the ASR scenario. Perhaps we should add a comment making this clear

armory/scenarios/scenario.py
```diff
@@ -0,0 +1,423 @@
...
+    def _load_defense(self, estimator, train_split_default="train"):
+        model_config = self.config["model"]
+        defense_config = self.config.get("defense") or {}
+        defense_type = defense_config.get("type")
+
+        if defense_type in ["Preprocessor", "Postprocessor"]:
+            logger.info(f"Applying internal {defense_type} defense to estimator")
+            estimator = config_loading.load_defense_internal(defense_config, estimator)
+
+        if model_config["fit"]:
```
lcadalzo, line 112: for the sake of modularity, would it make sense to have a separate method encapsulating model training? `_load_defense()` isn't the first place I'd expect training to live

davidslater: Yes, I agree.

lcadalzo:
Since `self.x`, `self.y`, `self.y_pred`, etc. are overwritten each iteration, would it be useful to keep track of this information with attributes that are appended to each iteration through the dataset? On one hand, it may be useful to have those artifacts accessible during interactive sessions, but on the other it may become a bit unwieldy

davidslater:
I would like to avoid those sort of append operations, which will quickly kill memory for large datasets. These are there mainly for easy interactive use for a single sample.

host-requirements.txt
davidslater: Probably should, but let's keep that out of this PR.

davidslater:
Moved to #1114

PR #1108, opened 2021-06-11
Merging 0.13.2 to dev

mwartell:
This is the result of 

     git merge -s recursive -X theirs r0.13.2

into the dev branch. 

The "mess" that you saw without the theirs strategy option was ~40 `+-2` entries which were all identically bumping up a version number.  There were other conventional merge conflicts that were resolved in the same direction. In this case (master → dev), the changes in master had truly jumped ahead of dev so preferring theirs (master / r0.13.2) was very likely correct.

I strongly advise that you review this pull request knowing that it used a strategy that you may not have seen before.


mwartell:
CI tests failures are bad enough to retract this.

PR #1112, opened 2021-06-15
improved host requirements definition for no-docker installations

mwartell:
Describes self-consistent dependencies and notes where special, non-pip actions need to be done according to local configuration.


mwartell:
Merged following my own review and test.

PR #1113, opened 2021-06-16
Optimize Kenansville attack and fixes bug

### LINKED ISSUE #1103: Are performers allowed to contribute ?
Hello,

I might have an improved(>40x speedup on a 3sec clip) version of the Kenansville attack. Is it something I could make a PR for ?

Best,

Guillaume

GuillaumeLeclerc:
Benchmark on a 3sec clip shows 40x+ speedup. More than fast enough to train models directly on attacked audio samples

Resolves #1103

Was tested outside of Armory.

### DIFF
--- a/armory/art_experimental/attacks/kenansville_dft.py
+++ b/armory/art_experimental/attacks/kenansville_dft.py
@@ -42,7 +42,7 @@ class KenansvilleDFT:
 
     def _attack(self, x):
         # Scale the threshold based on the target SNR
-        threshold = 10**(-self.snr_db / 10)
+        threshold = 10 ** (-self.snr_db / 10)
         x_fft = np.fft.fft(x)
         x_psd = np.abs(x_fft) ** 2
         # Scale the threshold based on the strenght of the signal
--- a/armory/art_experimental/attacks/kenansville_dft.py
+++ b/armory/art_experimental/attacks/kenansville_dft.py
@@ -46,7 +46,7 @@ class KenansvilleDFT:
             raise ValueError("Input must be real")
         if not len(x):
             return np.copy(x)
-        
+
         # Determine power spectral density using real FFT
         #     Double power spectral density for paired frequencies (non-DC, non-nyquist)
         x_rfft = np.fft.rfft(x)
@@ -57,14 +57,14 @@ class KenansvilleDFT:
             x_psd[1:-1] *= 2
 
         # Scale the threshold based on the power of the signal
-        # Find frequencies in order with cumulative perturbation less than threshold 
+        # Find frequencies in order with cumulative perturbation less than threshold
         #     Sort frequencies by power density in ascending order
         x_psd_index = np.argsort(x_psd)
         reordered = x_psd[x_psd_index]
         cumulative = np.cumsum(reordered)
         norm_threshold = self.threshold * cumulative[-1]
         i = np.searchsorted(cumulative, norm_threshold, side="right")
-        
+
         # Zero out low power frequencies and invert to time domain
         x_rfft[x_psd_index[:i]] = 0
         return np.fft.irfft(x_rfft).astype(x.dtype)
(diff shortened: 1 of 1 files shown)


davidslater:
@GuillaumeLeclerc is the speedup primarily due to using a quickselect algorithm in numpy instead of running through python loops?

GuillaumeLeclerc:
That plus:
- Not working in log space unnecessarily
- Doing everything that was done in the loop (square, sum, log) a single time

armory/art_experimental/attacks/kenansville_dft.py
```diff
@@ -42,56 +41,26 @@ def __init__(
...
-            noise = np.sum(x_psd[x_psd_ind[: min(id, x_len)]])
-            noise_db = 10 * np.log10(noise)
-
-        id -= 2
-
+        reordered = x_psd[x_psd_ind]
+        # Compute the cumulative perturbation size induced by zeroing frequencies up to index ix
+        # Then searching the first one that is below the threshold
+        id = np.searchsorted(np.cumsum(reordered), threshold)
+        id -= id % 2
```
kevinmerchant, line 57: In the initial implementation, `id` is guaranteed to be even immediately after the fine search, but with the change to the search, I believe the `id` returned by the search can now be odd.  Does this require any changes to how `id` is modified after this point? In particular, if `len(x)` is odd and `dc_ind < id`, do we still need to set `id -= id % 2`?

armory/art_experimental/attacks/kenansville_dft.py
```diff
@@ -42,56 +41,26 @@ def __init__(
...
         # make sure the only non-paired frequencies are DC and, if x is even, x_len/2
-        if (dc_ind in x_psd_ind[:id]) ^ (
-            x_len % 2 == 0 and x_len / 2 in x_psd_ind[:id]
-        ):
+        if (dc_ind < id) ^ (len(x) % 2 == 0 and len(x) / 2 < id):
```
kevinmerchant, line 59: The first part of this condition fixes a bug that was present in the original implementation. In the second part of this condition, `len(x) / 2` (which effectively indexes into an array of FFT bins) is compared to `id`, where `id` is an index of the reordered array. Should `len(x) / 2` also be updated to reflect the index of this bin in the reordered array?

kevinmerchant (commented):
A few comments about how the final updates to `id` are made, but otherwise this seems to be a much more efficient replacement of the original attack.

kevinmerchant:
Thinking about this a bit more, I think we may want to replace:
```
id -= id % 2
# make sure the only non-paired frequencies are DC and, if x is even, x_len/2
if (dc_ind < id) ^ (len(x) % 2 == 0 and len(x) / 2 < id):
    id -= 1
```
With something like:
```
# make sure the only non-paired frequencies are DC and, if x is even, x_len/2
if len(x) % 2:
    if (id % 2) ^ (dc_ind < id):
        id -= 1
else:
    upper_ind = np.where(x_psd_ind == int(len(x) / 2))[0][0]
    single_freq = ((dc_ind < id) ^ (upper_ind < id))
    if (id % 2) ^ single_freq:
        id -= 1
```
To test, temporarily place `assert(np.amax(np.abs(np.imag(x_ifft))) < 1e-15)` just before `return np.real(x_ifft).astype(np.float32)`.  If the frequencies which are zeroed out are balanced, `x_ifft` should be real, ignoring minor numerical imprecision.

GuillaumeLeclerc:
To be completely fair I don't have much experience with audio processing. The only thing I was doing was taking random audio samples and making sure that for every optimization I apply I get the same output as the original implementation (modulo the `dc_int` issue I changed first in the original code).

What about you just make an executive decision about the final modifications on `id`? I personally don't feel qualified to make any recommendation here.

kevinmerchant:
@yusong-tan Perhaps you could weigh in here?

davidslater:
Reading through the paper and how we're attacking this, I am going to suggest another update using real ffts.

yusong-tan:
@kevinmerchant your code looks correct. If using real ffts, then there's no need to check for non-paired frequencies when zeroing

davidslater:
@yusong-tan @kevinmerchant can you take a look at my update? I've done some testing and it looks good.

Results are slightly different than the original because `rfft` and `irfft` reproduce the original signal better than the `fft` and `ifft` for computational reasons.

kevinmerchant:
LGTM

kevinmerchant (approved):

PR #1115, opened 2021-06-17
pillow version update

davidslater:


mwartell (approved):
This hits all references to Pillow in the the repo and the changes from 8.1.1 to 8.2.0 and the version difference is expected to be "don't care" to our uses.

PR #1118, opened 2021-06-17
add/update scenario configs to reflect Eval 3

yusong-tan:
Added the following configs:

```
librispeech_asr_imperceptible_defended.json
librispeech_asr_imperceptible_undefended.json                             
librispeech_asr_kenansville_defended.json              
librispeech_asr_kenansville_undefended.json            
librispeech_asr_pgd_defended.json       
librispeech_asr_pgd_undefended.json         
librispeech_asr_pgd_multipath_channel_defended.json    
librispeech_asr_pgd_multipath_channel_undefended.json

ucf101_pretrained_flicker_defended.json
ucf101_pretrained_flicker_undefended.json
ucf101_pretrained_frame_saliency_defended.json
ucf101_pretrained_frame_saliency_undefended.json
ucf101_pretrained_masked_pgd_defended.json
ucf101_pretrained_masked_pgd_undefended.json

xview_frcnn_masked_pgd_defended.json
xview_frcnn_masked_pgd_undefended.json
xview_frcnn_robust_dpatch_defended.json
xview_frcnn_robust_dpatch_undefended.json

so2sat_eo_masked_pgd_defended.json           
so2sat_eo_masked_pgd_undefended.json         
so2sat_sar_masked_pgd_defended.json          
so2sat_sar_masked_pgd_undefended.json
```

Also updated `pgd_patch.py` and `robust_dpatch.py` to allow random patch locations.

### DIFF
--- a/scenario_configs/so2sat_sar_masked_pgd_defended.json
+++ b/scenario_configs/so2sat_sar_masked_pgd_defended.json
@@ -1,0 +1,146 @@
+{
+    "_description": "Baseline so2sat classification",
+    "adhoc": null,
+    "attack": {
+        "generate_kwargs": {
+            "patch_ratio": 0.05
+        },
+        "knowledge": "white",
+        "kwargs": {
+            "batch_size": 1,
+            "eps": 2.0,
+            "eps_step": 0.01,
+            "max_iter": 300,
+            "num_random_init": 0,
+            "random_eps": false,
+            "targeted": false,
+            "verbose": false
+        },
+        "module": "armory.art_experimental.attacks.pgd_patch",
+        "name": "PGDPatch",
+        "use_label": true
+    },
+    "dataset": {
+        "batch_size": 8,
+        "eval_split": "validation",
+        "framework": "numpy",
+        "module": "armory.data.datasets",
+        "name": "so2sat"
+    },
+    "defense": {
+        "kwargs": {
+            "apply_fit": false,
+            "apply_predict": true,
+            "channels_first": false,
+            "clip_values": [
+                0.0,
+                1.0
+            ],
+            "mins": [
+                -0.22,
+                -0.26,
+                -0.73,
+                -0.75,
+                0,
+                0,
+                0,
+                0,
+                0,
+                0,
+                0,
+                0,
+                0,
+                0
+            ],
+            "quality": 95,
+            "ranges": [
+                0.46,
+                0.48,
+                1.52,
+                1.11,
+                0.7,
+                0.7,
+                0.7,
+                0.61,
+                0.7,
+                0.7,
+                0.7,
+                0.7,
+                0.62,
+                0.54
+            ]
+        },
+        "module": "armory.art_experimental.defences.jpeg_compression_multichannel_image",
+        "name": "JpegCompressionMultiChannelImage",
+        "type": "Preprocessor"
+    },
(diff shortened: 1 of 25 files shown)


lcadalzo:
@yusong-tan do you agree that this PR would be a good time/place to update documentation on baseline results as well?

yusong-tan:
@lcadalzo Ideally, yes, this is a good PR to update the documentation. However, I'm rerunning ucf101 and it's going to take a while (1 week maybe), so I don't know if you can wait for it.

scenario_configs/so2sat_sar_masked_pgd_defended.json
```diff
@@ -39,37 +37,37 @@
...
-                0.0,
-                0.0,
-                0.0,
-                0.0,
-                0.0,
-                0.0,
-                0.0,
-                0.0,
-                0.0
+                -0.22,
```
lcadalzo, line 40: how were the `"mins"` and `"ranges"` values determined for so2sat configs? I do recall conversation that the previous bounds were loose compared to the actual distribution of benign data

yusong-tan: @lcadalzo I found the min/max by cycling through each channel in the test data. The ranges = max - min.

lcadalzo:
In the last evaluation phase, for PGDPatch, the location was fixed at (0, 0) unless "there’s evidence that the defense is overfitting to that coordinate, in which case, arbitrarily select another coordinate in the image". It appears now that the patch location is being randomized each call to `generate()` for xView and So2Sat (since those configs don't specify `xmin` and `ymin`, while (0,0) is being used for UCF101. 

Is there a reason this change was made for xView and So2Sat and not UCF101? Obviously these are just example configs and not necessarily the parameters that need to be used for Eval3, but want to make sure I'm clear on this point. In the case of xView, randomizing the patch location could cause the patch to cover up/overwrite the object of interest (this is also possible using (0,0), but perhaps less likely)

yusong-tan:
@lcadalzo Yes, the randomization of the patches for xView and so2sat is to make evaluations across defenses more comparable as the previous defenses all assume randomly positioned patches; at the same time, searching for the worst-case patch location is unduly expensive. The changes also make the attacks more realistic because while a patch could be fixed on the ground, its location in the overhead imagery cannot be guaranteed. 

In the ucf101 scenario, allowing a patch that could randomly land in the middle of the video to occlude the main action seems unreasonable and unrealistic.

lcadalzo:
That makes sense r.e. So2Sat and UCF. Regarding xView, is there any concern that a randomly placed patch which happens to occlude the ground-truth object is perhaps also an unreasonable/unrealistic threat model to defend against?

yusong-tan:
@lcadalzo Since we are trying to make all objects disappear, I think there's enough randomness (in number and location of objects as well as location of patches) that occlusion shouldn't be a big issue on average. Moreover, I don't have any reason to believe that patch at (0,0) is less likely to occlude objects.

lcadalzo:
The changes in this PR have been added to #1123, which instead has a destination branch of `dev_0.13.4` rather than `dev`. The `dev` branch is where development for version `0.14.0` is occurring, and Mike's changes here should be tacked onto 0.13.3 for a new 0.13.4 release

PR #1120, opened 2021-06-21
Add non-preloaded dirty-label backdoor attack with bullethole trigger

ng390:
Add non-preloaded dirty-label backdoor attack with bullethole trigger - the preloaded dataset now will be deprecated.

### DIFF
--- a/scenario_configs/gtsrb_scenario_bullethole.json
+++ b/scenario_configs/gtsrb_scenario_bullethole.json
@@ -1,0 +1,75 @@
+{
+    "_description": "GTSRB poison image classification, contributed by MITRE Corporation",
+    "adhoc": {
+        "detection_kwargs": {
+            "nb_clusters": 2,
+            "nb_dims": 43,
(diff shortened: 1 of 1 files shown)


ng390:
I benchmarked performance of the attack with the ART blended trigger versus the preloaded attack (5) trials. These can be repeated by running scenario_configs/gtsrb_scenario_poison.json on armory 0.13.3 with the value of config["adhoc"]["train_epochs"] set to be 50.

Results:
The test accuracies are comparable - 86.7% to 84.5%, however the preloaded attack had a 83.9% targeted success rate vs. 72.7% for the ART blended trigger. The preloaded attack appears to be slightly stronger - I don't have an immediate explanation for this. My understanding was the trigger, blend ratio, and positioning of the ART trigger match those used in the preloaded dataset.

davidslater:
LGTM. We can try to determine the reasoning during evals.

davidslater (approved):

PR #1121, opened 2021-06-23
Dataset split utilities for bullseye polytope attack

ng390:
Given a dataset with only two splits - train and test, these utilities divide the train set into subsets following a generalized version of the logic from the Bullseye Polytope attack, which in turn borrows the logic from the Convex Polytope [attack](https://arxiv.org/pdf/1905.05897.pdf).


davidslater (approved):
LGTM. The splits work, and will be helpful if we revisit bullseye polytope

PR #1129, opened 2021-07-16
WIP: Resisc10 poison label fix: convert np array label to int

lcadalzo:
prior to this fix, `poison_y` is a list which contains a mix of ints and np arrays. The `if` block adds np arrays to this list, while the `else` block adds ints. This results in `poison_y` being of type np object array after the list of mixed types is converted to numpy array. This causes one submission to error. 

If this PR is merged, this would then require a 0.13.5 release. Before merging, I'd like to investigate why this isn't causing other resisc10 poison configs to fail, so I've marked as WIP


davidslater:
This same behavior happens on the main GTSRB poisoning example. Since those are all running currently, I would rather not do another release but make these changes on `dev` and do a workaround in this specific case.

lcadalzo:
I've created and issue and am closing this PR since the fix should be included in a branch based off the `dev` branch

PR #1132, opened 2021-07-19
abstain metric

davidslater:
Adds the ability to measure whether a model is abstaining

### DIFF
--- a/armory/utils/metrics.py
+++ b/armory/utils/metrics.py
@@ -25,7 +25,9 @@ logger = logging.getLogger(__name__)
 
 def abstains(y, y_pred):
     """
-    Returns 1 if y_pred is all 0s (an abstention)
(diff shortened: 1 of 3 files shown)


armory/utils/metrics.py
```diff
@@ -23,6 +23,17 @@
 logger = logging.getLogger(__name__)
 
 
+def abstains(y, y_pred):
+    """
+    Returns 1 if y_pred is all 0s (an abstention)
```
mwartell, line 28: The return type of this method is `List` and I cannot tell if a return of `[0]` or `[0, 0]` would satisfy the calling contract.

mwartell (commented):
Either a doc bug or an implementation bug.

mwartell (approved):
LGTM.

PR #1136, opened 2021-08-03
Update scenarios md

lcadalzo:


lcadalzo:
@yusong-tan do you have intuition as to why, for all patch sizes, the defended DAPRICOT model has a [higher](https://github.com/yusong-tan/armory/blob/update_scenarios_md/docs/scenarios.md#dapricot-object-detection-updated-july-2021) `Target Success` than the undefended one (given that a higher value for this metric implies a more successful attack)?

yusong-tan:
@lcadalzo I don't. We see this in one or two other scenarios, too, like ASR (for undefended PGD), where compression improves adversarial attack success.

lcadalzo:
got it, thanks. And are DAPRICOT results on the _test_ dataset (as opposed to _dev_)? Should this be specified in the doc?

yusong-tan:
Yes, it's run on the _test_ data. Can you add that snippet when you merge it in?

lcadalzo:
If the `Impercetible ASR` attack is targeted, is the reported `Adversarial WER` w.r.t. the original labels or target label? I'm guessing the former since the values go down as attack strength increases. Would it be useful to also show the metric w.r.t. target labels?

lcadalzo:
For Robust DPatch on xView, the reported `Adversarial mAP (Defended)` is higher for an 100x100 patch than 75x75. The values are close enough that I suppose this is possible, but just want to confirm this is correct.

yusong-tan:
@lcadalzo Results for Imperceptibile ASR is wrt the target transcriptions. WER going down as attack strength increases means the target transcription is being generated more accurately. I don't think the logger logs metric wrt the original labels, but I'm not opposed to having if it's supported.

yusong-tan:
@lcadalzo Yes re: Robust DPatch. They are statistically similar, and the difference is probably in the noise.

PR #1137, opened 2021-08-05
Dev 0.13.5

lcadalzo:
For updates needed during the evaluation phase. This should be merged at the conclusion of the evaluation phase


mwartell (approved):
There is much parallel code addition to the poisoning scenario _evaluate methods. In a non-point release we might want to factor that out.

PR #1146, opened 2021-09-21
Version bump to 13.5

mwartell:
The CI build failed because it was trying to reupload 13.4 to pypi.org


lcadalzo (changes requested):
JSON version's should be updated to 0.13.5 using [this](https://github.com/twosixlabs/armory/blob/master/tools/update_config_version.py) script

lcadalzo (approved):

PR #1162, opened 2021-10-11
Dataset Indexing - Index and class filtering from command line

davidslater:
It turns out that most of the functionality was already done:
https://github.com/twosixlabs/armory/blob/dev/armory/data/datasets.py#L273-L337

With tests:
https://github.com/twosixlabs/armory/blob/dev/tests/test_docker/test_dataset.py#L71-L142

I just added hooks for the command line, `--index` and `--classes`, and updated the docs.

Will close the previous PR as there was unnecessary code there.

### DIFF
--- a/armory/__main__.py
+++ b/armory/__main__.py
@@ -13,6 +13,7 @@ import argparse
 import json
 import logging
 import os
+import re
 import sys
 
 import coloredlogs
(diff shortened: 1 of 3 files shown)


mwartell (approved):
Working as described. I've hit all the guarded conditions and they are indeed caught.   This implements a data filter and based on relative execution times it is filtering, but I don't know how that should alter the results (aside from "not as good").

PR #1170, opened 2021-10-12
Video tracking integration

yusong-tan:
Final dev dataset will be sent separately via email.


lcadalzo:
let's hold off on merging until #1186 is addressed, since this would break the dev branch

PR #1179, opened 2021-10-15
fix dependency ordering in tf1 docker creation

### LINKED ISSUE #1175: ci build broken for tf1 targets
Today the CI builds started breaking during the Tensorflow1 docker builds. The proximate error was a pip complaining that h5py installation failed, however it was the prior numpy==1.19.2 installation that appeared to be breaking the pip metadata. Anaconda released https://anaconda.org/anaconda/numpy-base today and there are both pip and conda numpy packages in the built containers with differing versions. A [CI run yesterday](https://github.com/twosixlabs/armory/actions/runs/1338444225/attempts/1) against dev/HEAD passed cleanly, rerunning it today generated the same pip failure.

The [conda docs](https://www.anaconda.com/blog/using-pip-in-a-conda-environment) are pretty firm on the idea of "conda first, then pip" and say that interleaving installations can damage the environment. I'll be reordering the TF1 Dockerfile to follow this prescription.

mwartell:
Fixes #1175 

The CI builds were breaking apparently correlated with the release of a new conda version of numpy. This induced dependency ordering issues in the TensorFlow Docker build that percolated through a few CI tests. 

Reordering the conda and pip install actions as [Anaconda recommends (conda then pip)](https://www.anaconda.com/blog/using-pip-in-a-conda-environment) cleared the CI tests as recorded in https://github.com/twosixlabs/armory/actions/runs/1347437832

### DIFF
--- a/docker/tf1/Dockerfile
+++ b/docker/tf1/Dockerfile
@@ -20,27 +20,33 @@ RUN wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-py37_4.9.2-Linux
 
 ENV PATH=/opt/conda/bin:$PATH
 
+RUN /opt/conda/bin/conda install \
+    tensorflow-gpu==1.15.0 \
+    numpy==1.19.2 \
+    h5py==2.10.0 \
+    pandas \
+    jupyterlab \
+    boto3 \
+    dill \
+    pytest
+
+
+RUN /opt/conda/bin/conda install -c conda-forge ffmpeg==4.2.3
 
 RUN /opt/conda/bin/pip install --no-cache-dir \
     tensorflow-datasets==3.2.0 \
-    jupyterlab==3.0.9 \
-    boto3==1.17.20 \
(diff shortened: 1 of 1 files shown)


lcadalzo:
A couple comments:

(1) should we apply the same process (of conda installing what we can before pip) to the other Dockerfiles?
(2) I noticed this PR unpins some packages. I would prefer to keep packages pinned so builds are consistent over time, what do you think? Regardless, I think it makes sense to update versions where applicable

mwartell:
@lcadalzo 
1. Yes we should reorder the other Dockerfiles. I made issue #1180 to allow this fix to merge.

lcadalzo:
R.e. (2): I agree we should pin to latest wherever we can. I was referring to pandas, jupyterlab, boto3, dill, pytest as packages that we should go back to pinning (not necessarily to the previously used version)

lcadalzo (approved):
LGTM

PR #1194, opened 2021-10-22
Refactor dataset config loading

swsuggs:
Standard items are removed from dataset_config and everything else is assumed to be a kwarg for the dataset loading function.


lcadalzo (approved):
LGTM

PR #1210, opened 2021-11-15
WIP: ART Tensorboard integration

### LINKED ISSUE #1208: Integrate with ART's Tensorboard feature
ART provides support for Tensorboard inspection of loss + gradients. More specifically, users can pass a `"tensor_board"` kwarg to certain attacks* which is either a boolean or a string. If `True` is passed, it will output the TF event files to `./runs`. If a string is provided, it will output to the provided directory.

For Armory's sake, I think it'd make the most sense to output TF event files to the same directory as where results JSON's and exported samples are outputted. This would perhaps mean overriding user inputs if a string is provided.

\*per Beat, just ProjectedGradientDescent* and AdversarialPatch* thus far

lcadalzo:
Closes #1208

### DIFF
--- a/armory/scenarios/scenario.py
+++ b/armory/scenarios/scenario.py
@@ -153,14 +153,14 @@ class Scenario:
         if targeted and use_label:
             raise ValueError("Targeted attacks cannot have 'use_label'")
 
-        if "tensor_board" in attack_config.get("kwargs", {}):
-            tensor_board_kwarg = attack_config.get("kwargs").get("tensor_board")
-            if isinstance(tensor_board_kwarg, str):
+        if "summary_writer" in attack_config.get("kwargs", {}):
+            summary_writer_kwarg = attack_config.get("kwargs").get("summary_writer")
+            if isinstance(summary_writer_kwarg, str):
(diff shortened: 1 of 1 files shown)


lcadalzo:
Marking WIP since I'm not sure if subsequent examples are overwriting previous data. E.g if I set `--num-eval-batches 2`, I want to view statistics for both batches

lcadalzo:
Subsequent examples are in fact being overridden. The ART team is working on incorporating fixes into ART 1.9. To be more specific, there are two separate but related issues: (1) within a single batch, data is overridden for subsequent _examples_ and (2) across multiple batches, data is overridden for subsequent _batches_

lcadalzo:
Another issue is that `loss` isn't being tallied for certain estimators, only `gradients`. The ART team will also work on addressing this

lcadalzo:
Adding a blocked label, since Tensorboard integration is first contingent on ART fixes for the previously described issues, and then second Armory upgrading to ART 1.9, where these fixes will live.

lcadalzo:
Closing due to #1240

PR #1211, opened 2021-11-16
adding databuilder for the test data for carla single and multi-modal…

yusong-tan:
…ity object detection scenarios

Test dataset will be sent separately via email.

### DIFF
--- a/armory/data/adversarial/carla_obj_det_test.py
+++ b/armory/data/adversarial/carla_obj_det_test.py
@@ -26,7 +26,7 @@ _CITATION = """
 """
 
 # fmt: off
-_URLS = "carla_obj_det_test.tar.gz"
+_URLS = "https://armory-public-data.s3.us-east-2.amazonaws.com/carla/carla_od_test.tar.gz"
 # fmt: on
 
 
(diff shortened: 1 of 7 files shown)


lcadalzo:
Are there supposed to `"small"`, `"large"`, and `"medium"` splits or is this accidentally copied code from DAPRICOT? If yes, what do these indicate and why do they exist for `"test"` but not `"dev"`set?

yusong-tan:
@lcadalzo Yes, those splits are supposed to be there, just as a way to explicitly understand better if there are differences in defenses under different patch sizes. Small is <= 0.5% patch ratio, medium is between 0.5% and 0.75%, and large is between 0.75% and 1% (one or two patch may be slightly larger than 1%). These splits were added after 'dev' dataset was developed. A good defense should generalize to different patch sizes, so 'dev' lacking such explicit splits shouldn't be a hindrance to research.

lcadalzo:
Got it. And to clarify dataset size: I should be seeing 10 examples per split?

yusong-tan:
@lcadalzo Yes.

swsuggs (approved):
looks good

PR #1212, opened 2021-11-16
update carla object detection patch attack to increase its efficacy

yusong-tan:


lcadalzo:
There's a comment in the code on dev branch which this PR would remove:
```
 #must be square for now, until it's determined why non-square patches fail to completely overlay green screen
```

Has it been determined this (non-square patches failing to completely overlay green screen) is no longer an issue?

yusong-tan:
@lcadalzo It's still an issue for DAPRICOT dataset due to a constraint on the rectangle aspect ratio, but this issue doesn't apply to the CARLA dataset.

lcadalzo:
@yusong-tan what effect on results did you observe the changes in this PR to have?

yusong-tan:
@lcadalzo I don't have metrics such as mAP. Prior to the change, it was difficult to generate any hallucinated objects. After the change, one can usually produce one or more hallucinated objects per patch, under the same attack parameters. Disappearances/misclassifications are minimally impacted by the change, primarily because they are generally more difficult to produce in the first place.

lcadalzo (approved):
LGTM

PR #1213, opened 2021-11-17
fix carla video attack

yusong-tan:
Fixed a bug in CARLA video patch attack that resulted in either a slightly misalignment between patch and green screen or a patch that's only partially perturbed.


lcadalzo (approved):
LGTM

PR #1223, opened 2021-12-07
refactor video tracking attack

### LINKED ISSUE #1220: patch_height, patch_width now required args for ART's AdversarialTexturePyTorch
As of 12/1/2021, ART updated AdversarialTexturePyTorch's `__init__()` to take in positional (i.e. required) args for `patch_height` and `patch_width`. Armory currently isn't providing any values for these, since they're updated each `generate()` call depending on the green screen size. This is currently causing an error if you run Armory with the up-to-date state of ART's `development_object_tracking_attack` branch.

lcadalzo:
Closes #1220. Instantiates new attack object each `generate()` call with the appropriate `patch_height`, `patch_width`, `x_min`, and `y_min`.

Per ART team, the current pattern of overwriting attributes might be challenging to maintain if future versions require changes or add extensions

### DIFF
--- a/armory/art_experimental/attacks/carla_adversarial_texture.py
+++ b/armory/art_experimental/attacks/carla_adversarial_texture.py
@@ -1,6 +1,5 @@
 import logging
 import numpy as np
-import torch
 
 from art.attacks.evasion import AdversarialTexturePyTorch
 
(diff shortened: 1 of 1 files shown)


swsuggs (approved):
Looks all right

PR #1226, opened 2021-12-13
draft new poison metrics and associated interpretive model

yusong-tan:
Model weights will be emailed separately.


yusong-tan:
In `get_data_level_stats()`, `data_list` argument is expected to be a list of tuples, but I think it should also work with the Armory data generator.

PR #1231, opened 2021-12-21
R0.14.2

mwartell:
There was a build error in 0.14.1 which pushed valid code to PyPI but did not complete. Because PyPI uploads are immutable, later runs of the build fail upon upload.

This version is identical to 30b98d87 except `0.14.1` has been changed to `0.14.2` throughout.


lcadalzo (approved):
LGTM

PR #1232, opened 2021-12-21
reconciliation of dev and master

mwartell:
I'm trying to find out why master into dev didn't work on the commandline.


mwartell:
this is not the way to do it

PR #1233, opened 2021-12-21
make dev equivalent to master

mwartell:
Following the 0.14.2 release, we want dev to mirror master. This PR was constructed with

    git checkout dev
    git merge --strategy-option=theirs master

and nothing else. Note that the commits in this PR date back to 9 months ago. I believe this is the result of not reconciling the branches since then, but I am going to bit-by-bit compare this mwartell:dev branch and master before merging.


mwartell:
I tried advancing dev to master with
    git checkout dev
    git reset --hard 45aaf81
    git push --force mwartell dev
And while this was acceptable to git, github does not like it and this PR is broken. I'm looking for a clean alternative.

PR #1237, opened 2021-12-23
removing duplicate metrics code

### LINKED ISSUE #1234: Remove duplicate code in metrics.py
See [these](https://github.com/twosixlabs/armory/blob/master/armory/utils/metrics.py#L1614-L1686) lines

lcadalzo:
Closes #1234


mwartell (approved):
I confirm that this does clear an accidental code duplication. It is pretty apparent that this could have resulted from an erroneous merge in the past.

PR #1238, opened 2021-12-27
Dev 0.14.1 poisoning spolavar

spolavar:
The `new_poisoning_metrics.py` derives subclass separation statistics using silhouette analysis and k-means for poisoning scenarios. The demo exemplifies loading from a pretrained model with pre-computed weights together with two lists - unfiltered and filtered list of training images. The output provide statistics for each sample in the filtered list, consisting of majority cluster label, minority cluster label, total number of clusters, silhouette_score, and the average cluster separation threshold. The typical samples have a score between [0, threshold], whereas the outliers can come either from [-1,0] or [threshold, 1] range.


reed-twosixlabs:
Hey @spolavar, this PR appears to be working off of a branch of 0.13.3 rather than 0.14.0 (where dev_0.14.1_poisoning branched). Can you check to see if your implementation works with 0.14.0?

reed-twosixlabs:
Closing with permission from @spolavar as dev_0.14.1_poisoning was updated outside of this PR.

Issue #1250, opened 2022-01-12
Confusing Metrics in Carla Object Detection in v0.14.3


mzweilin:
The v0.14.3 produces results that is not consistent with that of v0.14.2. What does an AP value that is larger than 1 mean?

### How to reproduce
Run this with Armory v0.14.2 or v0.14.3.
```bash
jq ".dataset.name = \"carla_obj_det_test\" | .dataset.eval_split = \"small\" | .scenario.export_samples = 10" scenario_configs/carla_obj_det_dpatch_undefended.json | armory run --no-docker --use-gpu -
```

v0.14.2 produces very readable results.
```json
"results": {
        "adversarial_carla_od_AP_per_class": {
            "1": 0.34,
            "2": 0.14,
            "3": 0.16
        },
        "adversarial_carla_od_AP_per_class_wrt_benign_preds": {
            "1": 0.52,
            "2": 0.69,
            "3": 0.39
        },
        "adversarial_mean_carla_od_AP_per_class": 0.21333333333333335,
        "adversarial_mean_carla_od_AP_per_class_wrt_benign_preds": 0.5333333333333333,
        "adversarial_mean_carla_od_disappearance_rate": 0.39452380952380955,
        "adversarial_mean_carla_od_disappearance_rate_wrt_benign_preds": 0.15716117216117215,
        "adversarial_mean_carla_od_hallucinations_per_image": 5.5,
        "adversarial_mean_carla_od_hallucinations_per_image_wrt_benign_preds": 4.7,
        "adversarial_mean_carla_od_misclassification_rate": 0.08071428571428571,
        "adversarial_mean_carla_od_misclassification_rate_wrt_benign_preds": 0.10463369963369962,
        "adversarial_mean_carla_od_true_positive_rate": 0.5247619047619048,
        "adversarial_mean_carla_od_true_positive_rate_wrt_benign_preds": 0.7382051282051281,
        "benign_carla_od_AP_per_class": {
            "1": 0.6,
            "2": 0.17,
            "3": 0.53
        },
        "benign_mean_carla_od_AP_per_class": 0.43333333333333335,
        "benign_mean_carla_od_disappearance_rate": 0.3724603174603175,
        "benign_mean_carla_od_hallucinations_per_image": 0.8,
        "benign_mean_carla_od_misclassification_rate": 0.06071428571428572,
        "benign_mean_carla_od_true_positive_rate": 0.5668253968253969,
        "perturbation_mean_l0": 0.005976319444444444
[...]

PR #1253, opened 2022-01-24
compute carla_od_AP_per_class after all predictions have been obtained

swsuggs:
Fixes [1250](https://github.com/twosixlabs/armory/issues/1250)


lcadalzo (approved):
Confirmed that this fixes an accidental deletion of code introduced in the 0.14.2 release

PR #1255, opened 2022-01-26
0.14.4

lcadalzo:


mwartell (approved):
LGTM: I have hand reviewed each file here to figure why a merge conflict

there is discussion with @mwartell and @lcadalzo about how we can
avoid these "should have been fast-forward" merge. We've not 
puzzled it out yet.

mwartell:
This has been closed in favor of the cleaner #1256 

It would be nice if we understood why this is better than that. :/

PR #1256, opened 2022-01-26
B14.4 candidate

mwartell:
This merges clean into master while develop does not for reasons we don't yet understand.

It was created with
```
git checkout develop
git checkout -b b14.4-candidate
git  merge -s ours master
```
Because it uses the `ours` strategy, there can be no conflict because they are resolved in favor of ours commits. We have hand checked the contents of this PR against #1255 and found this more complete.

After this is merged to master, develop needs to be caught up, presumably with
```
git checkout develop
git merge master --no-commit
```
where `--no-commit` allows us to check its correctness without breaking `origin`


mwartell:
As noted in  our discussion, merging this to master will require that we also bring `develop` to the same commit level.

lcadalzo (approved):
LGTM

PR #1257, opened 2022-01-26
WIP: don't merge Bogus branch for git exploration

mwartell:
This is just here for us to look at and then discard. We're trying to find why every version bump causes merge conflicts.


mwartell:
Experiment concluded. Recommendation for next time:
```
git checkout develop
git fetch master     # just to be sure we're current
git merge origin/master
gh pr create
```
There is some supporting discussion in https://stackoverflow.com/a/20103414/282912

PR #1260, opened 2022-01-28
pin to colour release v0.3.16

davidslater:


mwartell (approved):
LGTM, however, I'm concerned that this distributes dependencies out to a more obscure location where we'll only see errors like the one that forced this pinning at runtime if and only if present in the config.

There may well be cause for doing this that I am not aware of.

davidslater:
I see this as a quick bugfix for the current branch, not necessarily a long term solution.

Longer term, I think that we should:
A) add pre-release runtime testing of standard configs
B) figure out how to get these to work in our docker containers (even if that's literally git cloning and pip install -e).

PR #1262, opened 2022-01-31
GTSRB Bean Regularized model

yusong-tan:
Adding a BEAN regularized model for GTSRB scenarios and making some minor changes to catch special cases.


reed-twosixlabs:
Partially integrated, closing based on our discussion.

PR #1266, opened 2022-02-06
Dev v0.14.2 poisoning spolavar

spolavar:
@reed-twosixlabs @swsuggs made changes to provide binary array of majority /minority via s_majority from get_data_level_stats function. The binary split is based on Silhouette analysis instead of K-mean clusters. This will give us three different ways of scoring in total (clustering, Silhouette analysis, and typicality)


reed-twosixlabs:
Partially integrated, closing based on our discussion.

PR #1269, opened 2022-02-07
Fix docker colons 1267

### LINKED ISSUE #1267: protect docker volume specification with colons
If the path to the config file passed to armory has a colon in it, the docker mount of that file fails.  For example
```
$ armory run has_colon:in_it.json 
docker.errors.APIError: 500 Server Error for http+docker://localhost/v1.40/containers/create: Internal Server Error ("invalid volume specification: '/home/ubuntu/has_colon:in_it:/workspace:rw'")
```

Use the extended docker volume spec which doesn't cons names together with `:` to fix.

mwartell:
Fixes #1267 

Uses the `mount=` form of docker-py container creation instead of `volumes=` because the later cannot accept literal colons `:` in directory names.  Since armory mounts the expanded cwd, running armory in `/home/msw/foo:bar/` will fault.

The `mount=` form works properly if there is a colon. This is similar to the `docker --mount` and `--volumes` arguments.

### DIFF
--- a/armory/docker/management.py
+++ b/armory/docker/management.py
@@ -29,7 +29,6 @@ class ArmoryInstance(object):
     ):
         self.docker_client = docker.from_env(version="auto")
 
-
         host_paths = paths.HostPaths()
         docker_paths = paths.DockerPaths()
 
@@ -47,10 +46,6 @@ class ArmoryInstance(object):
             "mounts": mounts,
             "shm_size": "16G",
         }
-
-        print(f"{container_args}")
-        input("Press Enter to continue...")
-
 
         if ports is not None:
             container_args["ports"] = ports
--- a/armory/docker/management.py
+++ b/armory/docker/management.py
@@ -32,12 +32,15 @@ class ArmoryInstance(object):
         host_paths = paths.HostPaths()
         docker_paths = paths.DockerPaths()
 
-        mounts = [docker.types.Mount(
-            source=getattr(host_paths, dir),
-            target=getattr(docker_paths, dir),
-            type="bind",
-            read_only=False,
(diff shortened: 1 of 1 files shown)


shenshaw26 (commented):
Overall this looks to be in line.  However the tests are failing... @mwartell, is that expected?

mwartell:
No. Failed CI tests are a stopper for merge. 

If TF2 tests failed in two CI runs so I'm tracking that down now.

davidslater:
When moving to mount, the target directories will not exist in the container if they do not exist in the host (volumes will create a directory if they do not exist on the host). Should we explicitly test to see if these directories exist on host before creating the container? (This could possibly prevent weird user-facing errors if they haven't configured armory correctly.)

davidslater:
Apparently we already create the directories if they don't exist: https://github.com/twosixlabs/armory/blob/master/armory/paths.py#L87-L92

This worked fine for me.

davidslater (approved):
LGTM

mwartell:
The CI failure for this PR was fixed by 8dc203ca

mwartell:
The CI failure was accurate and my merge of this has infected branch `develop`.  However, [the test failure](https://github.com/twosixlabs/armory/runs/5240077590?check_suite_focus=true) is pretty deep inside the armory call-chain and I cannot make sense of it so ask @davidslater for guidance

shenshaw26:
FWIW, this looks like the same error we were getting when the version mismatch business what happening:  `ImportError: SystemError: <built-in method __contains__ of dict object at 0x7fc40c64fd20> returned a result with an error set`

I can take a look and see if I can figure it out.

PR #1274, opened 2022-02-11
remove dependency on pandas

davidslater:
This doesn't remove the library from our dockerfiles, it just removes the toolkit's dependence on pandas (which was only used for simple csv reading).


davidslater:
pulled in via #1277

PR #1275, opened 2022-02-11
remove apache_beam as explicit dependency

davidslater:


davidslater:
pulled in via https://github.com/twosixlabs/armory/pull/1277

PR #1277, opened 2022-02-12
Simple docker update

davidslater:
Removed apache_beam and pandas dependencies in the main codebase, as well as moved pymongo from requirements to host-requirements, since it is only used by scenarios.

The main update, though, was to regenerate the pytorch Dockerfile from basically scratch. I haven't tried remaking others, but I think it's a helpful step toward dealing with some of our larger dependency issues. I'd like some feedback to see if this process should be extended to tf1/tf2/pytorch-deepspeech.


davidslater:
As comments, I've added the outputs from my conda and pip installs.

One anti-pattern we were using before was doing anything conda-related after we've done a pip install. Now all the conda actions take place before pip.

armory/data/adversarial/pandas_proxy.py
```diff
@@ -0,0 +1,27 @@
...
+
+import numpy as np
+
+
+def read_csv_to_numpy_float32(path, header=None) -> "np.array":
+    if header is not None:
+        raise NotImplementedError("non-None header not supported")
+
+    rows = []
+    with open(path, newline="") as f:
```
shenshaw26, line 23: We may want to add support for csv files to have headers...

davidslater: There are multiple places this could fail, not just headers. I did not want to implement the full CSV reading capability of pandas - just what I needed to replace these, since we only use them for building 4 specific datasets and all in the same way.

PR #1279, opened 2022-02-14
Carla: export all relevant bboxes

swsuggs:
Only boxes with score greater than 0.9 were being drawn on exported samples, but the carla metrics count all predictions with score greater than 0.5.  This matches the export function to the metrics.


davidslater:
Can you update with the "Fixes #XXXX" syntax for all issues this corresponds to?

davidslater:
Oh, nevermind, I didn't realize it was so minor a mod.

davidslater (approved):
LGTM

PR #1281, opened 2022-02-15
update baseline to be up to date with ART

davidslater:
This updates this baseline model file to:
a) properly save the model weights to armory data directory instead of tmp (due to update in ART)
b) remove unnecessary subclass due to `compute_loss` being implemented in ART.


swsuggs (approved):
looks good

PR #1294, opened 2022-02-18
WIP: logging, testing, CI improvements

mwartell:
We changed some fundamental mechanisms so need the dev team evaluate how these work for them.
  
## dockerfile splitting

The new `docker/build.sh` and Dockerfiles create a base system image that is then augmented with platform dependent layers. For developers:

    bash docker/build.sh -f base      # yields twosixamory/armory-base:dev
    bash docker/build.sh -f pytorch   # (or tf2 or deepspeech or all)

where the base needs to be built infrequently. ??? will be pushed to dockerhub once a quarter or so ???

When run with no other arguments, build.sh creates the images locally with the `:dev` tag yielding `twosixarmory/base:dev` and for example `twosixarmory/pytorch:dev`.

## pytest refactoring

Reorganization of the various pytests has made them cleaner. A side effect of this work is that `cd armory ; pytest`
should just work. Developer tests can be run with

    pytest -m "not slow"

## ci changes

The github actions have been updated to use the revised tests. We currently ??? do from-scratch builds of the images
but may well go to the relatively static base described above

## more logging control

The `--debug` argument to armory has been deprecated in favor of the `--log-level` argument. The `--log` takes a value of the form module:level such as `art:debug` and multiple `--log spec` options are accepted. For example:

    armory run --log art:debug --log armory:trace config.json

You can find more detail on this feature in [logging.md](docs/logging.md).

## automatic logfile creation

Armory now creates `armory-log.txt` and `colored-log.txt` in the per-run output directory beside the output json file.
Because that output directory isn't created until the armory.scenario.main module is running, the logs before then cannot
be written until the armory "engine" is running. This will be fixed in a later revision.


mwartell:
And yeah there are > 140 files changed but a whole bunch are just `s/0.14.4/0.14.5/` and `s/logging/log/`

.gitignore
```diff
@@ -107,26 +108,6 @@ venv.bak/
 # mypy
 .mypy_cache/
 
-# idea
+# editors
```
mwartell, line 111: A example of "git merge guessed wrong". Too much changed in the space of a few lines so `.vscode/` got moved away from its `# editors` comment. It makes no functional difference because .gitignore has a `\n` separated grammar, but it is interesting.

Issue #1273, opened 2022-02-11
Refactor export sample feature


lcadalzo:
Refactor export_sample feature:

- the inclusion of predicted/ground-truth boxes should be extended in a clean way to the generic object detection scenario (i.e. not just CARLA, xView too)
-  we will likely want to move the exporting code outside of `run_attack()` into its own scenario method
- the box plotting logic inside `export.py` should be cleaned up. Currently there are various kwargs that get passed around and used in unintuitive ways. See [this](https://github.com/twosixlabs/armory/pull/1261#discussion_r800842651) comment. This may entail having export functions (as well as `Context` classes) separated by scenario rather than data type (e.g. image classification and object detection use different functions)

lcadalzo:
An additional reason for this refactor: currently, since exporting occurs inside the `run_attack()` method, if a user sets `export_samples` correctly in the config but then uses the `--skip-attack` flag, no samples will be exported nor will there be any warning message indicating this

davidslater:
Fixed by #1297

Issue #1270, opened 2022-02-09
Enabling exporting of all samples dynamically


lcadalzo:
Currently to export samples you must specify an int. It would be convenient to instead just specify `true` if you'd like for all samples to be exported, this way you don't need to always modify the config based on how many batches you plan to run.

davidslater:
Fixed by #1297

PR #1297, opened 2022-02-21
refactor export samples feature

lcadalzo:
See #1273. 

TODO:

- [x] get export samples working for DAPRICOT
- [x] add a method which returns (but doesn't write) viewable samples (e.g. PIL object)
- [x] address #1334 
- [x] address #1270
- [x] make it possible to export only predicted boxes or only ground-truth boxes
- [x] add tests
- [x] add documentation
- [x] fix bug that occurs when passing y=None
- [x] refactor so that each exporter uses a single get_sample() method?
- [x] address CI errors

Most likely out-of-scope:
- [ ] allow users to import Exporter's into their model code to export inputs after preprocessing steps?


armory/utils/export.py
```diff
@@ -1,74 +1,51 @@
...
-            self.export_fn = self._export_audio
-        elif isinstance(self.context, So2SatContext):
-            self.export_fn = self._export_so2sat
-        else:
-            raise TypeError(
-                f"Expected VideoContext, ImageContext, AudioContext, or So2SatContext, got {type(self.context)}"
-            )
         self._make_output_dir()
 
     def export(
```
davidslater, line 25: Could we modify the export function so that it could also just return the annotated image/video/audio? This would be very nice for jupyter notebooks / interactive use. This could either be its own function call, like "annotate", that shares the same underlying code, or a flag that gets passed in like `return_sample=True` and skips the last step of the saving process.

armory/scenarios/video_ucf101_scenario.py
```diff
@@ -19,3 +20,8 @@ def load_dataset(self):
                 "    If training, set config['model']['fit_kwargs']['fit_batch_size']"
             )
         super().load_dataset()
+
+    def _load_sample_exporter(self):
+        return VideoClassificationExporter(
+            self.scenario_output_dir, self.num_export_samples, frame_rate=25
```
davidslater, line 26: `frame_rate` should be accessible from the VideoContext

armory/scenarios/carla_video_tracking.py
```diff
@@ -75,3 +76,8 @@ def run_attack(self):
             )
 
         self.x_adv, self.y_target, self.y_pred_adv = x_adv, y_target, y_pred_adv
+
+    def _load_sample_exporter(self):
+        return VideoTrackingExporter(
+            self.scenario_output_dir, self.num_export_samples, frame_rate=10
```
davidslater, line 82: `frame_rate` should be accessible from the VideoContext

armory/utils/export.py
```diff
@@ -1,74 +1,51 @@
...
 
 logger = logging.getLogger(__name__)
 
 
 class SampleExporter:
-    def __init__(self, base_output_dir, context, num_samples):
+    def __init__(self, base_output_dir, num_samples):
         self.base_output_dir = base_output_dir
-        self.context = context
         self.num_samples = num_samples
```
davidslater, line 18: Does it make more sense for `num_samples` to be part of the scenario base class instead of the exporter? That would simplify its functionality and make it easier to test.

armory/scenarios/scenario.py
```diff
@@ -227,17 +227,22 @@ def load_metrics(self):
...
+
+    def load_sample_exporter(self):
+        self.num_export_samples = self.config["scenario"].get("export_samples")
+        if self.num_export_samples is not None and self.num_export_samples > 0:
+            sample_exporter = self._load_sample_exporter()
         else:
             sample_exporter = None
-
-        self.metrics_logger = metrics_logger
         self.sample_exporter = sample_exporter
```
davidslater, line 238: Since `_load_sample_exporter()` is required, what if we always instantiate the sample_exporter, and add logic to `evaluate_current()` to determine when to call `self.export_samples()`? This would ensure that in interactive use the user can always call `export_samples()`.

armory/scenarios/scenario.py
```diff
@@ -313,16 +319,24 @@ def run_attack(self):
             )
         self.metrics_logger.update_perturbation(x, x_adv)
 
-        if self.sample_exporter is not None:
-            self.sample_exporter.export(x, x_adv, y, y_pred_adv)
-
         self.x_adv, self.y_target, self.y_pred_adv = x_adv, y_target, y_pred_adv
 
+    def export_samples(self):
+        self.sample_exporter.export(
```
davidslater, line 325: If x and x_adv is None, how does this behave? Should we raise a user-facing Error to the effect of "Must call `evaluate_current` before exporting samples?

armory/utils/export.py
```diff
@@ -1,74 +1,51 @@
...
 from PIL import Image, ImageDraw
 from scipy.io import wavfile
 
-from armory.data.datasets import ImageContext, VideoContext, AudioContext, So2SatContext
-
 
 logger = logging.getLogger(__name__)
 
 
 class SampleExporter:
```
davidslater, line 15: Just a thought: do we want the ability to export just the annotated ground truth?

armory/utils/export.py
lcadalzo: I do think this would make the export code simpler. Somewhere we have to keep track and ensure we don't export more than the specified number of samples in the config value. We loop through the different elements of the batch inside the `Exporter` code, rather than the scenario code, so it's a little easier to keep track of this in the `Exporter` code with an `Exporter` instance attribute. 

But it can be done from the scenario code too. I added commit dcc11ac to address this. What I don't love about this change is having `export_samples()` take a positional arg of `num_samples`. Thoughts? Could also have this kwarg default to `None`, in which case all samples in the batch are exported. This way a user could simply call `self.export_samples()` in an interactive session

What I'm trying to avoid is, for ex., you set `batch_size` to 64 but `export_samples` to 1, you want 1 sample exported not 64. This would be a little simpler to handle if it was `export_batches` instead

davidslater: I like the idea of using `export_batches`. Since we tend to think in terms of batches (e.g., `--num-eval-batches`) and since we usually set `batch_size=1` anyway, I think that would be a nice simplification.

lcadalzo: made this switch in 7cfbd82

armory/scenarios/scenario.py
lcadalzo: In 8f376a0, I made a change s.t. an Exporter is always loaded but the scenario attribute `self.num_export_batches` defaults to 0 (rather than `None`). Then in `evaluate_current()`, a simple check of ` if self.num_export_batches > self.sample_exporter.saved_batches` is done

armory/scenarios/scenario.py
lcadalzo: it will properly handle if `x_adv` is `None` (e.g. when `--skip-attack` is set). `x` will never be `None` when running a scenario, unless I suppose a user interactively ran the following:

```
from armory import scenarios
s = scenarios.get("/armory/tmp/2022-03-15T202227.880004/interactive-config.json").load()
s.export_samples()
```

Is this the scenario where you envisioned `x` could be `None`? Here's the error that arises in this case:

```
AttributeError: 'ImageClassificationTask' object has no attribute 'x'
```

That's somewhat cryptic of an error message, although I think this is an unlikely edge case to occur in the first place. That said, I added a check in 6339176 which instead shows:

```
AttributeError: ImageClassificationTask has no attribute 'x'. Be sure to call evaluate_current() before attempting to export samples. 

```

davidslater: No, I was considering a scenario where we have an adversarial dataset with `x_adv` but no `x`. Both being None would be weird.

As an aside, I like the check.

armory/utils/export.py
lcadalzo: this seems straightforward for image scenarios: return the PIL object. I'm less clear on what this would look like for video and audio though. It doesn't seem particularly useful to return the np array since this can already be accessed via `self.x` or `self.x_adv`.

davidslater: For so2sat, I would return a list of PIL objects.
For video, it looks like you're separately saving frames. I would return either that set of PIL images, or the annotated numpy "pixels" arrays.
For audio,  it needs to be mapped back to the discrete np.int16 input plus sample rate. Maybe just return that? There's not a ton to do in the audio space, at least at the moment.

What I don't know yet (aside from PIL) is what is the easiest way to show a video or play audio in jupyter.

lcadalzo: I added a handful of commits in an attempt to address this point. They add a `get_sample()` method to each Exporter which can be used to return the PIL image, list of PIL frames, etc. depending on the scenario. This `get_sample()` method is then used internally by each Exporter's `_export()` method (the lone exception being Audio)

The `AudioExporter`'s `get_sample()` is unique in that it requires a `dataset_context` arg since `context.quantization` is needed to scale back the sample. However, I think I noticed in `datasets.py` [here](https://github.com/twosixlabs/armory/blob/develop/armory/data/datasets.py#L735) that  2 **15 is always used similar to how 255 is always used for images, so perhaps this is unnecessary?

tests/unit/test_export.py
```diff
@@ -0,0 +1,96 @@
...
+        "labels": np.array([1.0]),
+        "boxes": np.array([[0.0, 0.0, 1.0, 1.0]]).astype(np.float32),
+    }
+    y_pred = {
+        "scores": np.array([1.0]),
+        "labels": np.array([1.0]),
+        "boxes": np.array([[0.0, 0.0, 1.0, 1.0]]).astype(np.float32),
+    }
+
+    pil_img_with_both_boxes = exporter.get_sample_with_boxes(random_x, y_i=y, y_i_pred=y_pred)
```
shenshaw26, line 40: can we change the call to something like `.get_sample(..., with_boxes=True)` so that the method call stays the same?

tests/unit/test_export.py
```diff
@@ -0,0 +1,96 @@
...
+    assert isinstance(pil_img_with_boxes_list, list)
+    assert len(pil_img_list) == num_frames
+    for img in pil_img_list:
+        assert isinstance(img, PIL.Image.Image)
+
+
+def test_so2sat_export():
+    random_x = np.random.rand(32, 32, 14)
+    exporter = So2SatExporter(base_output_dir="/armory/tmp")
+    sample_vh = exporter.get_vh_sample(random_x)
```
shenshaw26, line 86: what does `vh` mean?  Can we just keep `get_sample` as method name and introduce this as an arg where value can be `vh, vv, eo`?

tests/unit/test_export.py
lcadalzo: working on making this change. I think the original motivation for multiple different methods was that case 1 (no boxes) only needs a single arg, whereas case 2 (with boxes) needs many more kwargs (`y`, `y_pred`, `with_boxes`, `classes_to_skip`, `score_threshold`). The `with_boxes` kwarg is almost superfluous since `y` and `y_pred` will be `None` unless you want to plot boxes. We used to (i.e. on develop we do) just effectively infer `with_boxes` from whether a `y` or `y_pred` is passed, but it feels not very transparent to the user if we don't have a kwarg explicitly stating whether boxes should be plotted

shenshaw26: yeah sounds good.

lcadalzo:
removing WIP since the 2 CI errors are occurring on `develop` as well

armory/scenarios/carla_object_detection.py
```diff
@@ -29,46 +30,46 @@ def load_dataset(self):
...
-    def run_benign(self):
-        x, y = self.x, self.y
-        y_object, y_patch_metadata = y
+    def next(self):
+        x, y = next(self.test_dataset)
+        i = self.i + 1
+        self.i, self.x, self.y = i, x, y
+        self.y_object = [y[0]]
+        self.y_patch_metadata = [y[1]]
+        self.y_pred, self.y_target, self.x_adv, self.y_pred_adv = None, None, None, None
```
davidslater, line 39: This could be simplified to:
```
    def next(self):
        super().next()
        self.y_object = [self.y[0]]
        self.y_patch_metadata = [self.y[1]]
```

armory/scenarios/carla_video_tracking.py
```diff
@@ -24,51 +25,67 @@ def load_dataset(self):
...
         super().load_dataset(eval_split_default="dev")
 
+    def next(self):
+        x, y = next(self.test_dataset)
+        i = self.i + 1
+        self.i, self.x, self.y = i, x, y
+        self.y_object, self.y_patch_metadata = y
+        self.y_pred, self.y_target, self.x_adv, self.y_pred_adv = None, None, None, None
+
     def run_benign(self):
```
shenshaw26, line 35: Same comment here....Does /should this be in `Scenario` base class?

armory/scenarios/carla_video_tracking.py
```diff
@@ -24,51 +25,67 @@ def load_dataset(self):
             raise ValueError("batch_size must be 1 for evaluation.")
         super().load_dataset(eval_split_default="dev")
 
+    def next(self):
+        x, y = next(self.test_dataset)
+        i = self.i + 1
+        self.i, self.x, self.y = i, x, y
+        self.y_object, self.y_patch_metadata = y
+        self.y_pred, self.y_target, self.x_adv, self.y_pred_adv = None, None, None, None
```
davidslater, line 33: Similarly to above:
This could be simplified to:
```
    def next(self):
        super().next()
        self.y_object, self.y_patch_metadata = self.y
```

armory/scenarios/dapricot_scenario.py
```diff
@@ -101,19 +102,19 @@ def run_attack(self):
                 raise ValueError("D-APRICOT batch size must be set to 1")
             # (nb=1, num_cameras, h, w, c) --> (num_cameras, h, w, c)
             x = x[0]
-            y_object, y_patch_metadata = y
+            self.y_object, self.y_patch_metadata = y
```
davidslater, line 105: Move this to `next()` function so that it is consistent with CARLA scenarios.

armory/scenarios/scenario.py
```diff
@@ -220,16 +220,25 @@ def load_metrics(self):
...
+        if num_export_batches is True:
+            num_export_batches = len(self.test_dataset)
+        self.num_export_batches = num_export_batches
+        self.sample_exporter = self._load_sample_exporter()
+
+    @abc.abstractmethod
+    def _load_sample_exporter(self):
+        raise NotImplementedError(
+            f"_load_sample_exporter() method should be defined for scenario {self.__class__}"
+        )
```
davidslater, line 241: I'm not sure that we want to enforce this method for Scenario, as we haven't enforced other methods. Instead, I would probably just have a runtime error like this:
```
    def _load_sample_exporter(self):
        raise NotImplementedError(
            f"_load_sample_exporter() is not implemented for scenario {self.__class__}"
        )
```

armory/utils/export.py
```diff
@@ -1,71 +1,54 @@
...
-            if plot_bboxes:
-                # For OD and VT scenarios
-                self.export_fn(x, x_adv, y, y_pred_adv, y_pred_clean, classes_to_skip)
-            else:
-                self.export_fn(x, x_adv)
+    @abc.abstractmethod
+    def get_sample(self):
+        raise NotImplementedError(
+            f"get_sample() method should be defined for scenario {self.__class__}"
+        )
```
davidslater, line 51: `self.__class__` here would refer to `SampleExporter`, not to a scenario.

armory/utils/export.py
```diff
@@ -92,369 +75,536 @@ def _make_output_dir(self):
...
-                    os.path.join(self.output_dir, f"{self.saved_samples}_depth.png")
-                )
-            else:
-                raise ValueError(f"Expected 1 or 3 channels, found {x_i.shape[-1]}")
+    @staticmethod
+    def get_sample(x_i):
+        if x_i.min() < 0.0 or x_i.max() > 1.0:
+            log.warning("Image out of expected range. Clipping to [0, 1].")
+
+        # Export benign image x_i
```
davidslater, line 108: Add some documentation here w.r.t. what shapes are expected and why. (E.g., grayscale, RGB, RGB-Depth) Also, that this expects a floating point [0.0, 1.0] input.

davidslater (changes requested):
I need to do a deeper review (i.e., run the code), but I gave a few comments below. The `export.py` could also use some more doctrings.

davidslater:
Code is working fine for me.

armory/utils/export.py
```diff
@@ -92,369 +77,579 @@ def _make_output_dir(self):
...
 
-            for n_frame, (x_frame, x_adv_frame) in enumerate(zip(x_i, x_adv_i)):
+        if x_i.min() < -1.0 or x_i.max() > 1.0:
+            log.warning("input out of expected range. Clipping to [-1, 1]")
+
+        wavfile.write(
+            os.path.join(self.output_dir, f"{self.saved_samples}_{type}.wav"),
+            rate=self.sample_rate,
+            data=np.clip(x_i, -1.0, 1.0),
+        )
```
davidslater, line 553: Audio inputs are limited to the [-1.0, 1.0] range, but this is not true after adversarial additions. Instead of clipping in this case, I would normalize by the max value, `np.abs(x_i).max()` and have the warning say that.

We may want to consider quantizing the output to np.int16 after dividing by 2**15 (see datasets.AudioContext).

We should also add a check for infinite and NaN values and throw a warning if they are there. (Possibly using `nan_to_num`).

lcadalzo: I assume you mean ` if not np.isfinite(x_i_copy).all():`, but this makes sense. Just pushed an update

lcadalzo: R.e. `We may want to consider quantizing the output to np.int16 after dividing by 2**15 (see datasets.AudioContext)`:

Are you saying what gets passed into `wavfile.write()` should be scaled back to int's? From looking at the datasets code, this is my understanding:

1. raw audio data is in `int` form
2. during preprocessing we divide by 2 ** 15 converting it to `float`

The `x_i` that `_export_audio()` sees is this scaled down `float` version, and currently it passes this along to `wavfile.write()`.

If you call `get_sample()`, on the other hand, per your recommendation earlier I'm scaling the float array back to ints after multiplying by 2 ** 15. Does that all make sense? Do you have thoughts on whether the Audio `get_sample()` method should

1. take in a `dataset_context` arg, from which `dataset.quantization` is obtained
2. take in a `quantization` arg where you pass the int used to scale back up the sample
3. take no extra arg, and always scale up with 2 ** 15

lcadalzo:
@shenshaw26 can you take a look at 8a0f490 ?

armory/scenarios/carla_object_detection.py
```diff
@@ -30,11 +30,11 @@ def load_dataset(self):
 
     def next(self):
         super().next()
-        self.y_object = [self.y[0]]
         self.y_patch_metadata = [self.y[1]]
+        self.y = [self.y[0]]
```
davidslater, line 34: Is y a 2-tuple? If so, it might be good to show that we are not throwing anything away.
I.e., 
```
self.y, self.y_patch_metadata = [[y_i] for y_i in self.y]
```

davidslater:
Do we have a reason we're outputting things as a pkl instead of some other format?

lcadalzo:
> Do we have a reason we're outputting things as a pkl instead of some other format?

as opposed to using Numpy's save/load functionality? I think [Numpy](https://numpy.org/doc/stable/reference/generated/numpy.save.html) pickles behind-the-scenes anyway. 

Numpy arrays aren't JSON serializable, which I think is the main reason we haven't been using JSON. But @swsuggs has a [PR](https://github.com/lcadalzo/armory/pull/3) up to save bounding box data in COCO-formatted JSON's

davidslater:
@lcadalzo we just merged in the poisoning PR. Can you ensure that `armory/scenarios/poison.py` uses the updated export_samples?

lcadalzo:
> @lcadalzo we just merged in the poisoning PR. Can you ensure that `armory/scenarios/poison.py` uses the updated export_samples?

after looking more into this, the `poison.py` is currently completely disconnected from any of the sample export logic (irrespective of the changes introduced in this PR). I think this could warrant a separate issue, which I'll go ahead and create, to address this and that this shouldn't necessarily block this PR from being merged

davidslater:
Sounds good!

armory/utils/export.py
davidslater: I would probably pass in the dataset_context. Then you can at least assert that the context input_type is np.int16 and the quantization is 2**15.
https://github.com/twosixlabs/armory/blob/develop/armory/data/datasets.py#L720

One corner case of note: after scaling up, you'll need to clip any 2**15 values to 2**15 - 1. (You can just clip by context.input_min and context.input_max).

lcadalzo:
> You should only need to use `input_min` and `input_max`, no +/- 1. The AudioContext already has it set:

Got it, updated

swsuggs:
> after looking more into this, the `poison.py` is currently completely disconnected from any of the sample export logic (irrespective of the changes introduced in this PR). I think this could warrant a separate issue, which I'll go ahead and create, to address this and that this shouldn't necessarily block this PR from being merged

@lcadalzo were you looking at poison.py as it currently exists on the develop branch?  It loads a sample_exporter [here](https://github.com/twosixlabs/armory/blob/11c02c3bc843230a2fdb2b81950d82fccfebca8b/armory/scenarios/poison.py#L289) and calls export() [here](https://github.com/twosixlabs/armory/blob/11c02c3bc843230a2fdb2b81950d82fccfebca8b/armory/scenarios/poison.py#L359).  It's fine if you still think this should go on a separate pr, just wanted to check

lcadalzo:
I was looking at that file on my `1273-refactor-export-samples` branch, which was branched off of `develop`. But I noticed that after pulling the latest changes from `develop`, changes were made to `poison.py` to include `SampleExporter`. 

Nevertheless, there's work to be done updating this code to the changes in `Exporters` made by this PR, so I'll look into that. I don't think this necessarily needs to hold up the merge

lcadalzo:
@swsuggs can you test poison exporting following the recent commit? As previously noted in a comment, `x_adv` is exported even when it's un-poisoned i.e. equal to `x`. While this doesn't seem very useful, I went ahead and pushed the recent commit since this fact predates this PR (and, if necessary, I think should be addressed in a separate poison-specific PR)

swsuggs:
Exporting in poison.py works, thanks @lcadalzo.  The actual functionality can be adjusted towards usefulness in another PR, once the poisoning group gets an idea of what they want

davidslater (approved):
This addressed my concerns, looks good.

PR #1299, opened 2022-02-23
bump develop version to 0.14.6

mwartell:
There was a source-only release which did not bump the version numbers outside of branch `no-docker-mode`. Since no-docker-mode might never be integrated to develop/master, I've cut this catch-up change.

This version represents perhaps the next version to ship. It IS NOT the version that we just shipped (as is often the case for typical releases).


davidslater (approved):
LGTM

PR #1308, opened 2022-03-07
Externally downloaded repos (requirements in armory repo but not in docker image)

### LINKED ISSUE #1302: untracked pygoturn dependency in carla baseline models
In [armory/baseline_models/pytorch/carla_goturn.py](https://github.com/twosixlabs/armory/blob/2b8dfe5925cb2563ce899e68f2555c129b00d410/armory/baseline_models/pytorch/carla_goturn.py#L6-L9) there is an import from a module which is required to be present but isn't in our source tree.

The problem is that pygoturn was never released on github nor on pypi so neither

    pip install pygoturn
    pip install git+https://github.com/amoudgl/pygoturn.git@1785ca7106fc0aa9e0fd03b440f34454e8a10e78 

work. 

The source repository hasn't had a commit since 2019, so I consider it effectively closed. I think the best approach is to fork the repo, add a pyproject.toml thus making the forked repo installable.

davidslater:
Fixes #1302

When the repo is not available, you should get a nice logging error message. To replicate, you can launch a pytorch container and run interactively: 
```
I have no name!@54d56b3aa2eb:/workspace$ python
Python 3.8.10 (default, Jun  4 2021, 15:09:15) 
[GCC 7.5.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import logging
>>> import coloredlogs
>>> coloredlogs.install(logging.DEBUG)
>>> from armory.baseline_models.pytorch import ucf101_mars
2022-03-07 21:25:58 54d56b3aa2eb armory.errors[25] ERROR MARS is an external repo.
Please download from https://github.com/yusong-tan/MARS and place on local environment PYTHONPATH
    OR place in experimental config `external_github_repo` field.
    See scenario_configs/ucf101_baseline_finetune.json for an example.
  File "/workspace/armory/baseline_models/pytorch/ucf101_mars.py", line 19, in <module>
    from MARS.opts import parse_opts
ModuleNotFoundError: No module named 'MARS'
```

When it is available (e.g., when using the basic configs), it shouldn't make any noise.

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -36,7 +36,7 @@ chmod 755 .git/hooks/pre-commit
 ```
 
 # Import Style
-Imports in python files should be organized into three blocks, after the docstring, and before other code:
+Imports in python files should be organized into three blocks, lexically ordered, after the docstring, and before other code:
 * Block 1: built-in package imports
 * Block 2: external package imports
 * Block 3: internal package imports
@@ -54,9 +54,8 @@ import requests
 import numpy as np
 from art import defences
 
-from armory.docker.management import ManagementInstance                                   
+from armory.docker.management import ManagementInstance
 from armory.utils.external_repo import download_and_extract_repos
-
 
 logger = logging.getLogger(__name__)
 # ...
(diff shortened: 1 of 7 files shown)


davidslater:
See #1325

PR #1309, opened 2022-03-08
move ucf101 clipping to data

### LINKED ISSUE #1150: UCF101 max frame clipping in data loading
Clipping during data loading, prior to getting into model

davidslater:
Fixes #1150 

This change is computationally helpful, as it moves clipping to the beginning of the preprocessing pipeline instead of the model's forward pass (and implicitly, the gradient computation).

Can test the functionality directly with
```
import numpy as np
from armory.data import datasets
max_frames = 5
x,y = next(datasets.ucf101(batch_size=1, max_frames=max_frames))
assert x.shape[1] == max_frames

x,y = next(datasets.ucf101(batch_size=2, max_frames=max_frames))
assert x.dtype == np.object
assert all([x_i.shape[0] == max_frames for x_i in x])
```

Otherwise, the scenarios should work as before.

### DIFF
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -1355,6 +1355,28 @@ def resisc10(
     )
 
 
+class ClipFrames:
+    """
+    Clip Video Frames
+        Assumes first two dims are (batch, frames, ...)
+    """
+
+    def __init__(self, max_frames):
+        max_frames = int(max_frames)
+        if max_frames <= 0:
+            raise ValueError(f"max_frames {max_frames} must be > 0")
+        self.max_frames = max_frames
+
+    def __call__(self, batch):
+        if batch.dtype == np.object:
+            clipped_batch = np.empty_like(batch, dtype=np.object)
+            clipped_batch[:] = [x[: self.max_frames] for x in batch]
+            print(clipped_batch.shape, clipped_batch.dtype, batch.shape, batch.dtype)
+            return clipped_batch
+        else:
+            return batch[:, : self.max_frames]
+
+
 def ucf101(
     split: str = "train",
     epochs: int = 1,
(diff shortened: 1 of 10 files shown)


armory/data/datasets.py
```diff
@@ -1355,6 +1355,28 @@ def resisc10(
...
+        max_frames = int(max_frames)
+        if max_frames <= 0:
+            raise ValueError(f"max_frames {max_frames} must be > 0")
+        self.max_frames = max_frames
+
+    def __call__(self, batch):
+        if batch.dtype == np.object:
+            clipped_batch = np.empty_like(batch, dtype=np.object)
+            clipped_batch[:] = [x[: self.max_frames] for x in batch]
+            print(clipped_batch.shape, clipped_batch.dtype, batch.shape, batch.dtype)
```
lcadalzo, line 1374: this print statement should I removed I assume?

armory/data/datasets.py
```diff
@@ -1365,13 +1387,21 @@ def ucf101(
     cache_dataset: bool = True,
     framework: str = "numpy",
     shuffle_files: bool = True,
+    max_frames: int = None,
```
lcadalzo, line 1390: do we want the default to be all frames or set to 512? I think the former makes more sense, which is what this PR is currently doing, but just want to confirm since I might not remember the full context of this issue

armory/data/datasets.py
```diff
@@ -1355,6 +1355,28 @@ def resisc10(
...
+        Assumes first two dims are (batch, frames, ...)
+    """
+
+    def __init__(self, max_frames):
+        max_frames = int(max_frames)
+        if max_frames <= 0:
+            raise ValueError(f"max_frames {max_frames} must be > 0")
+        self.max_frames = max_frames
+
+    def __call__(self, batch):
```
lcadalzo, line 1370: I believe an assumption made here is that a user will always want the _first_ N frames. Is that fair?

lcadalzo:
I wanted to see what would happen if `max_frames` is set in both the dataset and model kwargs: it looks like whatever is the smaller number is effectively what gets used, and this doesn't yield an error. I think this is ok but just pointing it out

armory/data/datasets.py
davidslater: Yep, removing.

armory/data/datasets.py
davidslater: Default for dataset should be all; setting to 512 in scenario_configs, however.

armory/data/datasets.py
davidslater: Correct. If the user wants a more complicated sampling/slicing mechanism, I would rather they set max_frames to None and instead implement the mechanism in their model code.

lcadalzo:
yep I did see the warning

davidslater:
Good to merge? I'm sure that removing the print statement didn't cause the TF1 error.

PR #1292, opened 2022-02-17
remove unused import statement

davidslater:
Fixes #1291

PR #1312, opened 2022-03-11
Interactive prompt fix

davidslater:
It turns out that the import in #1292 was actually used, but indirectly in the prompt for interactive mode. (I ran into this while using interactive mode today.) I don't like the warning that the import caused, so I have rewritten the interactive prompt statement to import directly from the `armory.scenarios.main`.


davidslater:
When we update some of the command line things, we should have the statement more programmatically generated or tested (maybe using `exec`?)

mwartell (approved):
LGTM.

PR #1323, opened 2022-03-16
Tfds v4, ART v1.10, and other docker/build updates 

### LINKED ISSUE #1282: Consider TFDS version 4

### LINKED ISSUE #1319: Integrate the new ART release (1.10)
This should also:
a) solve the TF1 compatibility issue that @shenshaw26 was seeing
b) fix the deep speech v3.0 issue we were seeing in the pytorch-deepspeech docker image and enable us to directly use ART instead of my fork.

davidslater:
Fixes #1282 (partially - dependencies and version migration)
Fixes #1319 

This PR requires an update to the base image.

I think we should make this (small) update now so that we have a solid base for docker images, and then move to the other dataset stuff.

This also makes it easier for developers by having armory look for the most recent release-version docker image.


armory/eval/evaluator.py
```diff
@@ -84,20 +84,28 @@ def ensure_image_present(self, image_name: str) -> str:
         # look first for  the versioned and then the unversioned, return if hit
         # if there is a tag present, use that. otherwise add the current version
         if ":" in image_name:
-            check = image_name
+            checks = (image_name,)
         else:
+            # TODO: This needs to be fixed if 'image_name' does not refer to twosixarmory image
+            #   There should be a more explicit check for specific armory image names.
             check = f"{image_name}:{armory.__version__}"
+            check_previous = ".".join(check.split(".")[:3])
```
mwartell, line 92: In some discussion we considered this order
1. major.minor.patch.hash
2. major.minor.patch
3. major.minor

To cut down on docker builds on every commit. This code implements (1) and (2) but not (3). I just wanted to check that omission was intentional. If we find we want (3) we can always add it later.

docker/Dockerfile-base
```diff
@@ -51,6 +51,11 @@ RUN /opt/conda/bin/conda install \
     cudatoolkit=11.3 \
     -c pytorch \
     && /opt/conda/bin/conda clean --all
+# pytorch==1.10.2
```
mwartell, line 54: This looks like dead code was added??

docker/Dockerfile-base
```diff
@@ -62,7 +67,7 @@ RUN /opt/conda/bin/conda install \
     # ART requires scikit-learn >=0.22.2,<1.1.0
 # scikit-learn==1.0.2 \
 # scipy==1.7.3 \
-# jupyterlab==3.2.9 \
+# jupyterlab==3.3.2 \
```
mwartell, line 70: this upgraded dead code which is still dead

docker/Dockerfile-base
```diff
@@ -80,14 +85,16 @@ RUN /opt/conda/bin/conda install \
     pandas=1.1 \
     -c anaconda \
     && /opt/conda/bin/conda clean --all
+# pandas==1.1.3
```
mwartell, line 88: added dead

docker/Dockerfile-base
```diff
@@ -80,14 +85,16 @@ RUN /opt/conda/bin/conda install \
...
 
 # Conda Install for packages with no channel
 RUN /opt/conda/bin/conda install \
     protobuf=3 \
     && /opt/conda/bin/conda clean --all
+# protobuf==3.19.1
```
mwartell, line 94: added dead

docker/Dockerfile-base
```diff
@@ -97,23 +104,17 @@ RUN /opt/conda/bin/pip install --no-cache-dir \
...
+# boto3==1.20.0 \
+# opencv-python==4.5.5.64 \
 # ffmpeg-python==0.2.0 \
 # pytest==7.0.1 \
-# pymongo==4.0.1 \
+# pymongo==4.0.2 \
 # coloredlogs==15.0.1 \
-# docker==5.0.3 \
-# adversarial-robustness-toolbox==1.9.1
+# docker==5.0.3
```
mwartell, line 117: more upgraded dead

docker/Dockerfile-tf2
```diff
@@ -3,30 +3,29 @@ FROM twosixarmory/base:${base_image_tag} as armory-base
...
 
-RUN /opt/conda/bin/pip install --no-cache-dir \
-    adversarial-robustness-toolbox==1.9.1
-
 # This section installs tensorflow object detection API from
 # https://github.com/tensorflow/models/blob/79354e14a4b41ff9019f4a5ebd12cfa498917182/research/object_detection/packages/tf2/setup.py
 # It is messy
-RUN /opt/conda/bin/conda install protobuf==3.14.0 && \
-    /opt/conda/bin/conda clean --all
+#RUN /opt/conda/bin/conda install protobuf==3.14.0 && \
```
mwartell, line 9: the general recommendation is to let git keep track of dead code, not the code itself

mwartell (changes requested):
Overall a good omnibus PR.  I've noted some specific oddities such as addition of new dead code which seems like it could result from a merge error or somesuch.

armory/eval/evaluator.py
davidslater: I omitted it because we haven't published any images with major.minor

.github/workflows/end2end_cron.yml
```diff
@@ -34,7 +34,8 @@ jobs:
           imageTag=$(python setup.py --version | sed  -e 's/dev[0-9][0-9]*+//' -e 's/\.d[0-9][0-9]*$//')
           echo "Image Tag -> $imageTag"
           echo "::set-output name=imageTag::$imageTag"
-          bash docker/build.sh base pytorch pytorch-deepspeech tf2 --tag $imageTag --base-tag $imageTag
+          bash docker/build-base.sh
```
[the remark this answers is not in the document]
davidslater: Yeah, I agree with not building base. It is liable to produce lots of spurious errors if dependencies change.

docker/Dockerfile-tf2
davidslater: This dead code I actually meant to remove, as it is now installed in base.

davidslater:
@mwartell For the remainder of the dead code, I was keeping it primarily for debugging purposes while building the image. If you look at, e.g., `Dockerfile-base`, you'll notice that there was already a lot of this there - what I added was just to be consistent.

The fully versioned "dead code" is actually just reporting what _did_ get installed. I wanted to specify as little as possible to prevent conda and pip getting confused or unable to solve.

I think that eventually we should remove it, but I want to get our `host-requirements.txt` or equivalent files all synchronized before removing those.

PR #1325, opened 2022-03-16
Update - Externally downloaded repos (requirements in armory repo but not in docker image)

### LINKED ISSUE #1302: untracked pygoturn dependency in carla baseline models

davidslater:
Fixes #1302 

Same as PR #1308 but merges in `develop`.


armory/errors.py
```diff
@@ -0,0 +1,25 @@
+import contextlib
```
mwartell, line 1: I really like the implementation of this as a context manager; nifty

I think errors.py is a confusing name for a module that imports, please change.

docs/style.md
```diff
@@ -75,3 +75,22 @@ from armory.logs import log
...
+with ExternalRepoImport(
+    repo="colour-science/colour@v0.3.16",
+    experiment="carla_obj_det_dpatch_undefended.json",
+):
+    import colour
+```
+
+`repo` refers to the GitHub repo name (optionally with `@tag`).
+`experiment` refers to the `.json` file in the `scenario_config` directory that uses this repo.
+These repos are specifically *NOT* installed in the armory-supported docker containers and conda environments, and are downloaded at runtime.
```
mwartell, line 96: A sentence of "why these are not downloaded" might be helpful.

mwartell (changes requested):
Overall LGTM.

I think errors.py needs a rename so I've requested that change.

docs/style.md
davidslater: Added

davidslater:
@mwartell ready for re-review. I moved the context manager to a different (existing) file: `armory.utils.external_repos`

PR #1327, opened 2022-03-17
prefer native importlib.metadata if available

mwartell:
Python 3.8+ has builtin `importlib.metadata` Python 3.7 needs a feature backport named `importlib_metadata`.


mwartell:
@lcadalzo If you could confirm that this works in your Python3.9, that's all that's needed

lcadalzo (approved):
3.8, but yes this works for me. LGTM

lcadalzo:
The CI error appears to be occurring on `develop` too, so we should be good to merge?

PR #1331, opened 2022-03-17
enable log.exception backtrace

### LINKED ISSUE #1328: log.exception does not produce backtrace
Loguru sinks need to have `backtrace=True` and `diagnose=True` in order to generate backtraces. This should be the default in logs.py.

mwartell:
Fixes #1328

Enables `backtrace` and `diagnose` options to loguru which trigger when there is an active exception. This makes log.exception work like the standard logger.exception.

The `diagnose` option uses loguru's embedded [better-exceptions](https://github.com/Qix-/better-exceptions) detailed and colorful backtraces. The detail might be excessive and can be set to False but let's see.


lcadalzo (approved):
LGTM

PR #1337, opened 2022-03-18
update requirements

davidslater:


davidslater:
This will (hopefully) fix the CI test problems (I was able to get them to pass locally on native and docker).

There is considerable overlap between `host-requirements.txt` and `armory-base-requirements.txt`. I think some of that will require discussion with @shenshaw26

host-requirements.txt
```diff
@@ -4,63 +4,75 @@
 # complexity to the installation and operation of armory, so we encourage the
```
mwartell, line 4: I thought this file had been obsoleted (though not deleted).  I can find no reference to it in current `develop` except some outdated docs.

armory/data/datasets.py
```diff
@@ -422,8 +422,10 @@ def _generator_from_tfds(
...
             shuffle_files=shuffle_files,
         )
-    except AssertionError as e:
-        if not str(e).startswith("Unrecognized instruction format: "):
+    except (AssertionError, ValueError) as e:
+        if not str(e).startswith(
+            "Unrecognized instruction format: "
+        ) and "Unrecognized split format: " not in str(e):
             raise
         log.warning(f"Caught AssertionError in TFDS load split argument: {e}")
```
mwartell, line 430: nit: this message is technically incorrect: "AssertionError or ValueError" but, :shrug:

davidslater:
Still hitting snags with the end-to-end tests, but it looks like the CI is passing now. @mwartell

mwartell (approved):
LGTM. But I am not merging it. I mostly came here to note that host-requirements.txt might be obsoleted but reviewed it anyway.

armory/data/datasets.py
davidslater: This is super wonky block of code that I plan to remove with the datasets update.

PR #1338, opened 2022-03-18
add checks for evaluate_current

### LINKED ISSUE #1335: catch user error when calling evaluate_current() before next()
Currently, when using Armory interactively, if you call `evaluate_current()` before `next()` you get a somewhat confusing error:

```
  File "/workspace/armory/scenarios/scenario.py", line 340, in evaluate_current
    self.run_benign()
  File "/workspace/armory/scenarios/scenario.py", line 271, in run_benign
    x, y = self.x, self.y
AttributeError: 'ImageClassificationTask' object has no attribute 'x'
```

We can add a check in `evaluate_current()` for this that would then instruct the user to first call `next()`. Note that this has already been added to the `export_samples()` method in [this](https://github.com/lcadalzo/armory/blob/1273-refactor-export-samples/armory/scenarios/scenario.py#L325-L329) feature branch

davidslater:
Fixes #1335

### DIFF
--- a/armory/scenarios/scenario.py
+++ b/armory/scenarios/scenario.py
@@ -257,7 +257,12 @@ class Scenario:
         self.i, self.x, self.y = i, x, y
         self.y_pred, self.y_target, self.x_adv, self.y_pred_adv = None, None, None, None
 
+    def _check_x(self, function_name):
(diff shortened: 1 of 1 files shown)


lcadalzo (approved):
LGTM

PR #1339, opened 2022-03-18
replace a line incorrectly removed

mwartell:
I failed to notice this CI variable was needed when I reviewed pull #1323. I'm putting it back.

### DIFF
--- a/.github/workflows/end2end_cron.yml
+++ b/.github/workflows/end2end_cron.yml
@@ -20,6 +20,8 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v2
+        with:
+            fetch-depth: 0
       - uses: actions/setup-python@v2
         with:
           python-version: '3.7'
@@ -30,6 +32,7 @@ jobs:
           pip install -e .
           python docker/build.py all
           imageTag=$(python setup.py --version | sed  -e 's/dev[0-9][0-9]*+//' -e 's/\.d[0-9][0-9]*$//')
+          echo "passing imageTag $imageTag to later jobs"
     outputs:
       imageTag: ${{ steps.docker-build.outputs.imageTag }}
   docker-pytorch-end2end:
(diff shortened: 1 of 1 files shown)


davidslater (changes requested):
Looks like the end2end test is still failing. Maybe add an echo of the imageTag at the end of the first stage to confirm the version?

I had previously run into problems (before this line was removed) where `python setup.py --version` was outputting a 0.1.g98a7a... version. There was also a warning about the clone being "shallow", indicating that the the checkout action did not retrieve the full git history. I think what was happening there is that the git history didn't go far enough back to find a version tag, and therefore it defaulted to a version 0, incremented to 0.1 with a hash tag.

davidslater:
It looks like the environment variable imageTag is set properly in the first stage, but somehow it's not propagating to the later stages.

davidslater:
@mwartell I think there's an additional line we need to add back:
https://github.com/shenshaw26/armory/blob/duo/.github/workflows/end2end_cron.yml#L34-L36

It looks like this line does more than echo:
```
echo "::set-output name=imageTag::$imageTag"
```

davidslater:
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions

davidslater:
Testing...

davidslater:
@shenshaw26 did these end2end tests ever work on github? I think that the built docker containers would not survive for the subsequent tests (unless they got pushed to dockerhub or somewhere else). These could work if they were put into the same test, though that would seem like a rather long test.

shenshaw26:
The end-to-end tests were put together to run on the self hosted runner, the last time I messed with them, however it now appears that they are trying to run in the github runner.  @mwartell, is that change intentional?  I thought we had setup the local runner on AWS to do the end-to-end tests (I realize that we did that on our own fork, but I assumed we would migrate that bit over as well)

davidslater:
@mwartell is this update pertinent to your current PR work, or can it be closed?

mwartell:
Fixed by WIP #1422. Transmitting values between steps in actions is dodgy; I cured the problem by having each step get the tag itself.

PR #1341, opened 2022-03-21
Fix GPU usage with PyTorch / TF

davidslater:


davidslater:
This ran properly on CARLA multimodal and UCF101 with 100 samples (the latter peaked at ~16 GB GPU memory use).

armory/__init__.py
```diff
@@ -61,6 +61,33 @@ def get_dynamic_version():
...
+# From: https://www.tensorflow.org/guide/gpu#limiting_gpu_memory_growth
+try:
+    import tensorflow as tf
+
+    gpus = tf.config.list_physical_devices("GPU")
+    if gpus:
+        # Currently, memory growth needs to be the same across GPUs
+        for gpu in gpus:
+            tf.config.experimental.set_memory_growth(gpu, True)
+        logical_gpus = tf.config.list_logical_devices("GPU")
```
lcadalzo, line 82: is there any purpose in obtaining `logical_gpus`? In the TF link it's used for logging purposes, but we deleted that line.

davidslater: Nope, I'll delete.

lcadalzo:
@davidslater I think there may be a problem caused by this merge, actually. Now torch is being imported in `datasets.py` without any try/except block. This import will fail inside a tf2 container

davidslater:
The tf2 container does have torch installed. But yes, I see your point. Probably good to add back in.

PR #1344, opened 2022-03-21
Improve Docker Image Lookup

davidslater:


davidslater:
This PR also moves most of the docker-related code into `armory/docker/images.py` (migrating code from `armory.utils.docker_api` and `armory.eval.evaluator`)

armory/docker/images.py
```diff
@@ -32,3 +36,218 @@
...
+    if tag:
+        tag = ":" + tag
+    return f"{repo}{name}{tag}"
+
+
+def is_armory(image_name: str):
+    """
+    Return whether image_name refers to an armory docker image
+    """
+    repo, name, _ = split_name(image_name)
```
mwartell, line 94: technically `user` and `repository` (I just checked when writing my split_name above.

armory/docker/images.py
```diff
@@ -32,3 +36,218 @@
...
+            raise
+        except requests.exceptions.ConnectionError:
+            log.error("Docker connection refused. Is Docker Daemon running?")
+            raise
+
+    canon_image_name = get_armory_name(image_name)
+    log.info(f"Retrieved canonical image name for {image_name} as {canon_image_name}")
+
+    prev_release = last_armory_release(canon_image_name)
+    if canon_image_name != prev_release:  # currently on hashed dev branch
```
mwartell, line 174: This is spooky mysterious action as is currently in discussion. The saving grace here is that the decision tree progress is logged.

armory/docker/images.py
```diff
@@ -32,3 +36,218 @@
...
+        try:
+            docker_client.images.get(canon_image_name)
+            log.success(f"found docker image {image_name} as {canon_image_name}")
+            return canon_image_name
+        except docker.errors.ImageNotFound:
+            log.trace(f"image {canon_image_name} not found")
+            log.info(f"reverting to previous release tag image {prev_release}")
+        except requests.exceptions.HTTPError:
+            log.trace(f"http error when looking for image {canon_image_name}")
+            raise
```
mwartell, line 185: I'd probably factor out this block that repeats at 188.  I'd also drop HTTPError case or merge it with ImageNotFound. The interesting case at 215 ConnectionError does give actionable help. This decoration doesn't.

armory/docker/images.py
```diff
@@ -32,3 +36,218 @@
...
+        repository, tag = repository.split(":")
+    elif ":" in repository:
+        raise ValueError(
+            f"cannot set tag kwarg and have tag in repository arg {repository}"
+        )
+    elif tag is None:
+        log.info("empty tag is set to latest by API")
+        tag = "latest"
+
+    if not is_progress():
```
mwartell, line 235: thanks for preserving `not is_progress`; I wondered if that was a Good Thing or just a good thing to me.

mwartell (approved):
LGTM. My comments are advisory only.

armory/docker/images.py
davidslater: Good! I was looking for the correct naming convention.

armory/docker/images.py
```diff
@@ -32,3 +36,218 @@
 # way to find out
 def is_old(tag: str):
     raise NotImplementedError
+
+
+def split_name(image_name: str):
```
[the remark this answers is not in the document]
davidslater: The user can have multiple '/' characters. E.g., https://cloud.google.com/artifact-registry/docs/docker/names

If there are multiple ":" characters, then this line will fail with a ValueError:
```name, tag = s.split(':')```

I like the simplification, though.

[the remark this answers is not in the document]
mwartell: good to know about multiple slashes, thanks

davidslater:
Could use a quick re-review @mwartell

PR #1346, opened 2022-03-22
remove mongoDB functionality

davidslater:


mwartell (approved):
Finally! I've been wanting to do this for nigh on a year now every time I bump into that dead code.

PR #1347, opened 2022-03-22
WIP: Docker dev fix

shenshaw26:
Trying to build work around so developers can dynamically change armory code on host machine and have it automagically update the pip installed code in the docker container.

### DIFF
--- a/docker/Dockerfile-pytorch
+++ b/docker/Dockerfile-pytorch
@@ -8,14 +8,19 @@ RUN echo "Installing ART"
 RUN /opt/conda/bin/pip install --no-cache-dir \
     adversarial-robustness-toolbox==1.10.0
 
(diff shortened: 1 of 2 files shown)


davidslater:
Yeah, this is pretty gross.

What puzzles me is why we would copy/paste and pip install at runtime instead of just mapping in the host code of armory. Is pip install even necessary?

If we want to remove ambiguity on a dev run, what about a simple `pip uninstall armory` followed by just using the mapped in armory code?

shenshaw26:
Yeah, I think we should think through the use cases and then we can make it consistent throughout.  For that discussion, seems like we can use: https://github.com/twosixlabs/armory/issues/1369.

If we don't need to pip install armory in the container, that makes it simple in that the user can specify at `docker run...` where to be and what to run.

PR #1352, opened 2022-03-22
provide consistent ART data pathing

### LINKED ISSUE #805: Fix ART directory pathing in container

### LINKED ISSUE #1280: Update ART data path inside container
Currently, ART data path defaults to `/tmp/.art/data`, which is destroyed after container teardown.

In `--no-docker` mode, ART config can be set separately.

This can be fixed a few different ways.
1) When armory imports `art`, do the following:
```
art.config.set_data_path("/armory/datasets/art/")
```

2) Add an ART config file to the container. However, this part in ART will default to `/tmp/`, which makes places a config difficult:
https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/art/config.py#L36-L39

3) Modify the armory configure config to include an ART path and map that to "/armory/datasets/art/" via docker.

davidslater:
Fixes #1280 
Fixes #805 

This ensures that ART data pathing is consistent across Armory use.

Currently, it does this in Deep Speech, but not other models.

### DIFF
--- a/armory/__init__.py
+++ b/armory/__init__.py
@@ -18,6 +18,7 @@ try:
 except ModuleNotFoundError:
     from importlib_metadata import version, PackageNotFoundError
 
+import os
 import pathlib
 import re
 import subprocess
@@ -86,6 +87,12 @@ except RuntimeError:
 except ImportError:
     pass
 
+# Handle ART configuration
+
+from armory import paths
+from art import config
+
+config.set_data_path(os.path.join(paths.runtime_paths().saved_model_dir, "art"))
 
 # typedef for a widely used JSON-like configuration specification
 from typing import Dict, Any
--- a/armory/baseline_models/pytorch/deep_speech.py
+++ b/armory/baseline_models/pytorch/deep_speech.py
@@ -4,13 +4,8 @@ Automatic speech recognition model
 Model contributed by: MITRE Corporation
 """
 
-import os
 from typing import Optional
 
-from armory import paths
-from art import config
-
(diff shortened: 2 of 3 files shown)


mwartell (approved):
LGTM

PR #1353, opened 2022-03-22
WIP: separate engine from launcher

shenshaw26:
Proposed Changes:

- [ ] update main armory entrypoint to use `armory.cli`.  This will handle the `engine` execution 
- [ ] add `launcher` tooling that is separate, distinct, and minimal to handle launching armory natively or inside docker, managing environment and resources
- [ ] move from `scenario` having a double meaning to an `experiment` which is basically the `config.json` and `scenario` indicating the type of evaluation being considered (image classifcation vs object detection vs audio things, etc.).


shenshaw26:
Note this is super early, just getting it out there for awareness.

armory/experiment.py
```diff
@@ -0,0 +1,155 @@
...
+    #     with open(filename, "w") as f:
+    #         f.write(self.json())
+
+    @classmethod
+    def load(cls, filename, overrides=[]):
+        overrides = parse_overrides(overrides)
+        valid_ext = (".aexp", ".json")
+        fname, fext = os.path.splitext(filename)
+        log.info(f"Attempting to Load Experiment from file: {filename}")
+        if fext == ".json":
```
mwartell, line 119: Confusing polymorphism. Don't give the user two options when one will do, especially since you can yaml.load scenario_configs/cifar10_baseline.json because yaml is a superset of json.

armory/experiment.py
```diff
@@ -0,0 +1,155 @@
...
+            with open(filename, "r") as f:
+                data = json.loads(f.read())
+        elif fext in (".aexp", ".yml", ".yaml"):
```
mwartell, line 122: Easier to use `.aexp` as a convention and not care about the file extension. yaml.load will fault on non-yaml and we don't have to test

tests/scenarios/experiments/base_image_classification.aexp
```diff
@@ -0,0 +1,42 @@
+# ###################################################################
```
mwartell, line 1: gee this is sooo much nicer than json

tests/scenarios/experiments/base_image_classification.aexp
```diff
@@ -0,0 +1,42 @@
...
+  weights_file: null
+  wrapper_kwargs: {}
+
+scenario:
+  kwargs: {}
+  module_name: "armory.scenarios.image_classification"
+  function_name: "ImageClassificationTask"
+
+overrides: []
+#
```
mwartell, line 40: I looked at "comment preserving yaml parsers" for about 2 minutes and decided to not use them here (if anywhere, ever)

tests/unit/test_environment.py
```diff
@@ -0,0 +1,98 @@
+import pytest
```
mwartell, line 1: tests with code, neato! :star:

update_armory_version.py
```diff
@@ -0,0 +1,48 @@
+""" Update Armory Version
```
mwartell, line 1: Isn't this mechanism dead since duo stripped out all versions in scenarios??

mwartell (commented):
The general theme here is start clean, it will get complicated later.

This does mostly track what we'd discussed as intent. Thanks.

PR #1357, opened 2022-03-23
WIP: Initial try at using tfds.builder in `datasets.py`

shenshaw26:
Making a new `dataset_builder` package that will construct `tfrecords` for each of the armory datasets and tooling to save the results to s3 for use by the armory tooling.


shenshaw26:
running `pytest -vs ./tests/unit/test_datasets.py::test_dataset_generation[tf-test-mnist]` fails because the tfds dataset returns a dictionary.  For example for `mnist` you get `{'image':<tf.Tensor: shape=(28, 28, 1)...>, 'label':<tf.Tensor: shape=(), ...>}`.  Need to figure out what we "want" this to be by default.

shenshaw26:
@mwartell @davidslater , I have pushed my current status on this PR.  Given that I am going to be out on PTO till 22 APR, just wanted to give you a status here in case y'all wanted to run with this.

Current State:
- I believe that the refactor of the dataset_builder components into tools/dataset_builder is largely complete.  There are a few issues with a few datasets but overall the methodology is working.  What needs to happen next is that the build and upload need to be run for all of the SUPPORTED_DATASETS (e.g. `python build.py -ds all --clean` && `python upload.py -p ~/.armory/dataset_builds/`) but note that this should done on a machine with enough space to generate all the datasets
- There are probably more tests to write, but I think we have decent coverage there
- There is more to clean out from `armory.data` but I would like to get some more rigorous testing in place before I go mucking around much more in there

Notes:
  - With the new namespace modules business I separated out the `tests` for `dataset_builder` because it makes it cleaner to understand.  I have not, yet, added CI tests to run those tests, but that is on the list.  
  
Hope this helps!  Let me know if you have questions or need anything while I am on break I will be checking things periodically.

shenshaw26:
@davidslater, I have refactored this PR, per our discussion with a couple notes:

- The `armory.datasets.builder` module contains files that are not in line with the rest of armory (e.g. `requirements.txt` and `README.md`, however, I don't have a better suggestion for placement as they will change with the introduction of `tiga/major-revamp`.  Do you have a preference?
- I am still working on the move from `armory.data` to `armory.datasets.loader` but will try to have something tomorrow.

davidslater:
@shenshaw26 I think it's fine having those top-level things in the `builder` subdir. Let's put them in there, and maybe we can have a link from our main docs to the builder top level.

shenshaw26:
Hey @davidslater @mwartell.  This is now working for the `mnist` baseline in `no-docker` mode.  To try run:
```
armory run -d ./tests/scenarios/datasets/mnist_baseline.json --check --no-docker
```

I am not a huge fan of the `preprocessing` implementation but it works as before and is now more configurable from the scenario.json file.   I would suggest that we revisit once we move over to the new dataclasses bits that are in `major-revamp` PR.

mwartell:
@christopherwoodall we should take a look at this for code or principles, it looks like it could be reasonably isolated. Look at the README.md and build.py primarily. I have scheduled a determination session for it with you.

PR #1361, opened 2022-03-23
remove armory clean

davidslater:
It has been forever since I've referenced or used.


davidslater:
@mwartell more old stuff to remove :)

mwartell (approved):

PR #1363, opened 2022-03-23
update scenario to properly override Scenario

### LINKED ISSUE #1343: Investigate metrics warning for CARLA default multimodal config
Running the following command:
```
python -m armory run scenario_configs/carla_obj_det_multimodal_dpatch_undefended.json
```

I'm seeing the following warning:
```
2022-03-21 21:18:26 13s WARNING  armory.utils.metrics:__init__:1605 No metric results will be produced. To change this, set one or more 'task' or 'perturbation' metrics
```

even though the config does specify metrics.

### LINKED ISSUE #1354: fix bug when trying to run carla scenario step-by-step
To reproduce the error:

```
>>> from armory.scenarios.main import get as get_scenario
>>> s = get_scenario("/armory/tmp/2022-03-22T203253.148375/interactive-config.json").load()
>>> s.next()
>>> s.evaluate_current()
```

This should work, but it yields the following error:

```
  File "/workspace/armory/scenarios/scenario.py", line 346, in evaluate_current
    self.run_attack()
  File "/workspace/armory/scenarios/carla_object_detection.py", line 73, in run_attack
    self.metrics_logger_wrt_benign_preds.update_task(
AttributeError: 'CarlaObjectDetectionTask' object has no attribute 'metrics_logger_wrt_benign_preds'

```

davidslater:
Fixes #1354
Fixes #1343 

Also fixes an unnoticed bug where if `"record_metric_per_sample": true` in config, it wouldn't apply to the CARLA-specific added metrics.

### DIFF
--- a/armory/scenarios/carla_object_detection.py
+++ b/armory/scenarios/carla_object_detection.py
@@ -4,8 +4,6 @@ CARLA object detection
 Scenario Contributor: MITRE Corporation
 """
 
-import copy
-
 from armory.scenarios.scenario import Scenario
 from armory.utils import metrics
 from armory.logs import log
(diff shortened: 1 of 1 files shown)


lcadalzo:
This mostly fixes things, but one minor thing: users have the ability to provide `"task_kwargs"` in the metric config, see [here](https://github.com/twosixlabs/armory/blob/master/docs/configuration_files.md#specifying-kwargs-for-metric-functions). These currently won't get included in `subset_config`. We could add this string to the tuple or do something like  `for k in metric_config if k != perturbation`.

PR #1380, opened 2022-03-25
remove madry training

davidslater:
Madry training is now in ART.


davidslater:
The CIFAR config used our local madry version.

lcadalzo (approved):
LGTM

Issue #1369, opened 2022-03-25
SCM Versioning - Armory fails when using code snapshots


davidslater:
One side effect of the SCM versioning method in develop is that armory must either be pip installed or be in a valid git repo in order to be used.

Snapshots from tags, such as https://github.com/twosixlabs/armory/tags, will fail. Armory cannot be imported from them and it also cannot be pip installed from them.

To replicate, download a snapshot of the develop branch: https://github.com/twosixlabs/armory/archive/refs/heads/develop.zip
Then:
```
cd armory-develop
pip install -r requirements.txt  # this will work
pip install -e .  # this will fail 
python setup.py --version # this will fail
python -m armory  # this will fail
python
>>> import armory  # this will fail
```
They all fail due to `python setup.py --version` failing. And since this happens in `armory/__init__.py`, it causes everything in the library to fail.

Is this also why we need to send an SETUPTOOLS_SCM_PRETEND_VERSION into the container when pip installing?

shenshaw26:
Yes, this is expected and is why we need `SETUPTOOLS_SCM_PRETEND_VERSION`.  Is this a common use case?  If so, maybe we need to consider an alternative in `__init__.py` where it defaults to version=UNKNOWN or something.  This would still allow the overrides but also allow use of armory even when it is not installed.  

Just to try to keep my mind clear about use-cases, its seems I am tracking the following:
- Run Armory locally with pip installed, using wheel from pypi
- run armory locally with/without pip installed from git clone
- run armory locally with/without pip installed from snapshot
- run armory within docker container where armory is prebuilt inside the container
- run armory within docker container where armory-dir is mounted into container and then overrides the pre-built version (not sure if this needs to support both pip installed of mounted drive or can just be run at repo root)
- others?

shenshaw26:
@davidslater @mwartell , In poking at this a bit more, its unclear to me how all the different combinations should behave.  For example, for the simple cases we have:

| armory on PYTHON_PATH | armory pip installed | Expected Result |
|-----------------------|----------------------|-----------------|
| No | No | Should NOT work <br> ModuleNotFoundError: No module named 'armory'|
| No | Yes | should work <br> `__version__ = importlib.metadata.version("armory-testbed")` |

which makes sense.  However when you get to the more complicated cases (e.g. armory src on PYTHON_PATH) I am not sure what should happen:

| armory on PYTHON_PATH | armory pip installed | Expected Result |
|-----------------------|----------------------|-----------------|
[...]

PR #1381, opened 2022-03-28
WIP: Fix Armory to support execution modes

shenshaw26:
This PR is intended to address: #1369 as well as define in tests/CI the supported execution modes.

Supported Armory Engine execution modes:

Local / Native Execution modes
 - [ ] Run armory locally with it pip installed, using wheel from pypi
 - [ ] Run armory locally with and without it pip installed, using git clone
 - [ ] Run armory locally with and without it pip installed, using git created snapshots

Docker Execution modes
 - [ ] run armory inside docker using pre-built armory from image
 - [ ] run armory inside docker using externally mounted armory-repo directory (NOT pip installed)


davidslater (changes requested):
Three issues.
1) Since the `PackageNotFoundError` is raised in the next line, the EXPERIMENTAL_UNSUPPORTED version is never returned.
2) Since `setup.py` is present, this will actually error out in line 49 in the `get_dynamic_version()` function call.
3) It's still not clear to the user what the desired remediation is.

Recommended fix:
1) remove line 61 so that the version is used.
2) change `raise` to `return None`
3) Add some additional user facing recommended remediation after line 59. Maybe something like
`log.critical("Please see armory docs [link] to see supported execution modes. Recommended uses are `pip install armory-testbed` and (for developers) a locally cloned repo with full git history.")

mwartell:
Obsoleted by various CI and SCM version changes in https://github.com/twosixlabs/armory/releases/tag/v0.16.0

Issue #1285, opened 2022-02-16
Enable export samples to grab data after preprocessing steps are applied?


davidslater:

lcadalzo:
@davidslater could you elaborate on this one?

davidslater:
This was recording a conversation we had earlier about folks that wanted to export samples after a preprocessing defense (I think primarily for CARLA scenarios).

My guess is that we punt on this for now, but reconsider it after the measurement update is done?

lcadalzo:
Got it, I think this overlaps with what we're trying to accomplish with the measurement update so I agree it makes sense to wait for that.

PR #1387, opened 2022-03-28
Measurement

### LINKED ISSUE #1045: Error arises when a scenario's metric updates are all targeted
During the attack stage of most scenarios, regardless of whether the attack is targeted, the MetricLogger's `update_task()` is first called with ground-truth labels  (i.e. `update_task(..., adversarial=True, targeted=False)`). This [populates](https://github.com/twosixlabs/armory/blob/v0.13.0/armory/utils/metrics.py#L1239) the metric in `self.adversarial_tasks` list with results.

A problem arises if a scenario never calls `update_task()` with `targeted=False` (i.e. with ground-truth labels). Note: this is the case for DAPRICOT, e.g.. The metric in the `self.adversarial_tasks` list is thus never populated with results because [this](https://github.com/twosixlabs/armory/blob/v0.13.0/armory/utils/metrics.py#L1224) variable never refers to `self.adversarial_tasks`. When the mean is computed [here](https://github.com/twosixlabs/armory/blob/v0.13.0/armory/utils/metrics.py#L1351), this causes an error. [This](https://github.com/twosixlabs/armory/blob/v0.13.0/armory/utils/metrics.py#L1334-L1335) workaround was added for DAPRICOT metric but this obviously doesn't address the underlying issue (note: an additional workaround is to set `metrics_logger.adversarial_tasks = []` in the scenario code as is done [here](https://github.com/lcadalzo/armory/blob/1039-dapricot-scale/armory/scenarios/dapricot_scenario.py#L121-L122)).

The `self.adversarial_tasks` variable is initialized [here](https://github.com/twosixlabs/armory/blob/v0.13.0/armory/utils/metrics.py#L1172) to be non-empty, but it will never actually be updated in this case since it really refers to `self.adversarial_nontargeted_tasks` (see [this](https://github.com/twosixlabs/armory/blob/v0.13.0/armory/utils/metrics.py#L1224-L1230) logic) of which there may be none.

### LINKED ISSUE #1424: Build probe/meter framework for flexible measurement

davidslater:
Fixes #1424 
Fixes #1045
Lays the groundwork for other metrics updates.

This is a skeleton for updating the measurement machinery.

Once we have settled on solid machinery and have that fully tested, then we can integrate into scenarios. However, I would like to keep those separate for now.

Todo:
- [x] hello world example
- [x] handling “global” metrics and averages like “mean”, "mAP", and "word error rate"
- [x] functionality to set up meters from current config files
- [x] ability to write to results json in expected format
- [x] demonstration with scenario
- [x] expand to all evasion scenarios
- [x] add tests
- [x] integrate entailment updates
- [x] expand to poisoning scenarios
- [x] ensure functional parity
- [x] remove old MetricsLogger code
- [x] remove old MetricList code
- [x] update documentation


davidslater:
[Reed] - Enable additions of attributes to the data object (e.g., whether a data point is poisoned)

armory/instrument/instrument.py
```diff
@@ -0,0 +1,552 @@
...
+        """
+        Primarily intended for metrics of metrics, like mean or stdev of results
+        """
+        pass  # TODO
+        # raise NotImplementedError()
+
+    def results(self):
+        if not self.keep_results:
+            raise ValueError("keep_results is False")
+        return copy.deepcopy(self._results)
```
mwartell, line 360: I find gratuitous copies cause more trouble than they avert. Since the test rig doesn't call this method, I can't tell the motivation for this.

mwartell (commented):
Overall, this is nicely heading where you want it to be.

If I read this correctly, you would want a config-file representation for the add_meter calls. I'll be thinking of a tiga compliant way for that.

armory/instrument/instrument.py
```diff
@@ -0,0 +1,552 @@
...
+        """
+        Return a list of (meter, arg) that are using the current variable
+        """
+        filter_map = self.probe_filter_meter_arg.get(probe_variable, {})
+        meters = filter_map.get(stage, [])
+        meters.extend(filter_map.get(None, []))  # no stage filter (default)
+        return meters
+
+    def get_meters(self):
+        return self.meters[:]
```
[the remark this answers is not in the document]
davidslater: I think I'm storing "meters" in too many places. I'm thinking that ProbeMapper should be a very specific data structure, and remove this method. Just have "add_meter", "remove_meter", and map methods essentially.

armory/instrument/instrument.py
```diff
@@ -0,0 +1,552 @@
...
+    else:
+        stage_filter = None
+    probe_variable = arg
+
+    return probe_variable, stage_filter
+
+
+class ProbeMapper:
+    """
+    Map from probe outputs to meters
```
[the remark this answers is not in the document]
davidslater: I envisioned it as a data structure primarily used by Context, but with easily separable functionality - I figured it would be easier to test as a separate class, and also easier to subclass Context.

For me the entire purpose of ProbeMapper was to map from probe variables to meter inputs. Maybe that isn't separable from Context?

armory/instrument/instrument.py
davidslater: Yeah, probably good to remove.

davidslater:
> Overall, this is nicely heading where you want it to be.
> 
> If I read this correctly, you would want a config-file representation for the add_meter calls. I'll be thinking of a tiga compliant way for that.

Yes, exactly. For starters, I would probably make it to parity with the current config, then do the translation to tiga after that.

armory/instrument/instrument.py
davidslater: Removed.

armory/instrument/instrument.py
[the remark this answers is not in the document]
davidslater: I've cleaned up the ProbeMapper class a bit.

armory/instrument/instrument.py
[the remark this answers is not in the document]
davidslater: Moved `get_meters` to Context, and it just returns `self.meters`

lcadalzo:
Overall this is nicely demonstrating the desired functionality although I am finding the code difficult to grok. I'll be adding more specific comments

armory/instrument/instrument.py
```diff
@@ -0,0 +1,670 @@
...
+        "postprocessed_l2_distance",
+        metrics.l2,
+        "model.prep_output[benign]",
+        "model.prep_output[adversarial]",
+    )
+    add_meter(  # TODO: enable adding context for iteration number of attack
+        "sum of x_adv", np.sum, "attack.x_adv",  # could also do "attack.x_adv[attack]"
+    )
+    add_meter(
+        "categorical_accuracy",
```
lcadalzo, line 623: In general, the passing of values to `add_meter()` via strings with `.` in the middle I'm finding a bit non-intuitive. 

E.g. where `"scenario.y"` is being passed, I originally assumed this was somehow using the attribute of a scenario object, but only later realized the `scenario` comes from the probe name and `y` the kwarg passed to `update()`. Perhaps this is a non-issue, but it initially confused me

armory/instrument/instrument.py
```diff
@@ -0,0 +1,670 @@
...
+    # Begin metric setup (could happen anywhere)
+
+    # from armory.instrument import add_writer, add_meter
+    from armory.utils import metrics
+
+    add_writer(PrintWriter())
+    add_meter(
+        "postprocessed_l2_distance",
+        metrics.l2,
+        "model.prep_output[benign]",
```
lcadalzo, line 616: similar to [this](https://github.com/twosixlabs/armory/pull/1387/files#r851465631) comment, the brackets-in-a-string notation I found a bit non-intuitive. It's not immediately clear to me what the possible values that could go inside the brackets are. The different stages?

Relatedly, the `process_meter_arg()` function feels potentially brittle. If I call

```
 add_meter(
        "postprocessed_l2_distance",
        metrics.l2,
        "model.prep_output[[benign]",
        "model.prep_output[adversarial]",
    )
```

I get a not-so-clear error:

```
 File "/workspace/armory/instrument/instrument.py", line 617, in main
    "model.prep_output[adversarial]",
  File "/workspace/armory/instrument/instrument.py", line 570, in add_meter
    connect_meter(meter)
  File "/workspace/armory/instrument/instrument.py", line 565, in connect_meter
    get_context().connect_meter(meter)
  File "/workspace/armory/instrument/instrument.py", line 267, in connect_meter
    self.mapper.connect_meter(meter)
  File "/workspace/armory/instrument/instrument.py", line 187, in connect_meter
    probe_variable, stage_filter = process_meter_arg(arg)
  File "/workspace/armory/instrument/instrument.py", line 148, in process_meter_arg
    arg, filt = arg.split("[")
ValueError: too many values to unpack (expected 2)
```

armory/instrument/instrument.py
```diff
@@ -0,0 +1,670 @@
...
+        self.stage = stage
+
+    def set_batch(self, batch: int):
+        # NOTE: batch could be a set of sample indices
+        self.batch = batch
+
+    def increment_batch(self):
+        self.batch += 1
+
+    def is_measuring(self, probe_variable):
```
lcadalzo, line 252: I understand this may be by design, but I could imagine the following kind of thing causing confusion in the future:

if there's a misalignment between the string I pass to `add_meter()` and the `kwarg` passed to `probe.update()` nothing will happen. E.g. if I make a typo and pass

```
add_meter(
        "categorical_accuracy",
        metrics.categorical_accuracy,
        "scenaro.y",
        "scenario.y_pred",
```
categorical accuracy won't be measured, nor will I see any error or warning (since `is_measuring()` will return `False`). I'd need to think more about a way to catch these kinds of issues without violating design decisions about how a probe/meter/etc will work

armory/instrument/instrument.py
```diff
@@ -0,0 +1,670 @@
...
+        result = self.final(self.results, **self.final_kwargs)
+        record = (self.final_name, None, result)
+        self.final_result = result
+        for writer in self.writers:
+            writer.write(record)
+
+    def results(self):
+        if not self.keep_results:
+            raise ValueError("keep_results is False")
+        return self._results
```
lcadalzo, line 446: should `self._results` be list of 3-tuples instead of a flat list? Not sure what useful things we could do with the following format I'm seeing:

```
Pdb) self._results
	['sum of x_adv', 0, 52.496500501843975, 'sum of x_adv', 0, 57.73304135870708]
```

docs/metrics.md
```diff
@@ -46,4 +46,136 @@ metrics, to prevent expanding the required set of dependencies. Please see [armo
...
+This involves specifying a function or sequence of functions to perform when a value is updated.
+However, this is done lazily - it is only performed if a meter is connected and is measuring.
+
+For instance, to extract a pytorch tensor `layer_3_output`, you could do:
+```
+probe.update(lambda x: x.detach().cpu().numpy(), a=layer_3_output)
+```
+Or in a longer form:
+```
+probe.update(lambda x: x.detach(), lambda x: x.cpu().numpy(), a=layer_3_output)
```
lcadalzo, line 87: one thing that helped my understanding but wasn't immediately clear is that when you call

```
probe.update(arbitrary_name=value)
```

you are defining `arbitrary_name` as a name that can then be used when passing a string `"probe_name.arbitrary_name"` to `add_meter()`

armory/instrument/instrument.py
```diff
@@ -0,0 +1,670 @@
...
+        stage_filter = filt[:-1].strip()
+        # tokens = [x.strip() for x in filt.split(",")]
+    else:
+        stage_filter = None
+    probe_variable = arg
+
+    return probe_variable, stage_filter
+
+
+class ProbeMapper:
```
lcadalzo, line 160: I want to make sure I understand how all the pieces (probe, meter, context, probemapper, writer) fit together. A context has a bunch of meters. Each meter corresponds to a metric you want to measure. A meter uses values stored via `probe.update()` calls to compute a metric

I don't quite see how/where a `ProbeMapper` fits in. Perhaps in other words, what is it doing that the context object do can't do? The `map_to_meters_args()` method and corresponding nested dictionary `self.probe_filter_meter_arg` I'm finding a little difficult to wrap my head around

armory/instrument/instrument.py
```diff
@@ -0,0 +1,670 @@
...
+        probe.set_sink(get_context())
+        _PROBES[name] = probe
+    return _PROBES[name]
+
+
+def connect_meter(meter):
+    get_context().connect_meter(meter)
+
+
+def add_meter(*args, **kwargs):
```
lcadalzo, line 568: If meters are tied a context, should `add_meter()` instead be a context method? Or should I be thinking of things differently

armory/instrument/instrument.py
```diff
@@ -0,0 +1,670 @@
...
+    def add_writer(self, writer):
+        """
+        Convenience method to add writer to all meters in this context
+        """
+        if writer in self.writers:
+            log.warning(f"writer {writer} already connected to {self.name} context")
+            return
+
+        self.writers.append(writer)
+        for meter in self.meters:
```
lcadalzo, line 283: So this current approach adds the writer to _all_ meters. What if, for example, one thing I want to do is write a JPG file of an image after going through a model's preprocessing ops. I imagine I'd need to create a `Writer` class that does what I want with the image. If this is the case, I wouldn't want this writer to apply to other meters. Should a writer thus be tied to a specific meter then?

armory/instrument/instrument.py
```diff
@@ -0,0 +1,670 @@
...
+    def get_arg_names(self):
+        return list(self.arg_index)
+
+    def clear(self):
+        """
+        Clear all of the current values
+        """
+        self.values = [None] * len(self.arg_index)
+        self.values_set = [False] * len(self.arg_index)
+        self.batches = [None] * len(self.arg_index)
```
lcadalzo, line 375: the `self.batches` attribute caused me some confusion. If I understand, it's a list of length equal to the number of args. And each element is the batch_index of where the arg came from. So all elements should be the same (i.e. come from same batch). If that's all correct, maybe something like `self.arg_batch_indices` would be more descriptive

armory/instrument/instrument.py
```diff
@@ -0,0 +1,670 @@
...
+        """
+        Return a list of (meter, arg) that are using the current probe_variable
+        """
+        filter_map = self.probe_filter_meter_arg.get(probe_variable, {})
+        meters = filter_map.get(stage, [])
+        meters.extend(filter_map.get(None, []))  # no stage filter (default)
+        return meters
+
+
+class Context:  # NOTE: may need to rename to something like Experiment or Procedure
```
lcadalzo, line 232: I think I agree that `Context` could use a rename, but am struggling to think of what would be best. `MetricRecorder`? Something that better captures the holistic nature of the class

armory/instrument/instrument.py
```diff
@@ -0,0 +1,670 @@
...
+def get_probe(name: str = ""):
+    """
+    Get a probe with specified name, creating it if needed
+    """
+    if name != "" and not str.isidentifier(name):
+        raise ValueError(f"name {name} should be an identifier or the empty string")
+
+    if name not in _PROBES:
+        probe = Probe(name)
+        probe.set_sink(get_context())
```
lcadalzo, line 559: do sink and context need a different name if sink is just set to the context?

armory/instrument/instrument.py
davidslater: I structured this so that if you wanted to, you could use the python namespace for this and ensure no collisions. E.g., `get_probe(__name__)`. Since However, if I want to guarantee no name collisions, I need to check for `'.'` in update. I will add that now, but we can discuss the broader naming convention later.

armory/instrument/instrument.py
davidslater: The latter was an actual error in my input checking ("and" instead of "or"), so that now the error is:
```
>>> instrument.process_meter_arg("model.prep_output[[benign]")
  File "/Users/david.slater/git/davidslater/armory/armory/instrument/instrument.py", line 160, in process_meter_arg
    raise ValueError(f"arg '{arg}' must have a single matching '[]' pair or none")
ValueError: arg 'model.prep_output[[benign]' must have a single matching '[]' pair or none
```

davidslater: Currently, it is just different stages that go inside the brackets. There is a larger question of how to enable more expressive filters.

armory/instrument/instrument.py
davidslater: I like the idea of adding a warning, though not an error, if a value name is added to a meter but never captured. I think it's nice to enable conditional measurement (e.g., measuring targeted attacks but only when targeted attacks happen in the code) without erroring out.

 One challenge is that probes may not be connected when meters are added (and vice versa). In fact, a probe update that is hooked in model code will only happen during the forward pass.

I think a fix for this would be to raise a warning when all the meters are finalized if a value has never been read.

davidslater: Now the output will do something like:
```
2022-04-15 21:47:35  4s WARNING  __main__:finalize:452 Meter 'categorical_accuracy' was never measured. The following args were never set: ['scenaro.y']
2022-04-15 21:47:35  4s WARNING  __main__:finalize:452 Meter 'targeted_categorical_accuracy' was never measured. The following args were never set: ['scenario.y_target']
```

armory/instrument/instrument.py
davidslater: Not sure what is most helpful. I meant this to be a list of the third value (the "result"), not a list of the records.
```
[51.74264097777119, 56.87368698209825, 61.65885794852083, ...
```

davidslater: Fixed.

docs/metrics.md
davidslater: Added in docs. Need to update them further once done.

armory/instrument/instrument.py
davidslater: `ProbeMapper` is meant to store mappings from probe outputs to meter inputs. It is not envisioned to be used outside of Context (since context args like "stage" must be passed in). However, since it is essentially a data structure, I separated it into its own class for easier testing.

armory/instrument/instrument.py
davidslater: `add_meter` was meant as a syntactic gloss, a wrapper for creating a Meter and connecting it to the main context object.

armory/instrument/instrument.py
davidslater: Yes, you would want to create a `Writer` and apply it to that specific Meter. Need to think about how to have optional pieces here.

armory/instrument/instrument.py
davidslater: Yes, that would be more informative.

armory/instrument/instrument.py
davidslater: More discussion later.

armory/instrument/instrument.py
davidslater: I need to think more about this - in theory, you could send probes to anything that used the correct interface, not just the context. You could send it directly to an image exporter, to use your previous example.

armory/instrument/instrument.py
lcadalzo: I agree a warning makes the most sense. The new output looks good

armory/instrument/instrument.py
lcadalzo: Got it, wasn't sure if there was a reason you were storing the strings

armory/instrument/config.py
```diff
@@ -0,0 +1,313 @@
...
+
+def _task_metric(name, metric_kwargs, use_mean=True, include_target=True):
+    metric = metrics.get_supported_metric(name)
+    final_kwargs = {}
+    if name in MEAN_AP_METRICS:
+        final_suffix = name
+        final = MeanAP(metric)
+        final_kwargs = metric_kwargs
+
+        name = "identity"
```
lcadalzo, line 254: Why is `name` being set to `"identity"`?

davidslater: Yeah, this should be something to indicate it's an input. Changed.

lcadalzo:
The results JSON is now storing predictions, it appears, and for object detection (and perhaps other scenarios?) I think this is far too verbose. For example, running a CARLA OD config with `--num-eval-batches 10` yields a JSON that is 3739 lines long, about 90% of which is the boxes. This only grows as `--num-eval-batches` does, too

davidslater:
Interesting - can you drop a results config in Slack?

armory/scenarios/poison.py
```diff
@@ -192,7 +196,23 @@ def poison_dataset(self):
...
+        self.n_poisoned = int(len(self.poison_index))
+        self.n_clean = (
+            len(self.y_clean) - self.n_poisoned
+        )  # self.y_clean is the whole pre-poison train set
+        poisoned = np.zeros_like(self.y_clean, np.bool)
+        poisoned[self.poison_index.astype(np.int64)] = True
+        self.probe.update(poisoned=poisoned)
+        self.hub.record("N_poisoned_train_samples", self.n_poisoned)
+        self.hub.record("N_clean_train_samples", self.n_clean)
+        self.train_set_class_labels = sorted(np.unique(self.y_clean))
```
swsuggs, line 208: I think 208-212 belong better in load_train_dataset()

lcadalzo:
I think there should be means of disabling the recording of each metric per sample (I'm of the opinion that it should be off by default)

armory/scenarios/poisoning_witches_brew.py
```diff
@@ -340,12 +342,49 @@ def load_dataset(self, eval_split_default="test"):
...
+                final=np.mean,
+                final_name="mean_accuracy_non_trigger_images",
+                record_final_only=False,
+            )
+        )
+        self.hub.connect_meter(
+            Meter(
+                "accuracy_trigger_images",
+                metrics.get_supported_metric("categorical_accuracy"),
+                "scenario.y[adversarial]",
```
swsuggs, line 360: Can we pick any name for the context/stage?  I think trigger/non-trigger is better than adversarial/benign, because it's ambiguous whether adversarial refers to the poisoned training images or the unpoisoned trigger images

davidslater: Yes, we can.

armory/instrument/instrument.py
```diff
@@ -0,0 +1,704 @@
...
+            if not keep_results:
+                log.warning("final is not None. Overriding keep_results to True")
+                keep_results = True
+        self.final = final
+        self.final_name = final_name
+        self.final_kwargs = final_kwargs or {}
+        self.final_result = None
+        if not isinstance(self.final_kwargs, dict):
+            raise ValueError(f"final_kwargs must be None or a dict, not {final_kwargs}")
+        self.keep_results = keep_results
```
lcadalzo, line 452: what is the purpose of the `keep_results` attribute? Is it to be set `False` when you have a metric that you want to log but not store in the output JSON?

lcadalzo: I see the docstring says `"whether to locally store results"` but I'm moreso trying to understand when we would want this to be `False`

davidslater: keep_results is whether to keep the results from each input batch in the class or just to push it to the writer. This is necessary if you are doing a "mean" or other final metric, but is otherwise unneeded. The main reason I put this in was in case the intermediate results got overly large.

However, maybe I should just remove this and always store them?

armory/instrument/config.py
```diff
@@ -0,0 +1,428 @@
...
+    """
+
+    def __init__(
+        self,
+        task=None,
+        task_kwargs=None,
+        perturbation=None,
+        means=True,
+        include_benign=True,
+        include_adversarial=True,
```
lcadalzo, line 29: need to dig further, but I'm noticing that `include_adversarial` is `True` even when I set `--skip-attack`

lcadalzo: `include_targeted` is also `True` even though I'm not doing a targeted attack. Again, not sure yet what implications this has if any

davidslater: I think those will still run fine and not output anything.

lcadalzo: are we expecting users to potentially set `include_benign` or `include_adversarial` to `False` in the config file (as this is the only way to override the default `True`)?

davidslater: I switched this back to the prior way, where the scenario flags would get passed in at construction.

davidslater: Just pushed an update.

armory/instrument/config.py
```diff
@@ -0,0 +1,454 @@
...
+                self.wrt = "benign predictions as"
+            else:
+                self.wrt = "ground truth"
+            self.task_type = "adversarial"
+        else:
+            self.wrt = "ground truth"
+            self.task_type = "benign"
+
+    def _write(self, name, batch, result):
+        # TODO: once metrics have also been updated, rewrite to be less error prone
```
lcadalzo, line 217: do you intend to address this TODO before merge?

davidslater: no, this is meant to be addressed after updating `armory.utils.metrics` in a future PR (post 0.15.0)

armory/instrument/instrument.py
davidslater: `keep_results` removed in latest push

armory/scenarios/poison.py
davidslater: The reason they are here is that the metrics are not set up until after `load_train_dataset` is called, do the `hub.record` call would fall on the floor.

armory/scenarios/poisoning_witches_brew.py
davidslater: I updated those stages in the latest push.

davidslater: They are now "trigger"/"non-trigger"

armory/scenarios/scenario.py
```diff
@@ -254,11 +253,15 @@ def evaluate_all(self):
         for _ in tqdm(range(len(self.test_dataset)), desc="Evaluation"):
             self.next()
             self.evaluate_current()
+        self.hub.set_context(stage="finished")
 
     def next(self):
+        self.hub.set_context(stage="next")
```
lcadalzo, line 259: for my knowledge/curiosity: what's the purpose behind setting stage to `"next"` as well as `"finished"`?

armory/scenarios/scenario.py
```diff
@@ -378,6 +377,8 @@ def evaluate(self):
             log.warning(f"{self._evaluate} returned None, not a dict")
         output = self._prepare_results(self.config, results)
         self._save(output)
+        log.info("Clearing global instrumentation variables")
```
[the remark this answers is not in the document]
davidslater: You're right. Probably something better to demote to debug or trace level logging.

armory/scenarios/scenario.py
davidslater: Mainly so that we don't accidentally tag things outside of `run_attack` or `run_benign` as `benign` or `attack` or `adversarial`.

armory/scenarios/scenario.py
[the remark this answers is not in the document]
davidslater: Updated (and moved before the save call).

lcadalzo:
Would it be a good example test case of this new feature to try and do what #1285 describes using probe/meter/etc or is that outside the scope?

davidslater:
> Would it be a good example test case of this new feature to try and do what #1285 describes using probe/meter/etc or is that outside the scope?

I think that would be a good use case, and that is what I was planning to do (albeit after this release). I think it may require adding new functionality, depending on what ExportSamples looks like. We can discuss if we have time before release.

armory/instrument/instrument.py
```diff
@@ -0,0 +1,690 @@
...
+        probe.update(lambda x: x.detach().cpu().numpy(), x_post=x_post)
+
+        # outside of model code
+        probe.hook(model, lambda x: x.detach().cpu().numpy(), x_post=x_post)
+
+        # elsewhere (could be reasonably defined in a config file as well)
+        from armory import instrument
+        from armory import metrics
+        meter = instrument.MetricMeter("l2_dist_postprocess", metrics.L2, "model.x_post[benign]", "model.x_post[adversarial]")
+        instrument.add_meter(meter)
```
lcadalzo, line 21: not sure if these comments are meant to make it into the merge, but if so, the instrument module no longer has a `add_meter` function

lcadalzo:
When running `scenario_configs/carla_obj_det_dpatch_defended.json`, I'm now seeing nan's in the logged output:

```
2022-04-26 21:06:58 54s SUCCESS  armory.instrument.config:_write:246 adversarial_mean_carla_od_disappearance_rate_wrt_benign_preds on adversarial examples w.r.t. benign predictions as labels: nan%
2022-04-26 21:06:58 54s SUCCESS  armory.instrument.config:_write:246 adversarial_mean_carla_od_hallucinations_per_image_wrt_benign_preds on adversarial examples w.r.t. benign predictions as labels: 0.0
2022-04-26 21:06:58 54s SUCCESS  armory.instrument.config:_write:246 adversarial_mean_carla_od_misclassification_rate_wrt_benign_preds on adversarial examples w.r.t. benign predictions as labels: nan%
2022-04-26 21:06:58 54s SUCCESS  armory.instrument.config:_write:246 adversarial_mean_carla_od_true_positive_rate_wrt_benign_preds on adversarial examples w.r.t. benign predictions as labels: nan%

```

armory/instrument/config.py
```diff
@@ -0,0 +1,454 @@
...
+
+
+def _task_metric(
+    name,
+    metric_kwargs,
+    use_mean=True,
+    include_benign=True,
+    include_adversarial=True,
+    include_targeted=True,
+    record_final_only=True,
```
lcadalzo, line 258: should `record_final_only` be passed as a kwarg to this function if the variable is being reset based on the metric name?

armory/instrument/config.py
```diff
@@ -0,0 +1,454 @@
...
+                final_kwargs=final_kwargs,
+                record_final_only=record_final_only,
+            )
+        )
+    else:
+        meters.append(None)
+    return meters
+
+
+def task_metrics(
```
lcadalzo, line 340: I think the code would be a bit more readable if `task_metrics()` was renamed to more concretely state what the function is doing

armory/instrument/instrument.py
```diff
@@ -0,0 +1,690 @@
...
+        else:
+            raise ValueError(f"mode {mode} not in ('pytorch', 'tf')")
+
+
+class MockSink:
+    """
+    Measures all probe inputs and prints to screen
+    """
+
+    def is_measuring(self, probe_variable):
```
lcadalzo, line 163: not sure if `MockSink` was meant to be temporary, but in order to get it working in an example, I had to make both `is_measuring()` and `update()` static methods and remove the `self` kwarg. Prior to this change, I would get a missing positional arg error

docs/metrics.md
```diff
@@ -46,4 +58,213 @@ metrics, to prevent expanding the required set of dependencies. Please see [armo
...
+This will publish the given value(s) to the given name(s) (also called probe variables) in the probe namespace of the connected Hub.
+To be more concrete:
+```
+probe = get_probe("my.probe_name")
+probe.update(arbitrary_variable_name=15)
+```
+will push the value 15 to `"my.probe_name.arbitrary_variable_name"`.
+These names will be used when instantiating `Meter` objects.
+
+However, this will fall on the floor (`del`, effectively) unless a meter is connected to the Hub to record values.
```
lcadalzo, line 104: I think it would be most clear/effective if immediately after we say "this will fall on the floor unless you do XYZ", we show an example of exactly how to do XYZ

Reading these docs, it seems very clear to me how to set up a probe but it's less obvious how to actually do anything with the value I'm storing. Perhaps even just rearranging the order of different pieces of the docs would make this more clear

docs/metrics.md
```diff
@@ -46,4 +58,213 @@ metrics, to prevent expanding the required set of dependencies. Please see [armo
...
+### Meter
+
+A Meter is used to measure values output by probes.
+It is essentially a wrapper around the functions of `armory.utils.metrics`.
+
+To instantiate a Meter:
+```
+from armory.instrument import Meter
+meter = Meter( 
+        self,
```
lcadalzo, line 166: do you mean to be passing in `self` to the instantiation?

docs/metrics.md
```diff
@@ -46,4 +58,213 @@ metrics, to prevent expanding the required set of dependencies. Please see [armo
...
+        keep_results - whether to locally store results
+            if final is not None, keep_results is set to True
+
+
+```
+To set up a meter, you need to connect it to the probe.
+This can be done as follows:
+```
+from armory.metrics import instrument
+meter = instrument.LogMeter()
```
lcadalzo, line 207: it seems this example is outdated as the `LogMeter` class no longer exists

armory/instrument/instrument.py
```diff
@@ -0,0 +1,690 @@
...
+            writer.close()
+
+        self.closed = True
+
+
+class Meter:
+    def __init__(
+        self,
+        name,
+        metric,
```
lcadalzo, line 383: could it make sense to allow metric to be `None`, in which case it just records whatever value you're measuring (in other words, basically an identity function). I'm again imagining the use case, which I think would be common, of users just wanting to store an intermediate model output

armory/instrument/instrument.py
```diff
@@ -0,0 +1,693 @@
...
+
+    def finalize(self):
+        """
+        Primarily intended for metrics of metrics, like mean or stdev of results
+        """
+        if self.never_measured:
+            unset = [arg for arg, i in self.arg_index.items() if not self.values_set[i]]
+            log.warning(
+                f"Meter '{self.name}' was never measured. "
+                f"The following args were never set: {unset}"
```
lcadalzo, line 516: so we have a warning if a meter is never measured, but would it also make sense to have a warning if a probe value "falls on the floor" i.e. is never used to measure anything / isn't connected to a meter?

armory/instrument/instrument.py
lcadalzo: perhaps I could use achieve this by passing in a simple metric function that merely returns the input, but that feels a little hacky

armory/instrument/instrument.py
```diff
@@ -0,0 +1,693 @@
...
+        self.never_measured = False
+
+    def finalize(self):
+        """
+        Primarily intended for metrics of metrics, like mean or stdev of results
+        """
+        if self.never_measured:
+            unset = [arg for arg, i in self.arg_index.items() if not self.values_set[i]]
+            log.warning(
+                f"Meter '{self.name}' was never measured. "
```
lcadalzo, line 515: Need to look more closely at how/where the flag is propagated w.r.t. when/where the Meters are created for a scenario but: it seems odd that if you set `--skip-attack` you get a bunch of warnings saying that `Meter adversarial_*` was never measured. If `--skip-attack` is set, I'd think those meters shouldn't be created in the first place

armory/scenarios/poison.py
swsuggs: Ok makes sense.  Then all of the new additions to this function (`poison_dataset`) need to be copied to or otherwise inherited by poisoning_witches_brew.py, which overrides that function.  Currently, running witches_brew raises an `attribute error: no attribute self.train_set_class_labels`

armory/scenarios/poisoning_witches_brew.py
```diff
@@ -340,12 +342,51 @@ def load_dataset(self, eval_split_default="test"):
...
+        # log.info(
+        #     f"Accuracy on trigger images: {self.trigger_accuracy_metric.mean():.2%}"
+        # )
 
-        self.benign_test_accuracy_per_class = (
-            {}
-        )  # store accuracy results for each class
         if self.config["adhoc"].get("compute_fairness_metrics", False):
             self.fairness_metrics = FairnessMetrics(
                 self.config["adhoc"], self.use_filtering_defense, self
```
swsuggs, line 392: This might be missing a couple things, just comparing to load_metrics() in poison.py:
- add the filter perplexity metric (if self.use_filtering_defense): `self.fairness_metrics.add_filter_perplexity()`
- set up the results writer: 
```
self.results_writer = ResultsWriter(sink=None)
self.hub.connect_writer(self.results_writer, default=True)
```

armory/scenarios/poison.py
```diff
@@ -283,33 +321,112 @@ def load_dataset(self, eval_split_default="test"):
...
+                record_final_only=True,
+            )
+        )
+
         if self.config["adhoc"].get("compute_fairness_metrics", False):
             self.fairness_metrics = FairnessMetrics(
                 self.config["adhoc"], self.use_filtering_defense, self
             )
+            if not self.check_run:
+                self.fairness_metrics.add_filter_perplexity()
```
swsuggs, line 421: Adding filter_perplexity should be conditional on self.use_filtering_defense

armory/instrument/config.py
davidslater: It's only being overridden by the mAP metrics, which do not have a reasonable intermediate value to record. I think in the future we'll want to subclass Meter to have specific modes for Global and per-Sample metrics, but I don't think we're quite there yet.

In some ways, the `armory/instrument/config.py` file is a bridge from the old MetricsLogger to the new framework, but I think we'll want to revise it once we update `armory/utils/metrics.py`.

armory/instrument/config.py
davidslater: Renamed, albeit verbosely.

armory/instrument/instrument.py
lcadalzo: Update: I would propose, in the `add_tasks_wrt_benign_predictions()` method, changing

```
if self.task is not None:
```

to 

```
if self.task is not None and self.include_adversarial:
```

This way, we don't create meters that we know will never be used. The lone exception is the perturbation meter, for which I still see a warning that it's never used

lcadalzo: similarly, in the `__init__` of the `MetricsLogger` class, you could change

```
if perturbation is not None:
```

to 

```
if perturbation is not None and self.include_adversarial:
```

to avoid creating a perturbation meter that we know won't be used (and thus get rid of the warning)

docs/metrics.md
davidslater: I added a Quick Start example to the previous section and clarified this line.

armory/scenarios/poisoning_witches_brew.py
```diff
@@ -340,12 +342,51 @@ def load_dataset(self, eval_split_default="test"):
...
+        )
+
+        per_class_accuracy = metrics.get_supported_metric("per_class_accuracy")
+        self.hub.connect_meter(
+            Meter(
+                "sample_non_trigger_test_accuracy_per_class",
+                metrics.get_supported_metric("identity_unzip"),
+                "scenario.y[non-trigger]",
+                "scenario.y_pred[non-trigger]",
+                final=lambda x: per_class_accuracy(*metrics.identity_zip(x)),
```
swsuggs, line 375: This one is still outputting per-sample accuracy.  Should it include an np.mean in the lambda function?

armory/instrument/instrument.py
```diff
@@ -0,0 +1,690 @@
...
+        auto_measure=True,
+        final=None,
+        final_name=None,
+        final_kwargs=None,
+        record_final_only=False,
+    ):
+        """
+        StandardMeter(metrics.l2, "model.x_post[benign]", "model.x_post[adversarial]")
+
+        probe_name.probe_variable[stage]
```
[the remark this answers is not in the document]
davidslater: Updated.

armory/instrument/instrument.py
davidslater: The main issue is when you have multiple values and multiple batches - you need to do some weird zipping and unzipping to make this work properly.

armory/instrument/instrument.py
davidslater: If no sink is connected to the probe, then it logs a warning. I can think of lots of valid reasons a probe value would not be measured, however, due to filtering by the meters.

armory/instrument/instrument.py
davidslater: Updated.

armory/scenarios/poisoning_witches_brew.py
davidslater: Good catch!

armory/scenarios/poison.py
davidslater: Done.

armory/scenarios/poisoning_witches_brew.py
davidslater: It was a problem with my per_class_accuracy metric, not the meter call. Fixed.

armory/scenarios/poisoning_witches_brew.py
swsuggs: Load_metrics is also missing the tpr/fpr metric:
```
        if self.use_filtering_defense:
            # Filtering metrics
            self.hub.connect_meter(
                Meter(
                    "filter",
                    metrics.get_supported_metric("tpr_fpr"),
                    "scenario.poisoned",
                    "scenario.removed",
                )
            )
```

armory/scenarios/poison.py
swsuggs: 199-207 should be copied to witches_brew also, sorry if that wasn't clear:
```
        self.n_poisoned = int(len(self.poison_index))
        self.n_clean = (
            len(self.y_clean) - self.n_poisoned
        )  # self.y_clean is the whole pre-poison train set
        poisoned = np.zeros_like(self.y_clean, np.bool)
        poisoned[self.poison_index.astype(np.int64)] = True
        self.probe.update(poisoned=poisoned)
        self.hub.record("N_poisoned_train_samples", self.n_poisoned)
        self.hub.record("N_clean_train_samples", self.n_clean)
```

armory/utils/poisoning.py
```diff
@@ -313,34 +319,19 @@ def add_cluster_metrics(
...
         for class_id in test_set_class_labels:
-            self.majority_x_class_prediction_chi2_metrics[class_id].add_results(
-                majority_x_correct_prediction_tables[class_id]
-            )
-            self.majority_x_class_prediction_spd_metrics[class_id].add_results(
-                majority_x_correct_prediction_tables[class_id]
+            majority_x = majority_x_correct_prediction_tables[class_id]
+            chi2 = np.mean(chi2_metric(majority_x))
+            spd = np.mean(spd_metric(majority_x))
+            self.hub.record(f"model_bias_chi^2_p_value_{str(class_id).zfill(2)}", chi2)
```
swsuggs, line 328: AttributeError: 'FairnessMetrics' object has no attribute 'hub'

swsuggs:
Could you merge develop again so I can test poisoning with the cifar stuff that was added this morning?

armory/scenarios/poison.py
```diff
@@ -283,33 +321,114 @@ def load_dataset(self, eval_split_default="test"):
...
+                    "scenario.poisoned",
+                    "scenario.removed",
+                )
+            )
+
+        self.hub.connect_meter(
+            Meter(
+                "sample_accuracy_on_benign_test_data_all_classes",
+                get_supported_metric("categorical_accuracy"),
+                "scenario.y[benign]",
```
swsuggs, line 342: This one didn't work: `Meter 'sample_accuracy_on_benign_test_data_all_classes' was never measured. The following args were never set: ['scenario.y[benign]']`  

(`sample_accuracy_on_poisoned_data_all_classes` had the same error but I think will have the same fix)

armory/utils/poisoning.py
davidslater: Fixed

armory/scenarios/poison.py
davidslater: I removed the stage filters, as they are unnecesary, which should fix the problem.

davidslater:
@swsuggs I merged with develop. Pushing.

davidslater:
Finished docs.

Issue #1379, opened 2022-03-25
Incorporate new metrics for poisoning - Eval 5


davidslater:
Needs more info

Nathalie-B:
@davidslater
Metrics should include: 
- In the fairness section: 
   1. Add samples per class
   2. Accuracy per class
- For filtering defenses: 
   1. Percentage of filtered data by the defense 
   2. Number of samples filtered 
   2. Percentage of filtered data that was not poison

Nathalie-B:
P.S. this is in addition to all the dev. metrics already implemented for Eval 4

swsuggs:
Completed in #1393

Issue #1355, opened 2022-03-23
Write additional info to json results for poisoning


swsuggs:
The poisoning team would like the following info to be added to the results json file:
- percent/number of datapoints filtered by a filtering defense (currently logged but not written to results)
- number of test samples per class, and number correctly classified per class

lcadalzo:
@swsuggs this can now be closed, correct?

swsuggs:
Completed in #1393

PR #1393, opened 2022-03-30
additional poison results output

swsuggs:
Writes additional info in results json for poisoning scenario, such as f1 score, true negative/positive rate, accuracy per class, etc.

Addresses #1379 and #1355

### DIFF
--- a/armory/scenarios/poison.py
+++ b/armory/scenarios/poison.py
@@ -124,10 +124,7 @@ class Poison(Scenario):
         )
 
     def set_dataset_kwargs(self):
-        self.dataset_kwargs = dict(
-            epochs=1,
-            shuffle_files=False,
-        )
+        self.dataset_kwargs = dict(epochs=1, shuffle_files=False)
 
     def load_train_dataset(self, train_split_default=None):
         """
@@ -305,9 +302,9 @@ class Poison(Scenario):
                 "categorical_accuracy"
             )
 
-        self.benign_test_accuracy_per_class = {
-            y: [] for y in np.unique(self.y_clean)
-        }  # store accuracy results for each class
+        self.benign_test_accuracy_per_class = (
+            {}
+        )  # store accuracy results for each class
         if self.config["adhoc"].get("compute_fairness_metrics", False):
             self.fairness_metrics = FairnessMetrics(
                 self.config["adhoc"], self.use_filtering_defense, self
@@ -345,6 +342,9 @@ class Poison(Scenario):
         self.source = source
 
         for y_, y_pred_ in zip(y, y_pred):
+            if y_ not in self.benign_test_accuracy_per_class.keys():
+                self.benign_test_accuracy_per_class[y_] = []
+
             self.benign_test_accuracy_per_class[y_].append(
                 y_ == np.argmax(y_pred_, axis=-1)
             )
@@ -403,6 +403,12 @@ class Poison(Scenario):
         )  # self.y_clean is the whole pre-poison train set
         self.results["N_poisoned_train_samples"] = n_poisoned
         self.results["N_clean_train_samples"] = n_clean
+        train_set_class_labels = sorted(list(np.unique(self.y_clean)))
+        test_set_class_labels = sorted(list(self.benign_test_accuracy_per_class.keys()))
+        if test_set_class_labels != train_set_class_labels:
+            log.warning(
+                "Test set contains a strict subset of train set classes.  Some metrics for missing classes may not be computed."
+            )
 
         if self.use_filtering_defense:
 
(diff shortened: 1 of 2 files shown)


armory/scenarios/poison.py
```diff
@@ -386,6 +390,46 @@ def finalize_results(self):
...
+            self.results["filter_false_positive_rate"] = false_positive_rate
+            self.results["filter_true_negative_rate"] = true_negative_rate
+            self.results["filter_false_negative_rate"] = false_negative_rate
+            self.results["filter_f1_score"] = f1_score
+
+            self.results["fraction_data_removed_by_filter"] = filtered.mean()
+            self.results["N_datapoints_removed_by_filter"] = int(filtered.sum())
+            
+
+        for y in np.unique(self.y_clean):
```
lcadalzo, line 428: instead of doing `np.unique(self.y_clean)` again, you could just use `self.class_accuracy_dict.keys()`, no? If `self.y_clean` ever gets really huge this could be a little bit faster

armory/scenarios/poison.py
```diff
@@ -386,6 +390,46 @@ def finalize_results(self):
...
+            self.results["filter_false_positives"] = false_positives
+            self.results["filter_true_negatives"] = true_negatives
+            self.results["filter_false_negatives"] = false_negatives
+            self.results["filter_true_positive_rate"] = true_positive_rate
+            self.results["filter_false_positive_rate"] = false_positive_rate
+            self.results["filter_true_negative_rate"] = true_negative_rate
+            self.results["filter_false_negative_rate"] = false_negative_rate
+            self.results["filter_f1_score"] = f1_score
+
+            self.results["fraction_data_removed_by_filter"] = filtered.mean()
```
lcadalzo, line 424: cosmetics, but I might change `"fraction_data_removed_by_filter"` to `"filter_fraction_data_removed"` and similarly for `"N_datapoints_removed_by_filter"` so that all `filter` related metrics start with `filter_` and thus appear side-by-side in the output JSON

armory/scenarios/poison.py
```diff
@@ -386,6 +390,46 @@ def finalize_results(self):
...
+            self.results["N_datapoints_removed_by_filter"] = int(filtered.sum())
+            
+
+        for y in np.unique(self.y_clean):
+            self.results[f"N_samples_in_class_{y}"] = int(np.sum(self.y_clean == y))
```
lcadalzo, line 429: for `"N_samples_in_class_y"`, I was thinking instead of `int(np.sum(self.y_clean == y))`, you could just do `len(self.class_accuracy_dict[y])`. 

However when I `breakpoint()` in the code to confirm this, the results aren't what I expect:

```
(Pdb) np.sum(self.y_clean == 0)
210

(Pdb) len(self.class_accuracy_dict[0])
60
```
After looking closer, I think it's because the former comes from the train set while the latter is from the test set? If this is correct, I think it'd be clearer to add `train` and `test` to more variable names and dict keys

Side note: when I was looking at the number of data points in `self.class_accuracy_dict` I noticed it was 12630:

```
>>> num_per_class = [len(self.class_accuracy_dict[k]) for k in self.class_accuracy_dict]
>>> sum(num_per_class)
12630
```

This isn't divisible by the batch_size (512), so is this just the total length of the test set and the last batch has fewer examples? Otherwise I'm not sure how this number is derived

lcadalzo:
Looks like this comment in `poison.py` should be deleted:

```
# TODO: measure TP and FP rates for filtering
```

armory/scenarios/poison.py
swsuggs: You are right about train/test data.  I'm clarifying the variable names.

You also seem to be right about the last batch having fewer samples.  When I print out the batch sizes, I'm seeing 24 batches of size 512 and one of size 342 for a total of 12630

armory/scenarios/poison.py
```diff
@@ -156,6 +156,7 @@ def load_train_dataset(self, train_split_default=None):
             **self.dataset_kwargs,
         )
         self.x_clean, self.y_clean = (np.concatenate(z, axis=0) for z in zip(*list(ds)))
+        self.class_accuracy_dict = {y:[] for y in np.unique(self.y_clean)} # store accuracy results per class
```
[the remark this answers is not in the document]
swsuggs: Agree, moved it.

armory/scenarios/poison.py
swsuggs: good idea

armory/scenarios/poison.py
```diff
@@ -302,7 +305,9 @@ def load_metrics(self):
                 "categorical_accuracy"
             )
 
-        self.benign_accuracy_per_class = {y:[] for y in np.unique(self.y_clean)} # store accuracy results for each class
+        self.benign_test_accuracy_per_class = {
```
lcadalzo, line 308: So there are two datasets in play here: a `train` and a `test`. The variable `self.y_clean` is really `self.y_clean_train`, so it seems like a potential issue to create the `self.benign_test_accuracy_per_class` keys from the `train` set. In the example I'm running, the `train` and `test` have exactly the same classes, but I imagine if they didn't (e..g if the test set only contains a subset of the train classes) then issues could arise. There's a similar part later on in `finalize_results()` that uses the `test` set keys to update both `test` and `train` statistics. It seems like there should be some further bifurcation of `test` and `train` variables to cover any case where the sets of labels aren't identical

swsuggs: Ok makes sense, new commit allows for that case

lcadalzo:
LGTM

PR #1398, opened 2022-04-11
1268 export bounding box annotations

### LINKED ISSUE #1268: Export bboxes as JSON
I noticed that there is an in-progress implementation of drawing bounding boxes on images in #1261. Would it be possible to support exporting the bounding box annotations as a JSON file? There is already a well-defined JSON results schema for COCO: https://cocodataset.org/#format-results. Using this schema will enable tools like [FiftyOne](https://voxel51.com/docs/fiftyone/) to load and annotate the images.

swsuggs:
Fixes #1268


docs/exporting_data.md
```diff
@@ -120,11 +119,16 @@ bounding boxes to the image. When this is set to `True`, you must provide values
 Traceback (most recent call last):
   File "<stdin>", line 1, in <module>
   File "/workspace/armory/utils/export.py", line 205, in get_sample
-    raise TypeError("Both y_i and y_pred are None, but with_boxes is True")
+    raise TypeError("Both y_i and y_i_pred are None, but with_boxes is True")
 TypeError: Both y_i and y_pred are None, but with_boxes is True
 
 ```
-If you'd only like to include ground-truth boxes (or only predicted boxes), don't provide an arg for `y_i` or `y_i_pred`.
+If you'd like to include only ground-truth boxes (or only predicted boxes), provide an arg for for only `y_i` (or only `y_i_pred`).
```
lcadalzo, line 126: minor typo here with `for for only`

armory/utils/export.py
```diff
@@ -242,26 +242,26 @@ def get_coco_formatted_bounding_box_data(
...
-        image_id = y_i["image_id"][0]  # All boxes in y_i are for the same image
-
-        if y_i is not None:
-            bboxes_true = y_i["boxes"]
-            labels_true = y_i["labels"]
+        if y_i is None:
+            # We can't export anything without y_i, because
+            # even exporting y_i_pred requires y_i['image_id'].
+            log.warning("Cannot export bounding box annotations due to missing image_id")
+            return ground_truth_boxes_coco_format, predicted_boxes_coco_format
```
lcadalzo, line 249: I think we could make `y_i` a required arg since it's necessary for this method to work (as opposed to having a warning and the function returning early)

docs/exporting_data.md
```diff
@@ -120,11 +119,16 @@ bounding boxes to the image. When this is set to `True`, you must provide values
...
   File "/workspace/armory/utils/export.py", line 205, in get_sample
-    raise TypeError("Both y_i and y_pred are None, but with_boxes is True")
+    raise TypeError("Both y_i and y_i_pred are None, but with_boxes is True")
 TypeError: Both y_i and y_pred are None, but with_boxes is True
 
 ```
-If you'd only like to include ground-truth boxes (or only predicted boxes), don't provide an arg for `y_i` or `y_i_pred`.
+If you'd like to include only ground-truth boxes (or only predicted boxes), provide an arg for only `y_i` (or only `y_i_pred`).
+
+For object detection, if `with_boxes` is `True`, the sample_exporter will also generate json files containing coco-formatted bounding box annotations.  These are saved in the scenario output directory as
```
lcadalzo, line 128: I would delete the `if with_boxes is True` in the docs since it might make the user think they have to go set another field/kwarg somewhere. The object detection scenario code ensures that `with_boxes` is `True`

lcadalzo: That's fair, given that the preceding instructions are for `--interactive` I think that makes sense.

R.e. your question: I think a user should be able to simply export the raw image and only pass `x` to the call. If we always export boxes, then `y` becomes required. We used to decide whether to draw boxes based on whether a `y` is provided, but it feels more transparent to have an explicit kwarg.

lcadalzo: One thing I noticed, and this is unrelated to this PR so we can address in a separate PR, but when you're manually calling `export()` in `--interactive` mode you can no longer provide a `with_boxes` kwarg (or any kwarg not listed [here](https://github.com/twosixlabs/armory/blob/develop/armory/utils/export.py#L23)) due to [this](https://github.com/twosixlabs/armory/pull/1297/commits/00f9b761e560b489b144b6fa16a1f6ace47b689a) commit. So the function call in the docs actually will error

lcadalzo:
@swsuggs in commit `bf225900` I addressed #1399. It will now override any default kwargs with the values explicitly passed to `export()` e.g. if you're calling this method in `interactive` mode

docs/exporting_data.md
swsuggs: So just to be clear, you feel ok about this as is?

swsuggs:
Thanks @lcadalzo, bf25900 looks good to me

docs/exporting_data.md
lcadalzo: maybe one minor suggestion: how would you feel about changing `if with_boxes is True, the sample_exporter will ...` to `if with_boxes=True is passed to the export() call, the sample_exporter will...`

swsuggs: see latest commit

Issue #1375, opened 2022-03-25
Update CARLA OD attack to support attacking depth (RGB-D) - Eval 5


davidslater:

yusong-tan:
Submitted PR #1400 and #1409 for this issue.

PR #1400, opened 2022-04-11
For Issue #1375 - Eval5 CARLA Object Detection AdversarialPatchPyTorch

yusong-tan:
Adding AdversarialPatchPyTorch with differentiable perspective transform and modifications to attack RGB (for single modality) and RGB+Depth (for multimodality)


lcadalzo:
Confirmed that on a single RTX 6000 GPU I'm seeing ~7.5 iters/s compared to ~3.3 using RobustDPatch. Before I do so myself, @yusong-tan have you done any benchmarking to confirm that adversarial results are similar between the two attacks?

yusong-tan:
@lcadalzo No, I have not done any metrics comparison other than speed. AdversarialPathPyTorch has differentiable perspective transform, and RobustDPatch doesn't so the number of attack iterations needed for comparison will be different.

lcadalzo:
Interestingly I'm seeing far lower adversarial mAP using `AdversarialPatchPyTorch` compared to `RobustDPatch` when testing the undefended RGB-only baseline model on the `"large"` split of the test set. As noted in the scenario docs, the RobustDPatch adversarial mAP and hallucinations/img for the undefended model are 0.17 and 10.3 respectively.

When I run the config in this PR, `carla_obj_det_adversarialpatch_undefended.json`, I see adversarial mAP and hallucinations/img of 0.05 and 31.6 respectively. My intuition would be that this is due to the fact that perspective transform is now differentiable

yusong-tan:
@lcadalzo Your observation is similar to mine. I suspect to get comparable results, Robust DPatch may need many more attack iterations.

When we have a PyTorch defended model, we may want to also attack it once with Robust DPatch to provide an idea of the degradation from using it.

lcadalzo:
There's a lot of attack code here and much of it seems to be copied from ART. If possible, I'm looking to see if there are ways to reduce the amount of copied code. Do you have thoughts in this direction?

lcadalzo:
One major example is the `_get_losses()` method that's added. It appears to be identical to ART except for [this](https://github.com/twosixlabs/armory/blob/5a3e67f486f8e6d36ee793d2119bd196d91c6562/armory/art_experimental/attacks/carla_obj_det_adversarial_patch.py#L103) line which allows a batch dim of 6. I wonder if this is something that could be modified in ART instead

lcadalzo:
And if we need to modify `_get_losses()`, I think it might be cleaner to do so where the baseline model is defined (as opposed to in the attack code) since this is a model-specific change. Patching the method in the attack code means that anyone who uses this attack will have their model updated with this method

yusong-tan:
@lcadalzo I prefer to make the one line `_get_losses()` change to be in ART, but I don't think we can get that change into the upcoming Armory release. If you think this is possible, we can reach out to Beat.

I agree that `_get_losses()` should be in the baseline model. It was an error on my part.

lcadalzo:
For my understanding: what's the motivation behind averaging the gradient over the depth channels [here](https://github.com/yusong-tan/armory/blob/eval5_carla_obj_det_adversarialpatchpytorch/armory/art_experimental/attacks/carla_obj_det_adversarial_patch.py#L191-L193)?

Additionally, why are the depth-channel gradients divided by `self.learning_rate * self.learning_rate_depth` [here](https://github.com/yusong-tan/armory/blob/eval5_carla_obj_det_adversarialpatchpytorch/armory/art_experimental/attacks/carla_obj_det_adversarial_patch.py#L198)?

lcadalzo:
> @lcadalzo I prefer to make the one line `_get_losses()` change to be in ART, but I don't think we can get that change into the upcoming Armory release. If you think this is possible, we can reach out to Beat.

I've reached out to Beat to see whether this is possible

armory/art_experimental/attacks/carla_obj_det_adversarial_patch.py
```diff
@@ -0,0 +1,598 @@
...
+                    max=self.estimator.clip_values[1],
+                )
+
+                if self._patch.shape[0] == 6:
+                    self._patch[3:, :, :] = torch.clamp(
+                        self._patch[3:, :, :], min=self.min_depth, max=self.max_depth
+                    )
+        else:
+            raise ValueError(
+                "Adam optimizer for CARLA Adversarial Patch not supported."
```
lcadalzo, line 214: is there meant to be modified code below the `ValueError` that won't be reached? If the only possible values for `self._optimizer_string` are `"pgd"` and `"Adam"`, then it seems like the following would make the most sense:

```
if "pgd":
    ...
elif "Adam":
    raise the ValueError not supported for CARLA
else:
    raise a ValueError for unrecognized _optimizer_string
```

armory/art_experimental/attacks/carla_obj_det_adversarial_patch.py
```diff
@@ -0,0 +1,598 @@
...
+        import torch  # lgtm [py/repeated-import]
+
+        self.estimator.model.zero_grad()
+        loss = self._loss(images, target, mask)
+        loss.backward(retain_graph=True)
+
+        if self._optimizer_string == "pgd":
+            patch_grads = self._patch.grad
+            if patch_grads.shape[0] == 6:
+                patch_grads[3:, :, :] = torch.mean(
```
lcadalzo, line 191: what's the motivation for averaging over the depth-channel gradients?

armory/art_experimental/attacks/carla_obj_det_adversarial_patch.py
```diff
@@ -0,0 +1,598 @@
...
+                    patch_grads[3:, :, :], dim=0, keepdim=True
+                )
+
+            gradients = patch_grads.sign() * self.learning_rate
+            if patch_grads.shape[0] == 6:
+                gradients[3:, :, :] = (
+                    gradients[3:, :, :] / self.learning_rate * self.learning_rate_depth
```
lcadalzo, line 198: why are depth-channel gradients divided by `self.learning_rate * self.learning_rate_depth`?

armory/art_experimental/attacks/carla_obj_det_adversarial_patch.py
yusong-tan: Depth only has one channel, but we repeat three times for it to work with the model on the forward pass. On the backward pass, the gradients are calculated for three depth channels, but we need to change them into three identical channels. To me, it makes more sense to average the gradients in the three depth channels than to randomly pick gradients for one of the channels.

armory/art_experimental/attacks/carla_obj_det_adversarial_patch.py
yusong-tan: Initially, when I used `self.learning_rate` for updating both RGB and depth, the Adam optimizer can be used. Later when I decided that the depth channel needed a separate learning rate, i.e., `self.learning_rate_depth`, the Adam optimizer could no longer accommodate two different learning rates. That's why I added the `raise ValueError()` but forgot to delete everything after it.

lcadalzo:
I've created the following issue in ART to support 6-channel inputs: https://github.com/Trusted-AI/adversarial-robustness-toolbox/issues/1630. Per Beat, this can go in ART 1.10.1, and I'm waiting to hear when that will be released

lcadalzo:
For the multimodal attack, no surprise but I'm seeing a much more successful attack on the "large" split of the test set: adversarial mAP of 0.08 and 42.6 hallucinations per image. The RobustDPatch RGB-only attack yielded an adversarial mAP of 0.42 and 1.7 hallucinations per image

PR #1403, opened 2022-04-12
Integrate cifar10 with poison scenario

### LINKED ISSUE #1372: Integration of CIFAR10 for poisoning for Eval 5
Moving from RESISC10 to CIFAR10

lcadalzo:
Fixes #1372. Successful integration means a config file we can run to completion and get sensible results


swsuggs:
You are right, to compute the fairness metrics for a model trained on cifar, we'll need a cifar explanatory model.  I'll create an issue for that and bring it up with the poisoning group at some point.

In the meantime, you can avoid that error by setting adhoc/compute_fairness_metrics to false.  I'm updating the error statement to be more clear.

swsuggs (approved):
I believe this is good to go

lcadalzo:
LGTM, able to run succesfully

PR #1404, opened 2022-04-13
Move apricot metadata to proper file

davidslater:
Currently, if you import `armory.utils.metrics`, it will end up importing `armory.data.adversarial_datasets` and `armory.data.datasets` in order to get a single constant, which actually takes some time and seems like an odd pattern.

Since we already have a metadata file for the other apricot constants, I am moving this constant to that file. Now `armory.utils.metrics` will only import that single metadata file, which is quite small.


mwartell (approved):
LGTM. Satisfies the intent of the PR.

Issue #1377, opened 2022-03-25
Integrate Witches Brew into Poisoning - Eval 5


davidslater:

Nathalie-B:
This requires adding functionality to load a surrogate model that will be used by the attack to create poison samples

davidslater:
(The surrogate model could be identical to the attacked model if it's differentiable, or it could be different. White-box vs transfer attack.)

PR #1406, opened 2022-04-13
WIP: Integrate witches brew

swsuggs:
#1390 
#1377 
Probably also #1372

### DIFF
--- a/armory/scenarios/poisoning_cifar10_witches_brew.py
+++ b/armory/scenarios/poisoning_cifar10_witches_brew.py
@@ -1,53 +1,61 @@
+import os
 from armory.scenarios.poison import Poison
 from armory.logs import log
 from armory.utils import config_loading
+from armory import paths
+from art.utils import to_categorical 
 import numpy as np
 
 
+
 class DatasetPoisonerWitchesBrew():
-    def __init__(self, attack, source_class, target_class, x_test, y_test):
+    def __init__(self, attack, x_test, y_test, source_class, target_class, data_filepath):
         """
         Individual source-class triggers are chosen from x_test.  At poison time, the
         train set is modified to induce misclassification of the triggers as target_class.
 
         """
         self.attack = attack
-        self.source_class = source_class
-        self.target_class = target_class
         self.x_test = x_test
         self.y_test = y_test
+        self.source_class = source_class
+        self.target_class = target_class
+        self.data_filepath = data_filepath
 
 
-    def poison_dataset(self, x_train, y_train, return_index=True):
+    def poison_dataset(self, x_train, y_train, trigger_index, return_index=True):
         """
         Return a poisoned version of dataset x, y
             if return_index, return x, y, index
-        If fraction is not None, use it to override default
         """
         if len(x_train) != len(y_train):
             raise ValueError("Sizes of x and y do not match")
+        
+        x_trigger = self.x_test[trigger_index]
+        if len(x_trigger.shape) == 3:
+            x_trigger = np.expand_dims(x_trigger, axis=0)
 
-        from art.utils import to_categorical
-        print(self.y_test[:10])
-        print(type(self.y_test))
-        index_target = np.where(self.y_test==self.source_class)[0][5] # TODO index 0 because where returns 1-length tuple.  5 to just pick one.
+        y_trigger  = to_categorical([self.target_class], nb_classes=len(np.unique(y_train)))
 
-        # Trigger sample
-        x_trigger = self.x_test[index_target:index_target+1]
-        y_trigger  = to_categorical([self.target_class], nb_classes=10)
-        print ("HELLO FROM OVER HERE")
-        print(index_target)
+        # TODO is armory data oriented and scaled right for art_experimental
+        poison_x, poison_y, poison_index, returned_trigger_index = self.attack.poison(self.data_filepath, x_trigger, y_trigger, x_train, y_train, trigger_index) 
         
 
-        filepath = "stuff.npz" # TODO where to get this from
+        if returned_trigger_index != trigger_index:
+            # We loaded a presaved dataset, but it was trained for a different trigger
+            # TODO move this into attack.poison probably
+            raise ValueError(f"The trigger image index requested in the config ({trigger_index}) "
+            + f"does not match the trigger image index from the saved adversarial dataset ({returned_trigger_index}).  "
+            + "Please clarify your intent by updating adhoc['trigger_index'] or attack['kwargs']['data_filepath'] in the config. "
+            )
 
-        poison_x, poison_y, poison_index = self.attack.poison(filepath, x_trigger, y_trigger, x_train, y_train) # TODO is armory data oriented and scaled right for art
-
+        # Return trigger_index to make sure 
         if return_index:
-            return poison_x, poison_y, poison_index
-        return poison_x, poison_y
+            return poison_x, poison_y, trigger_index, poison_index
+        return poison_x, poison_y, trigger_index
 
 
+# TODO Config: decide what goes in adhoc vs attack kwargs
 
 
 class CifarWitchesBrew(Poison):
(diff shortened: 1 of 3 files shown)


armory/scenarios/poison.py
```diff
@@ -377,17 +377,97 @@ def evaluate_current(self):
...
+        if self.use_filtering_defense:
+            log_lines = self.fairness_metrics.add_filter_perplexity(
+                self.y_clean, self.poison_index, self.indices_to_keep
+            )
+            for line in log_lines:
+                log.info(line)
+
+    def _add_accuracy_metrics_results(self):
+        """ Adds the main accuracy results to self.results:
+            poisoned and benign performance on whole test set and on source class # TODO is it source?
```
davidslater, line 456: Is the "TODO" here resolved?

armory/scenarios/poisoning_cifar10_witches_brew.py
```diff
@@ -0,0 +1,260 @@
...
+                attack_config["kwargs"].pop("data_filepath")
+                if "data_filepath" in attack_config["kwargs"].keys()
+                else None
+            )
+
+            attack = config_loading.load_attack(attack_config, self.model)
+            if data_filepath is not None:
+                data_filepath = os.path.join(
+                    paths.runtime_paths().dataset_dir, data_filepath
+                )
```
davidslater: Let's use the models directory, actually, as I would like to keep the dataset_dir just populated by TFDS datasets.

I think we should also put it in a subdir like "attacks" as so:
```
attack_dir = os.path.join(paths.runtime_paths().saved_model_dir, "attacks")
os.makedirs(attack_dir, exist_ok=True)
data_filepath = os.path.join(attack_dir, data_filepath)
```

armory/scenarios/poison.py
```diff
@@ -398,85 +478,41 @@ def finalize_results(self):
...
                 self.benign_test_accuracy_per_class[y]
             )
 
-        if hasattr(self, "fairness_metrics") and not self.check_run:
-            # Get unpoisoned test set
-            dataset_config = self.config["dataset"]
-            test_dataset = config_loading.load_dataset(
-                dataset_config, split="test", num_batches=None, **self.dataset_kwargs
-            )
+    def finalize_results(self):
```
davidslater, line 507: Does the poisoning scenario not use MetricsLogger at all?

armory/scenarios/poisoning_witches_brew.py
```diff
@@ -0,0 +1,260 @@
...
+
+        self._add_accuracy_metrics_results()
+
+        self._add_supplementary_metrics_results()
+
+        if self.use_filtering_defense:
+            self._add_filter_metrics_results()
+
+        if hasattr(self, "fairness_metrics") and not self.check_run:
+            self._add_fairness_metrics_results()
```
davidslater, line 260: Identical to `poison.py`. Remove.

armory/scenarios/poisoning_witches_brew.py
```diff
@@ -0,0 +1,260 @@
...
+                self.sample_exporter._make_output_dir()
+            name = "trigger" if self.i in self.trigger_index else "non-trigger"
+            self.sample_exporter._export_image(self.x[0], name=name)
+            self.sample_exporter.saved_samples += 1
+            self.sample_exporter.saved_batches = (
+                self.sample_exporter.saved_samples
+                // self.config["dataset"]["batch_size"]
+            )
+
+    def _add_accuracy_metrics_results(self):
```
davidslater, line 234: Need to call `super()._add_accuracy_metrics_results()` if you want those results too.

armory/scenarios/poisoning_witches_brew.py
```diff
@@ -0,0 +1,260 @@
...
+        eval_split = dataset_config.get("eval_split", eval_split_default)
+        log.info(f"Loading test dataset {dataset_config['name']}...")
+        self.test_dataset = config_loading.load_dataset(
+            dataset_config,
+            split=eval_split,
+            num_batches=self.num_eval_batches,
+            **self.dataset_kwargs,
+        )
+        self.i = -1
+
```
davidslater, line 177: Move this before `load_metrics` so that it is in the same order as `poison.py`.

armory/scenarios/poisoning_witches_brew.py
```diff
@@ -0,0 +1,260 @@
...
+        attack_config = self.config["attack"]
+        if attack_config.get("type") == "preloaded":
+            raise ValueError("preloaded attacks not currently supported for poisoning")
+
+        self.use_poison = bool(adhoc_config["poison_dataset"])
+
+        dataset_config = self.config["dataset"]
+        test_dataset = config_loading.load_dataset(
+            dataset_config, split="test", num_batches=None, **self.dataset_kwargs
+        )
```
davidslater, line 77: Do we need to make sure it is ordered in a specific way (e.g., shuffle=False)?

armory/scenarios/poisoning_witches_brew.py
```diff
@@ -0,0 +1,260 @@
...
+            self.trigger_index,
+        )
+
+        if return_index:
+            return poison_x, poison_y, poison_index
+        return poison_x, poison_y
+
+
+class WitchesBrewScenario(Poison):
+    def load_poisoner(self):
```
davidslater, line 66: The `load_poisoner` method seems very long. I think that a lot of this should be passed into the init method of DatasetPoisonerWitchesBrew, if possible.

armory/scenarios/poisoning_witches_brew.py
```diff
@@ -0,0 +1,260 @@
...
+            y_train,
+            self.trigger_index,
+        )
+
+        if return_index:
+            return poison_x, poison_y, poison_index
+        return poison_x, poison_y
+
+
+class WitchesBrewScenario(Poison):
```
davidslater, line 65: Is there a more general class of attacks that WitchesBrew falls into that we could create this as a more general scenario?

davidslater (changes requested):
Some questions and comments. I haven't tried running anything yet, however.

armory/scenarios/poisoning_cifar10_witches_brew.py
swsuggs: Done

armory/scenarios/poison.py
swsuggs: It has not done so as long as I've been around, nor does it use the metrics section of the config.  I suppose because some poisoning metrics don't fit into the mold assumed by the metricslogger class?  I haven't looked closely.  But it will need an overhaul with the measurement pr anyway

armory/scenarios/poisoning_witches_brew.py
swsuggs: good catch

armory/scenarios/poisoning_witches_brew.py
swsuggs: We don't; the original metrics were not relevant for this attack.

armory/scenarios/poisoning_witches_brew.py
swsuggs: Done

armory/scenarios/poisoning_witches_brew.py
swsuggs: yes thanks, I was thinking that was the default

davidslater: It probably should be, given that we are typically calling it for test, not training.

armory/scenarios/poisoning_witches_brew.py
swsuggs: Perhaps.  I'm not sufficiently versed in poisoning literature to know what other attacks share the specific features of this one.  Is "attacks that don't poison test images" a thing?  Non-backdoor attacks?  Attacks that target individual test images rather than a whole class?

armory/scenarios/poisoning_witches_brew.py
swsuggs: Now I remember, dataset_kwargs is set to include shuffle=False in poison.py.  So it was all right

armory/scenarios/poisoning_witches_brew.py
swsuggs: The trick is, the DatasetPoisoner is constructed with an `attack`, and the attack is constructed with these source_class and target_class things.  What do you think about the latest commit, where I just extracted a lot of it into a separate function

davidslater:
@swsuggs how long does it take to run the scenarios?

swsuggs:
The cifar10_witches_brew.json config takes a few hours, not sure exactly since I mostly ran it with reduced parameters for testing.  Spots to reduce (if you want) would be max_epochs under attack/kwargs, train_epochs under adhoc, or set use_poison_filtering_defense to false.

armory/scenarios/poisoning_witches_brew.py
davidslater: That's fine. We can think about generalizations and refactoring later.

armory/scenarios/poisoning_witches_brew.py
davidslater: I like it.

davidslater:
@swsuggs The main remaining things on my end are the TODOs in the comments. Have those been resolved?

armory/scenarios/poison.py
swsuggs: Yes.

PR #1407, opened 2022-04-13
Entailment ASR Integration

### LINKED ISSUE #1373: Updated Threat Model for LibriSpeech ASR - Eval 5
Currently, we just point to a list of transcripts and try to minimize targeted attack WER.

New (potential) material:
        phrase A to phrase A' (change sentence semantics, etc.)
        Entailment attack (measure of two sentences - whether they have same meaning, different, or neural)
        BERT score models
        Breakdown of WER for insertion, deletion, substitution

davidslater:
Fixes #1373


davidslater:
I need to move the test to tests and elaborate on it, but I think it's ready for code review. Will need to rebuild deepspeech container.

Should be able to test with
```
python docker/build.py pytorch-deepspeech --no-pull
armory run scenario_configs/librispeech_asr_entailment.json --num-eval-batches=1
```

scenario_configs/librispeech_entailment_asr.json
```diff
@@ -0,0 +1,86 @@
...
+        "name": "AutomaticSpeechRecognition"
+    },
+    "sysconfig": {
+        "docker_image": "twosixarmory/pytorch-deepspeech",
+        "external_github_repo": [
+            "SeanNaren/deepspeech.pytorch@V3.0"
+        ],
+        "gpus": "all",
+        "local_repo_path": null,
+        "num_eval_batches": 100,
```
lcadalzo, line 81: is it intentional to have `"num_eval_batches"` in the config? This will error out when a `--check` flag is added

armory/utils/metrics.py
```diff
@@ -2011,6 +2098,13 @@ def results(self):
                         raise ZeroDivisionError(
                             f"No values to calculate WER in {prefix}_{metric.name}"
                         )
+                if metric.name == "entailment":
```
lcadalzo, line 2089: While this conditional block maps integers to strings, in the `if self.full:` code block above, the integer is added to the results dict. 

So you get an output JSON like such:
```
"benign_entailment": [
            2
        ],
 "benign_total_entailment": {
            "entailment": 1,
            "total": 1
        },
```
The integers don't mean anything to a user right? So should they be omitted from the results JSON?

armory/utils/metrics.py
```diff
@@ -11,17 +11,94 @@
...
 from armory.logs import log
+from armory.utils.external_repo import ExternalPipInstalledImport
+
+# TODO: update label_mapping with total counts after measurement PR
+_ENTAILMENT_MODEL = None
+
+
+class Entailment:
+    # TEST Reference (from Mike):
+    # For reference, on the first 100 benign predictions, I got 94 entailment and 6 contradictions. The Eval 5 target phrases should produce 100 contradictions.
```
lcadalzo, line 35: could you add a description of what this is measuring?

armory/attacks/librispeech_target_labels.py
```diff
@@ -0,0 +1,544 @@
+# ground truth labels for first 100 samples in the test set
```
lcadalzo, line 1: we discussed in the meeting whether this is the best location for this. I thought it seemed a bit awkward since it's not an attack, per se, and it's now the only thing living in this dir. You had already considered placing it in `armory/utils/labels.py` and opted against that, correct?

armory/utils/metrics.py
```diff
@@ -11,17 +11,94 @@
...
+        )
+        with torch.no_grad():
+            scores = self.model(**features).logits
+            labels = [self.label_mapping[i] for i in scores.argmax(dim=1)]
+
+        return labels  # return list of labels, not (0, 1, 2)
+
+
+def test_entailment():
+    metric = Entailment()
```
lcadalzo: TODO: move to tests dir

scenario_configs/librispeech_entailment_asr.json
davidslater: Yes. The entailment labels are only defined for the first 100 samples.

armory/utils/metrics.py
lcadalzo: You also get logs as such:

```
2022-04-19 17:29:59 1m21s SUCCESS  armory.utils.metrics:log_task:2054 Average entailment on adversarial test examples relative to ground truth labels: 200.00%
```

It appears confusing to me but again I might be lacking context. Is that the desired behavior?

scenario_configs/librispeech_asr_snr_targeted.json
```diff
@@ -0,0 +1,81 @@
...
+        "name": "librispeech",
+        "train_split": "train_clean100"
+    },
+    "defense": null,
+    "metric": {
+        "means": false,
+        "perturbation": "linf",
+        "record_metric_per_sample": true,
+        "task": [
+            "word_error_rate"
```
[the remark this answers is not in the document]
davidslater: We didn't have a config that includes targeted WER with the full set of the targeted labels.
Since I was putting these labels into a file with the ground truth and entailment, I wanted to add a config to show the new usage.

scenario_configs/librispeech_entailment_asr.json
lcadalzo: `--num-eval-batches` in the config will be overriden by CLI, so do we need to catch the case where the CLI dictates something > 100?

lcadalzo: I ran with `--num-eval-batches 105` and `--skip-attack` and got 

```
"benign_total_entailment": {
            "contradiction": 6,
            "entailment": 99,
            "total": 105
        },
```

davidslater: If the attack is active and `--num-eval-batches > 100`, the attack will error at runtime (after the 100th sample).

The benign results are not actually limited to 100 - those are with respect to the transcripts from the dataset.

armory/utils/metrics.py
davidslater: Moved.

armory/attacks/librispeech_target_labels.py
davidslater: Correct. I think the data portion of `librispeech_target_labels.py` and code portion of `labels.py` should be separate. I think that `labels.py` should actually be moved out of utils at some point, but don't want to modify too much at present.

armory/utils/metrics.py
davidslater: docstring added

armory/utils/metrics.py
davidslater: Yeah, a lot of this has to do with the brittleness of the current MetricsLogger system. I will try making something a little nicer.

davidslater: This update should be a lot better.

davidslater:
@lcadalzo I think it's ready for a re-review

tests/unit/test_metrics.py
```diff
@@ -13,6 +13,58 @@
...
+def test_entailment():
+    """
+    Slow due to 1 GB file download and multiple model predictions
+    """
+    print("Loading Entailment Model")
+    metric = metrics.Entailment()
+    metric_repeat = metrics.Entailment()
+    assert metric.model is metric_repeat.model
+
+    print("Loading Target Transcripts")
```
lcadalzo, line 28: were the `print()` statements temporary or intended to make it into develop?

davidslater: Removed.

lcadalzo:
LGTM

PR #1408, opened 2022-04-13
adding mean success rate for video tracking

### LINKED ISSUE #1405: object tracking success rate metric
USCI proposed a metric to better assess the single object tracking scenario. It involves calculating average IoU for a range of IoU thresholds. This is similar to mAP for object detection. Their description can be found in their Eval4 submission (USCI_CARLA_video_tracking GARD Evaluation 4 Submission Documentation.docx). It would be good to include in Armory, if possible.

lcadalzo:
Closes #1405

### DIFF
--- a/armory/utils/metrics.py
+++ b/armory/utils/metrics.py
@@ -735,7 +735,7 @@ def video_tracking_mean_success_rate(y, y_pred):
     if len(y_pred) > 1:
         raise ValueError(f"y_pred expected to have length of 1, found {len(y_pred)}.")
 
-    thresholds = np.arange(0, 1.05, 0.05)
+    thresholds = np.arange(0, 1.0, 0.05)
     mean_success_rates = (
         []
     )  # initialize list that will have length num_videos, which currently is forced to be 1
--- a/armory/utils/metrics.py
+++ b/armory/utils/metrics.py
@@ -721,6 +721,47 @@ def video_tracking_mean_iou(y, y_pred):
     return mean_ious
 
 
+def video_tracking_mean_success_rate(y, y_pred):
+    """
+    Mean success rate averaged over all thresholds in {0, 0.05, 0.1, ..., 1.0} and all frames.
+    This function expects to receive a single video's labels/prediction as input.
+
+    y (List[Dict, ...]): list of length equal to number of examples. Each element
+                  is a dict with "boxes" key mapping to (num_frames, 4) numpy array
+    y_pred (List[Dict, ...]): list of length equal to number of examples. Each element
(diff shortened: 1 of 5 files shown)


lcadalzo:
A couple decisions being made here:

1. Compute the metric on a per-video basis, as opposed to accumulating all videos' predictions and measuring it at once
2. Simply return the mean rather than an array of success rates at different thresholds

armory/utils/metrics.py
```diff
@@ -721,6 +721,47 @@ def video_tracking_mean_iou(y, y_pred):
...
+
+    thresholds = np.arange(0, 1.05, 0.05)
+    mean_success_rates = (
+        []
+    )  # initialize list that will have length num_videos, which currently is forced to be 1
+    for video_idx in range(len(y_pred)):
+        success = np.zeros(len(thresholds))
+
+        # Selecting first element since y_pred is forced to have length 1
+        y_pred_boxes = y_pred[0]["boxes"]
```
swsuggs, line 746: Why exactly are you bothering with a loop if y_pred only has predictions for one video?  I would say either remove the loop or change this to `y_pred_boxes = y_pred[video_idx]["boxes"]` if you expect to handle longer y_preds in the future

armory/utils/metrics.py
```diff
@@ -721,6 +721,47 @@ def video_tracking_mean_iou(y, y_pred):
...
+    y (List[Dict, ...]): list of length equal to number of examples. Each element
+                  is a dict with "boxes" key mapping to (num_frames, 4) numpy array
+    y_pred (List[Dict, ...]): list of length equal to number of examples. Each element
+                  is a dict with "boxes" key mapping to (num_frames, 4) numpy array
+    """
+    _check_video_tracking_input(y, y_pred)
+    if len(y_pred) > 1:
+        raise ValueError(f"y_pred expected to have length of 1, found {len(y_pred)}.")
+
+    thresholds = np.arange(0, 1.05, 0.05)
```
swsuggs, line 738: Given that line 757 defines success to be iou > threshold, I don't think you should include a threshold of 1.  Success will always be False for that threshold which means the mean success rate will be artificially reduced.  On the other hand, if you change line 757 to be iou >= threshold, then maybe you shouldn't include a threshold of 0, because iou >= 0 is always True

lcadalzo: I had copied this from the other implementation but I agree with you, it didn't make sense to me either. I'll change it to stop at 0.95

lcadalzo:
Re-running CI passed all tests: https://github.com/lcadalzo/armory/actions/runs/2208260451

swsuggs:
looks great

PR #1409, opened 2022-04-14
For Issue #1375 - Eval5 carla obj det robust dpatch

yusong-tan:
Updated `art_experimental carla_obj_det_patch` to attack both RGB and Depth.

For reference, I got the following results on `test` using `large` split (note, the multimodal defended model was trained to defend against RGB attack only, so ignore its performance on the updated attack)

- carla_obj_det_multimodal_dpatch_defended.json: 0.33 mAP / 42 hallucinations per image
- carla_obj_det_multimodal_dpatch_undefended.json: 0.34 mAP / 14 hallucinations per image
- carla_obj_det_dpatch_undefended.json: 0.18 mAP / 11 hallucinations per image
- carla_obj_det_dpatch_defended.json: 0.18 mAP / 9 hallucinations per image

### DIFF
--- a/armory/art_experimental/attacks/carla_obj_det_patch.py
+++ b/armory/art_experimental/attacks/carla_obj_det_patch.py
@@ -122,7 +122,9 @@ def apply_ccm(patch, ccm, gamma=2.2, Vandermonde=True, degree=1):
     return corrected_RGB
 
 
-def insert_transformed_patch(patch, image, gs_shape, rgb, patch_coords=[], image_coords=[]):
+def insert_transformed_patch(
+    patch, image, gs_shape, rgb, patch_coords=[], image_coords=[]
+):
     """
     Insert patch to image based on given or selected coordinates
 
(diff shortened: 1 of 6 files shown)


lcadalzo:
So the sole aim of this PR is to allow for the DPatch attack to perturb depth channels?

lcadalzo:
Is there repeated code that could be shared between the two CARLA patch attacks?

yusong-tan:
> So the sole aim of this PR is to allow for the DPatch attack to perturb depth channels?

Yes, 

> Is there repeated code that could be shared between the two CARLA patch attacks?

I don't think so. One uses PyTorch perspective transform, and the other uses openCV perspective transform.

lcadalzo:
It looks like `in_polygon()`, `get_avg_depth_value()`, `linear-to_log()` and `log_to_linear()` are needed by both `CARLADapricotPatch` as well as `CARLAAdversarialPatchPyTorch`. These can be moved to a separate file which each attack imports from

lcadalzo:
Confirmed that I also get an adversarial mAP of 0.18 and 11.3 adversarial hallucinations per image for the RGB case

lcadalzo:
Currently running benchmarking experiments for the mulitmodal case. Once those results come in, we should be good to merge

lcadalzo:
For multimodal (undefended), I'm seeing an adversarial mAP of 0.32 and 15.9 hallucinations per image

yusong-tan:
> For multimodal (undefended), I'm seeing an adversarial mAP of 0.32 and 15.9 hallucinations per image

It's close to my results as different patch initialization do create some variance in attack performance.

PR #1412, opened 2022-04-18
Updated CARLA Object Detection Training Data

yusong-tan:
Dataset url to be sent separately


lcadalzo:
Building the cached dataset is moving very slowly (~15s per example). Given that there are ~9000 examples, I estimate this will take ~36 hours. Shouldn't be an issue for next week's release if things work the first time, but something to note

yusong-tan:
> ~Building the cached dataset is moving very slowly (15s per example). Given that there are around 9000 examples, I estimate this will take about 36 hours. Shouldn't be an issue for next week's release if things work the first time, but something to note~
> 
> Update: I'm able to build it much faster on a different machine

For reference, it took only a few minutes to build on my machine.

lcadalzo:
> For reference, it took only a few minutes to build on my machine.

Thanks for the context, same for me after using a different machine

swsuggs:
looks good

Issue #1263, opened 2022-02-02
Jupyter - TQDM errors and Image printing errors


davidslater:

davidslater:
Needs to follow this: https://pypi.org/project/tqdm/#ipython-jupyter-integration
The underlying error looks like it's from ART, when using RobustDPatch:
```
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
/workspace/armory/scenarios/scenario.py in evaluate(self)
    351         try:
--> 352             results = self._evaluate()
    353         except Exception as e:

/workspace/armory/scenarios/carla_object_detection.py in _evaluate(self)
    118 
--> 119         self.evaluate_all()
    120         self.finalize_results()

/workspace/armory/scenarios/scenario.py in evaluate_all(self)
    254             self.next()
--> 255             self.evaluate_current()
    256 

/workspace/armory/scenarios/scenario.py in evaluate_current(self)
    324         if not self.skip_attack:
--> 325             self.run_attack()
    326 

/workspace/armory/scenarios/carla_object_detection.py in run_attack(self)
     66                 y_patch_metadata=[y_patch_metadata],
---> 67                 **self.generate_kwargs
     68             )

/workspace/armory/art_experimental/attacks/carla_obj_det_patch.py in generate(self, x, y, y_patch_metadata)
    447                 patch = super().generate(
--> 448                     np.expand_dims(x[i], axis=0)
    449                 )  # untargeted attack

/opt/conda/lib/python3.7/site-packages/art/attacks/evasion/dpatch_robust.py in generate(self, x, y, **kwargs)
    176 
--> 177         for i_step in trange(self.max_iter, desc="RobustDPatch iteration", disable=not self.verbose):
    178             if i_step == 0 or (i_step + 1) % 100 == 0:

/opt/conda/lib/python3.7/site-packages/tqdm/auto.py in trange(*args, **kwargs)
     39         """
---> 40         return tqdm(range(*args), **kwargs)
     41 

/opt/conda/lib/python3.7/site-packages/tqdm/notebook.py in __init__(self, *args, **kwargs)
    223         self.container = self.status_printer(
--> 224             self.fp, total, self.desc, self.ncols)
    225         self.sp = self.display

/opt/conda/lib/python3.7/site-packages/tqdm/notebook.py in status_printer(_, total, desc, ncols)
     96             raise ImportError(
---> 97                 "IProgress not found. Please update jupyter and ipywidgets."
[...]

PR #1416, opened 2022-04-19
jupyter widgets and tests

davidslater:
Fixes https://github.com/twosixlabs/armory/issues/1263 and adds manual tests and test data for jupyter.


davidslater:
You’ll need to rebuild base to test:
```
bash docker/build-base.sh 
python docker/build.py pytorch --no-pull
armory launch pytorch --jupyter
# connect to jupyter lab and run tests in tests/test_jupyter
```

mwartell:
Because it alters Dockerfile-base, this PR requires a rebuild and push of the `twosixarmory/base` docker image.

I just wanted to call this out explicitly in case it wasn't intended. Publishing anew is not hard, but I do believe we are hoping to update base infrequently. Of course [practicality beats purity](https://peps.python.org/pep-0020/#:~:text=Although%20practicality%20beats%20purity.).

davidslater:
Yes, wanted to do that instead of modifying all of the derivative ones. I was hoping to get a bunch of base changes in prior to `0.15.0` release, and then get a bit more constant after that.

davidslater:
@mwartell I think it's good for re-review. The only other addition is that I noticed an issue with jupyter prompt when using `armory launch`, so I've fixed that.

mwartell (approved):
LGTM

PR #1417, opened 2022-04-19
add pytest loguru logcap fixure with tests

mwartell:
This uses the loguru suggested change to the pytest `caplog` mechanism for assertions on logging output. Sample use can be seen in tests/unit/test_logging.py.


lcadalzo (approved):
LGTM

PR #1420, opened 2022-04-20
CARLA --skip-attack bug fix

### LINKED ISSUE #1419: Error when running CARLA with --skip-attack
This issue occurs because the CARLA scenario code instantiates another `MetricsLogger` instance [here](https://github.com/twosixlabs/armory/blob/develop/armory/scenarios/carla_object_detection.py#L110). The variable `self.skip_attack` is not propagated to this instantiation, so when the instantiation occurs, `skip_attack` is set to the default `None`. Consequentially, in `metrics.py`, a list of adversarial tasks is generated [here](https://github.com/twosixlabs/armory/blob/develop/armory/utils/metrics.py#L1805) when it shouldn't be (assuming `--skip-attack` is in fact set). This causes the following error in `finalize_results()`:

```
File "/workspace/armory/utils/metrics.py", line 622, in _check_object_detection_input                                                 
    raise ValueError("Received no labels or predictions")   
```

To reproduce, run one of the CARLA configs with `--skip-attack --num-eval-batches 1` set.

lcadalzo:
Fixes #1419

### DIFF
--- a/armory/scenarios/carla_object_detection.py
+++ b/armory/scenarios/carla_object_detection.py
@@ -108,5 +108,8 @@ class CarlaObjectDetectionTask(Scenario):
             if k in metric_config
         }
         self.metrics_logger_wrt_benign_preds = metrics.MetricsLogger.from_config(
-            subset_config, skip_benign=True, targeted=False
+            subset_config,
+            skip_benign=True,
+            targeted=False,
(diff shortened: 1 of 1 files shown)


davidslater (approved):
LGTM

davidslater (commented):
LGTM

PR #1426, opened 2022-04-21
Pin Ffmpeg to conda forge

davidslater:


davidslater:
The witches brew update was due to black.

davidslater:
Added a test in case the encoder library isn't present again.

Issue #1371, opened 2022-03-25
Updated CARLA Video Tracking Dataset for Eval 5


davidslater:
Higher resolution / graphic texture than current dataset.

yusong-tan:
Submitted as PR #1431

PR #1431, opened 2022-04-22
For Issue #1371 - Eval5 carla video tracking dev dataset

yusong-tan:
Link to dataset sent separately


yusong-tan:
For reference, against `undefended` model and using `max_iter=20`, `step_size=0.1`, I got 54.9% benign mean IoU and 18.2% adversarial mean IoU. When exported as videos, the patch texture in a video may _appear_ to be dynamic, but this is a visual artifact caused by imperfect projections across frames due to camera motion.

lcadalzo:
Against an undefended model using `max_iter=100` and `step_size=0.02` I got 54.9% benign mean IoU and 16% adversarial mean IoU

PR #1432, opened 2022-04-22
document armory base image build

mwartell:
In documenting this simple task I found there was too much operator knowledge needed, so I fixed it.

The [docker.md instructions for publishing a new base](docs/docker.md#publishing_a_new_base) should be
self explanatory. If not, that's a bug

### DIFF
--- a/docker/build-base.sh
+++ b/docker/build-base.sh
@@ -1,4 +1,44 @@
+#!/usr/bin/env bash
+set -e
+
+usage() { echo "usage: $0 [--dry-run] [--push]" 1>&2; exit 1; }
+
+dryrun=
+push=
+
+while [ "${1:-}" != "" ]; do
+    case "$1" in
+        -n|--dry-run)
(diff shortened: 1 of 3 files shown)


mwartell:
The broken CI test is meaningful. Merging prohibited until fixed.

mwartell:
This most recent fix works, but it doesn't make sense.  The failure was because `armory --version` in the ci_test.yml was giving a bad version ("0.1") instead of "0.14.7.ge01675f" as was being built.  

The fix for this is pretty standard with setuptools_scm. By default actions/checkout in an Action does a shallow clone at depth=1 of the repo which defeats the tag chasing code in scm.  Setting that job to checkout the whole tree (fetch-depth: 0) clears the problem.

The thing I don't understand is why the pytorch and pytorch-deepspeech stanzas above and below the `docker-tf2-unit` step do not do a full checkout but work.  I'd like @shenshaw26 to take a look at ci_test.yml to see if he groks why they are different.

PR #1433, opened 2022-04-22
Improve trigger selection for witches' brew

swsuggs:
For #1423 

If the user specifies trigger_index in the config, they can leave source_class null and scenario will infer it.  Alternatively, if they specify source class, they may leave trigger_index null and scenario will choose randomly from the specified source(s).  Determines quantity from longest of source_class, target_class, trigger_index.  Raises helpful errors if input is bad.

If a pre-poisoned dataset already exists at `config/attack/kwargs/data_filepath`, then the config parameters are verified against those used to create the existing dataset, raising an exception if there is a discrepancy.

Examples:

```
source_class: [1,2,3]
target_class: 0
trigger_index: null
```
This will randomly choose a trigger image from each source class, using target of 0 for each of them.

```
source_class: 1
target_class: 0
trigger_index: [null, null, null]
```
This will randomly select three triggers from class 1, using a target of 0 for each.

```
source_class: null
target_class: [1,2]
trigger_index: [12, 20]
```
This will use test images 12 and 20, inferring their labels from the test set, using a target class of 1 for image 12 and 2 for image 20.

Finally, when comparing against a pre-existing poisoned dataset, a null value for trigger_index is interpreted as a wildcard; i.e. it does not matter what triggers the dataset was poisoned for as long as they are from the correct class.


swsuggs:
Changing to wip because I thought of something else this needs: if trigger_index, source_class, and target_class are all null, then use the prepoisoned dataset at the filepath in the config.  right now it would raise an error.

lcadalzo:
So let's say you know you want a `source_class` of 1 and a `target_class` of 0, but you want to vary how many triggers are selected. Do I understand correctly that the way to do that is to vary the number of null's in the `trigger_index` list?

I.e.
```
source_class: 1
target_class: 0
trigger_index: [null, null, null]
```
will select 3 images as triggers but
```
source_class: 1
target_class: 0
trigger_index: [null]
```
would only select one? If I'm correct here, this feels like an unintuitive interface. Not immediately sure what the smoothest way to address this would be but perhaps a field that specifies the number of triggers

davidslater:
@swsuggs you'll need to run the pre-commit checks and update

swsuggs:
@lcadalzo You are correct, but you can also do this:
```
source_class: [1,1,1]
target_class: 0
trigger_index: null
```
or this
```
source_class: 1
target_class: [0,0,0]
trigger_index: null
```
or this
```
source_class: [1,1,1]
target_class: [0,0,0]
trigger_index: null
```
or this
```
source_class: [1,1,1]
target_class: [0,0,0]
trigger_index: [null, null, null]
```
or at least two other things that I won't put here for the sake of brevity, although it's probably too late for that.

I was thinking with enough right ways to do it, it will be easy for people to guess one of them.  I do see what you are saying, but I actually feel like having another field like `n_triggers` would increase the number of ways to do this _wrong_, thus making it _less_ intuitive.  This is because that field would have to match the length of the other lists, if we did have (for example) `source_class:[1,2,3,4]`.

davidslater (changes requested):
Can you revert the updates to `poison.py`? I am updating this on my measurement PR, and this will make merging a bit of a pain. (I will update names on my branch to match, though)

davidslater (approved):
LGTM

PR #1438, opened 2022-04-25
shortcircuit init when querying version

### LINKED ISSUE #1437: armory --version doesn't just output version
```
(armory) √ david.slater@mac0909:~/armory% ARMORY=$(armory --version)
(armory) √ david.slater@mac0909:~/armory% echo $ARMORY
2022-04-25 09:39:37  8s INFO     art.config:set_data_path:55 set ART_DATA_PATH to /Users/david.slater/.art/data
0.14.7.g517058ab
```

Relatedly, getting the armory version takes 8 seconds.

This is due to `armory/__init__.py`. There are a lot of imports there, including PyTorch, TensorFlow, and ART. It also sets the art path automatically. The init block seems to be the best place to enable all of this, but shouldn't need to happen if just getting the version.

davidslater:
Fixes #1437 

May not be the best long-term fix, but can revisit as we break out the armory runtime from the outer component.

### DIFF
--- a/armory/__init__.py
+++ b/armory/__init__.py
@@ -20,6 +20,7 @@ except ModuleNotFoundError:
 
 import pathlib
 import re
+import sys
 import subprocess
 
 from armory.logs import log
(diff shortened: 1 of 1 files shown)


davidslater:
To test, just run `armory --version` or `python -m armory --version` and compare the `develop` branch from the fix.

If you don't have art in the current environment, however, you won't run into this problem.

shenshaw26:
Yeah I agree, not awesome but good for now.

shenshaw26 (approved):

PR #1440, opened 2022-04-25
Check fix

### LINKED ISSUE #1249: armory --check option should override num_eval_batches
In https://github.com/twosixlabs/armory/blob/39e07d30c275da53a1266cd01abb7c2b2d83a932/armory/scenarios/main.py#L232 there is a line that warns the `--check` option will override num_eval_batches but there is no code that actually implements that override.

davidslater:
Fixes #1249

### DIFF
--- a/armory/scenarios/main.py
+++ b/armory/scenarios/main.py
@@ -217,8 +217,9 @@ if __name__ == "__main__":
 
     if args.check and args.num_eval_batches:
         log.warning(
-            "--num_eval_batches will be overwritten and set to 1 since --check was passed"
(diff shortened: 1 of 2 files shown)


lcadalzo:
are you able to reproduce the error described in the issue? I.e. that when you run a config containing `"num_eval_batches"` specified in the config with the `--check` flag it errors out? I am not able to reproduce the error and am not confident that the value in the config is being propagated to the scenario. Need to dig into the code

lcadalzo:
Actually, I can reproduce the error when I omit the `--interactive` flag which I was previously using

lcadalzo:
LGTM

PR #1441, opened 2022-04-25
armory step-by-step docs and/or notebook

### LINKED ISSUE #1336: Add docs (and/or jupyter notebook?) for how to run Armory scenarios step-by-step
The release of Armory 0.14.0 refactored scenarios such a way that they can be run more interactively/step-by-step using the `next()` and `evaluate_current()` methods. We currently lack documentation on how to do so

lcadalzo:
Closes #1336


davidslater (changes requested):
It would be good to have some link in the documentation that points to the notebook, even if small.

At the introduction, it would be helpful to explicitly state how to follow along, e.g. "do the following":
`armory run scenario_configs/carla_obj_det_adversarialpatch_undefended.json --jupyter`

Second line in Loading the Scenario should be in code.

You may need to add:
```
# If running in container you may need to:
import os
import sys
sys.path.insert(0, os.path.abspath(".."))
```
(Or, just ensure that they have the latest docker images - was not an issue after updating)

After `next()`, you can also mention that they have access to `s.i`, which is the batch number.

The attack is taking about 5000 seconds on my system. You can do a modification in place to the config and reload the attack to go from 1000 iterations to 5 iterations, like so:
```
print(s.config["attack"]["kwargs"]["max_iter"]). # 1000
s.config["attack"]["kwargs"]["max_iter"] = 5
s.load_attack()
```

In Exporting Data, it is unclear how to find the output directory. Probably easiest:
```s.sample_exporter.output_dir```
Though you may want to add a caveat that if running in docker, this is the path inside docker.

**Review by davidslater — CHANGES_REQUESTED** [review]
Very minor:

at the end, I would add code blocks for
`s.save()`
and
```
results = s.prepare_results()
print(results)
```

davidslater (approved):
You call `.load()` twice, so remove one of those and go ahead and merge in.

PR #1442, opened 2022-04-26
Fix #1321 - Documentation for using cuda 10.2

### LINKED ISSUE #1321: Docs: how do we support CUDA 10?
Our docker images start at 11.3. How would someone who needs to use CUDA 10.2 build the images from there? (I don't expect us to directly support / debug that, but some documentation pointer would be helpful.)

shenshaw26:
Attempt to resolve #1321

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -169,6 +169,18 @@ armory exec pytorch --gpus=0 -- nvidia-smi
 
 The TensorFlow versions we support require CUDA 10+.
 
+Our current docker images use CUDA 11.3 as the base image.  To use CUDA 10.2, you will need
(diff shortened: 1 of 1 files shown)


davidslater:
Can you pull in the caveats I added in https://github.com/twosixlabs/armory/pull/1439 ?

davidslater:
I think that will likely fail. I would say something like:
```
This is not actively tested, so it may require further modification of library dependencies.

While PyTorch does support CUDA 9, we do not provide support in armory due to TFDS dependencies and we do not recommend using.
```

davidslater (approved):
LGTM

PR #1446, opened 2022-04-27
Fix 1434 - Understand Armory Version in CI Tests

shenshaw26:


shenshaw26:
I think something else was going on, bu tall these tests pass without the fetch-depth bit.  See:

https://github.com/twosixlabs/armory/runs/6196743409?check_suite_focus=true

I recommend we merge this.

.github/workflows/ci_test.yml
```diff
@@ -120,6 +115,7 @@ jobs:
           pip install -r requirements.txt
           pip install -e .
           python docker/build.py tf2
+          echo "armory version: $(armory --version)"
```
davidslater, line 118: Remove, unless you think it makes sense to add to the other docker tests.

shenshaw26: done.

davidslater (approved):
LGTM. Unclear why that test had failed.

Armory versions are wrong, but they are consistent across tests and tied to the current commit.

PR #1450, opened 2022-04-28
refactor scenario finalization

### LINKED ISSUE #1448: Move results preparing/saving outside of evaluate() method
The intended use case is as follows:

if I'm running Armory interactively/step-by-step, I may be calling `s.next()`, `s.run_benign()`, `s.run_attack()` in succession until I decide I want to close out the experiment and finalize my results.

At this point, I could call `s.finalize_results()` and this will log the metrics results. However, it won't save the output JSON. This step is occurring inside of `evaluate()`, which I don't want to call.

lcadalzo:
Closes #1448

### DIFF
--- a/armory/scenarios/scenario.py
+++ b/armory/scenarios/scenario.py
@@ -376,10 +376,10 @@ class Scenario:
 
         if results is None:
             log.warning(f"{self._evaluate} returned None, not a dict")
-        output = self._prepare_results(self.config, results)
-        self._save(output)
 
-    def _prepare_results(self, config: dict, results: dict, adv_examples=None) -> dict:
+        self.save()
+
+    def _prepare_output(self, adv_examples=None) -> dict:
         """
         Build the JSON results blob for _save()
 
@@ -391,16 +391,15 @@ class Scenario:
 
         output = {
             "armory_version": armory.__version__,
-            "config": config,
-            "results": results,
+            "config": self.config,
+            "results": self.results,
             "timestamp": int(self.time_stamp),
         }
         return output
 
-    def _save(self, output: dict):
-        """
-        Save json-formattable output to a file
-        """
+    def save(self):
+        output = self._prepare_output()
+
         override_name = output["config"]["sysconfig"].get("output_filename", None)
         scenario_name = (
             override_name if override_name else output["config"]["scenario"]["name"]


lcadalzo:
With the current approach, you'll need to call `finalize_results()` (which compiles and logs results) and then `save()` to write the JSON. 

An alternative would be to have `save()` do both these things, but I think having them split makes sense. I've added a check for the error case where one calls `save()` without calling `finalize_results()`

davidslater (changes requested):
A few minor things:
1) since you've updated the method already, can you remove the "adv_examples" from _prepare_results kwarg? It's never used and we save examples in a different way now.
2) maybe we can make `_prepare_results` into `prepare_results`, and document it, so that if someone just wants to get the result dictionary, and maybe save it elsewhere, they have an easier way to access that?
3) `self._evaluate()` both sets and returns self.results. Then the returned version is checked. Maybe we just have _evaluate set self.results (via finalize_results) and evaluate just references self.results? I.e., remove line 358, change line 365 to self._evaluate(), and modify 377 to `if self.results is None`.

lcadalzo:
Only potential TODO is to document use of `prepare_results()`, but I think that might best belong in the Jupyter notebook

armory/scenarios/scenario.py
```diff
@@ -374,33 +373,35 @@ def evaluate(self):
...
             log.warning(f"{self._evaluate} returned None, not a dict")
-        output = self._prepare_results(self.config, results)
-        self._save(output)
 
-    def _prepare_results(self, config: dict, results: dict, adv_examples=None) -> dict:
+        self.save()
+
+    def prepare_results(self) -> dict:
         """
         Build the JSON results blob for _save()
```
davidslater, line 383: docstring should be `save()`, not `_save()`, now.

armory/scenarios/scenario.py
```diff
@@ -374,33 +373,35 @@ def evaluate(self):
                 log.exception("Encountered error during scenario evaluation.")
             sys.exit(1)
 
-        if results is None:
+        if self.results is None:
             log.warning(f"{self._evaluate} returned None, not a dict")
```
davidslater, line 377: Similarly, change to reflect update in _evaluate.

armory/scenarios/scenario.py
```diff
@@ -374,33 +373,35 @@ def evaluate(self):
                 log.exception("Encountered error during scenario evaluation.")
             sys.exit(1)
 
-        if results is None:
+        if self.results is None:
```
davidslater, line 376: Actually, if `_evaluate` never sets `results`, then this raise an AttributeError.

This could be fixed by adding `self.results = None` in the scenario `__init__` method or changing this to something like `if getattr(self, "results", None) is None

davidslater:
> 

Agreed, juypter notebook would work best for documenting `prepare_results`

davidslater (approved):
LGTM

PR #1452, opened 2022-04-28
Minor update to CARLA Robust Dpath attack 

yusong-tan:
@lcadalzo Minor update to remove dependency on `shape`, `cc_ground_truth`, and `cc_scene` keywords in Eval5 adversarial datasets


lcadalzo:
thanks @yusong-tan. What's the purpose behind setting `apply_realistic_effects` to `False`?

yusong-tan:
> thanks @yusong-tan. What's the purpose behind setting `apply_realistic_effects` to `False`?

`apply_realistic_effects = False` means no color correction will be done on the patch. This makes it more aligned with the `AdversarialPatch` attack, which doesn't apply color correction by default.

lcadalzo:
LGTM

Issue #1455, opened 2022-04-28
Measurement fix for witches' brew


swsuggs:
Add:
N_poisoned_train_samples
N_clean_train_samples
filter metrics (tpr, fpr, f1, etc)
test_set_class_labels
filter_perplexity

PR #1456, opened 2022-04-28
Fix measurement for witches brew

swsuggs:
Addresses #1455

### DIFF
--- a/armory/scenarios/poisoning_witches_brew.py
+++ b/armory/scenarios/poisoning_witches_brew.py
@@ -324,7 +324,18 @@ class WitchesBrewScenario(Poison):
         self.config["adhoc"]["trigger_index"] = [int(t) for t in self.trigger_index]
         self.config["adhoc"]["source_class"] = [int(c) for c in self.source_class]
         self.config["adhoc"]["target_class"] = [int(c) for c in self.target_class]
+
+        self.n_poisoned = int(len(self.poison_index))
+        self.n_clean = (
+            len(self.y_clean) - self.n_poisoned
+        )  # self.y_clean is the whole pre-poison train set
+        poisoned = np.zeros_like(self.y_clean, np.bool)
+        poisoned[self.poison_index.astype(np.int64)] = True
+        self.probe.update(poisoned=poisoned, poison_index=self.poison_index)
+        self.hub.record("N_poisoned_train_samples", self.n_poisoned)
+        self.hub.record("N_clean_train_samples", self.n_clean)
         self.train_set_class_labels = sorted(np.unique(self.y_clean))
+        self.probe.update(y_clean=self.y_clean)
         for y in self.train_set_class_labels:
             self.hub.record(
                 f"class_{y}_N_train_samples", int(np.sum(self.y_clean == y))
@@ -345,8 +356,21 @@ class WitchesBrewScenario(Poison):
             **self.dataset_kwargs,
         )
         self.i = -1
+        self.test_set_class_labels = set()
+
 
     def load_metrics(self):
+        if self.use_filtering_defense:
+            # Filtering metrics
+            self.hub.connect_meter(
+                Meter(
+                    "filter",
+                    metrics.get_supported_metric("tpr_fpr"),
+                    "scenario.poisoned",
+                    "scenario.removed",
+                )
+            )
+
         self.hub.connect_meter(
             Meter(
                 "mean_accuracy_non_trigger_images",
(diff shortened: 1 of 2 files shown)


lcadalzo:
Is the only difference between `poison_dataset()` method of witches brew compared to the standard that it has these lines?

```
 # make sure config has updated and serializable versions of source, target, and trigger
 self.config["adhoc"]["trigger_index"] = [int(t) for t in self.trigger_index]
 self.config["adhoc"]["source_class"] = [int(c) for c in self.source_class]
 self.config["adhoc"]["target_class"] = [int(c) for c in self.target_class]
```

If so, I think that in the witches brew file you should be to call `super().poison_dataset()` and then add the above lines. That would remove a lot of copied code

swsuggs:
I agree it's a lot of copied code (it wasn't, before the measurement stuff), but the other difference is right above those lines, where the call to `self.poisoner.poison_dataset(()` returns several things beyond what it does in poison.py.

lcadalzo:
To me it seems like everything in `poison_dataset()` that happens after the `self.poisoner.poison_dataset()` call could go in a different (shared) method.

armory/scenarios/poisoning_witches_brew.py
```diff
@@ -419,6 +442,7 @@ def run_benign(self):
         self.probe.update(y=y, y_pred=y_pred)
 
         self.y_pred = y_pred  # for exporting when function returns
+        self.test_set_class_labels.update(y)
```
lcadalzo, line 445: is `self.test_set_class_labels` only supposed to be used if `hasattr(self, "fairness_metrics")`? Just confirming that's the case bc otherwise I don't see it being used

lcadalzo:
I see `filter_perplexity` in the issue but not anywhere in the PR. Is that intentional?

armory/scenarios/poisoning_witches_brew.py
swsuggs: Yes, fairness_metrics is the only part that uses `test_set_class_labels`

lcadalzo:
> `filter_perplexity` wasn't measuring successfully because line 338 was missing (updating a probe variable)

did you mean to push a commit that addresses this? Or is it measuring successfully for you at this state?

lcadalzo:
So when you run the config are you seeing filter_perplexity in the output JSON? I am not:

```
I have no name!@a9b03297115a:/workspace$ cat /armory/outputs/2022-04-29T134744.524284/WitchesBrewScenario_1651248197.json | grep perp
I have no name!@a9b03297115a:/workspace$ 

```

swsuggs:
Oh, sorry I should have clarified, you have to add to the config under adhoc:
```
"compute_fairness_metrics":true,
"explanatory_model":"cifar10_explanatory_model"
```
These lines were included in the eval5 configs pr that you merged to develop.  I just merged develop to this branch so you should be good to try again if you want.

lcadalzo:
Got it, the refactor looks good.

I think there may be an issue with the config in develop, though. The explanatory model it selects, see [here](https://github.com/twosixlabs/armory/blob/develop/scenario_configs/eval5/poisoning/cifar10_witches_brew.json#L6), is invalid and should be set to `cifar10_silhouette_model` instead. I get the following error running it:

```
ValueError: Config should specify model for fairness metrics.  Set adhoc/explanatory_model to one of ['gtsrb_silhouette_model', 'resisc10_silhouette_model', 'cifar10_silhouette_model'], or set adhoc/compute_fairness_metrics to false if these metrics are not desired.

```

lcadalzo:
Oddly, though, after I change it to `cifar10_silhouette_model` and run the config I get the following error:

```
 File "/workspace/armory/scenarios/poisoning_witches_brew.py", line 69, in poison_dataset
    ) = self.attack.poison(
        │    │      └ <function GradientMatchingWrapper.poison at 0x7fe38ffacc10>
        │    └ <armory.art_experimental.attacks.gradient_matching.GradientMatchingWrapper object at 0x7fe38ff866d0>
        └ <armory.scenarios.poisoning_witches_brew.DatasetPoisonerWitchesBrew object at 0x7fe38ff86850>

  File "/workspace/armory/art_experimental/attacks/gradient_matching.py", line 70, in poison
    raise ValueError(

ValueError: Adversarial dataset at filepath /armory/saved_models/attacks/cifar_gradient_matching_precomputed.npz was trained for 3 trigger images, not 1 as requested by the config.

```

Are you able to reproduce?

swsuggs:
Fixed the explanatory model, thanks for catching that.

For the other error, that's exactly what it should do.  Like it says, the dataset you generated before is different from what the config is now requesting.  (That's because I changed the config since you last ran it)  So just delete `/armory/saved_models/attacks/cifar_gradient_matching_precomputed.npz` and it will generate a new dataset (set `attack/kwargs/max_epochs` to 1 if you're just testing)

lcadalzo:
LGTM

PR #1463, opened 2022-04-29
fix carla test set preprocessing

### LINKED ISSUE #1460: Separate CARLA dev and test preprocessing
Since the new dev set is of a different shape than the test set, these 2 datasets should have different preprocessing functions at least until we receive the new test set.

When trying to `get_batch()` from the test set, this is the error you'll get:

```
ValueError: shape (1, 600, 800, 3) does not match shape (None, 960, 1280, 3)

```

Note: this applies to both object detection and video tracking datasets

lcadalzo:
Closes #1460


swsuggs (approved):

swsuggs:
I can successfully load all four datasets

Issue #1459, opened 2022-04-29
add preprocessing function to gtsrb dataset


swsuggs:
`poisoning_gtsrb_scenario.py` defines a function `gtsrb_scenario_preprocessing()` that is used to scale all gtsrb images to the same size.  It seems like this should occur in datasets.py when the dataset is loaded so that other scenarios can more easily use this dataset.

PR #1473, opened 2022-05-02
Gtsrb preprocessing

swsuggs:
For #1459 
Moves gtsrb preprocessing function to dataset.py instead of the gtsrb scenario.
This allows other scenarios like witches' brew to use gtsrb.


davidslater (approved):
LGTM

PR #1474, opened 2022-05-02
fix so2sat metrics

### LINKED ISSUE #1469: Error running so2sat configs
Note that this applies to any/all so2sat configs:

```
File "/opt/conda/lib/python3.8/runpy.py", line 194, in _run_module_as_main
return _run_code(code, main_globals, None,
File "/opt/conda/lib/python3.8/runpy.py", line 87, in _run_code
exec(code, run_globals)
File "/workspace/armory/scenarios/main.py", line 225, in <module>
run_config(
File "/workspace/armory/scenarios/main.py", line 142, in run_config
scenario = get(*args, **kwargs)
File "/workspace/armory/scenarios/main.py", line 134, in get
scenario = ScenarioClass(config, **kwargs)
File "/workspace/armory/scenarios/multimodal_so2sat_scenario.py", line 24, in __init__
self.perturbation_metrics = self.config["metrics"].pop("perturbation")
AttributeError: 'So2SatClassification' object has no attribute 'config'
```

To reproduce:

```
python -m armory run scenario_configs/eval1-4/so2sat/<CONFIG>
```

davidslater:
Fixes #1469

### DIFF
--- a/armory/scenarios/multimodal_so2sat_scenario.py
+++ b/armory/scenarios/multimodal_so2sat_scenario.py
@@ -12,7 +12,7 @@ from armory.utils.export import So2SatExporter
 
 
 class So2SatClassification(Scenario):
-    def __init__(self, *args, **kwargs):
+    def __init__(self, config, *args, **kwargs):
         if "attack_modality" not in kwargs.keys():
             raise ValueError("`attack_modality` must be defined for So2Sat scenario")
         attack_modality = (kwargs.pop("attack_modality") or "").lower()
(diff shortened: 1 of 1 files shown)


lcadalzo:
I was actually also working on a (slightly different) fix for this and pushed it [here](https://github.com/lcadalzo/armory/commit/89d0f6f62930d8cb066b4cbd97c6d6016326cf63). Thoughts?

armory/scenarios/multimodal_so2sat_scenario.py
```diff
@@ -12,7 +12,7 @@
 
 
 class So2SatClassification(Scenario):
-    def __init__(self, *args, **kwargs):
+    def __init__(self, config, *args, **kwargs):
```
lcadalzo, line 15: Given that the scenario is always instantiated here with `config` and `kwargs`, I thought having both `config` and `*args` was a little confusing since I don't think there will ever be any args passed besides `config`

davidslater: The base scenario doesn't have a `*` delimiter, so you can technically pass in the kwargs arguments as args.

davidslater: Perhaps a modification to add that?

davidslater:
I like the use of super in your PR - will add.

davidslater:
@lcadalzo thoughts now?

lcadalzo:
Looks good. The only remaining question I have, and this is unrelated to the fix in this PR, is do you know the purpose of the following line?

```
self.config["metric"]["perturbation"] = None
```

davidslater:
This prevents the MetricsLogger.from_config code from creating a perturbation metric that acts on the full input (since it's composed of EO and SAR and wouldn't make sense). Instead, the EO and SAR specific perturbation metrics are set up in `load_metrics`.

lcadalzo:
LGTM

Issue #1444, opened 2022-04-26
Comprehensive baseline configs for poisoning


swsuggs:
Create a subfolder of `scenario_configs` and fill it with baseline poisoning configs representing every combination of trigger, attack, dataset, and defense.

Defenses to include random filter, perfect filter, activation defense, and spectral signature defense.

Random filter and perfect filter need to be implemented.

This should replace the existing set of poorly-named gtsrb configs

PR #1484, opened 2022-05-03
Baseline poisoning configs

swsuggs:
To fulfill #1444.

Add random filter and perfect filter baseline defenses.

Add configs for the following threat models:
- gtsrb dlbd bullet holes
- gtsrb dlbd peace sign
- gtsrb clbd bullet holes
- gtsrb clbd peace sign
- gtsrb witches' brew
- cifar10 dlbd watermark
- cifar10 witches' brew

for each of the following baseline defenses (35 configs total):
- activation defense
- spectral signatures
- random filter
- perfect filter
- undefended


armory/scenarios/poison.py
```diff
@@ -225,47 +225,59 @@ def filter_dataset(self):
...
-            classifier_for_defense, _ = config_loading.load_model(defense_model_config)
-            log.info(
-                f"Fitting model {defense_model_config['module']}.{defense_model_config['name']} "
-                f"for defense {defense_config['name']}..."
-            )
-            # Flag to determine whether defense_classifier is trained directly
-            #     (default API) or is trained as part of detect_poisons method
-            if adhoc_config.get("fit_defense_classifier_outside_defense", True):
-                classifier_for_defense.fit(
+            if (
```
lcadalzo, line 229: Would it be possible to create a simple `PerfectFilter` module in `armory.art_experimental.poison_detection` that uses the same interface as the other poison filter defenses? If so, this would seem cleaner than checking for a specific kwarg in the scenario code

armory/art_experimental/poison_detection/random_filter.py
```diff
@@ -0,0 +1,49 @@
...
+        """
+        Create a :class:`.RandomFilterBaselineDefense` object with the provided classifier.
+        :param classifier: Model evaluated for poison.
+        :param x_train: dataset used to train the classifier.
+        :param y_train: labels used to train the classifier.
+        """
+        super().__init__(classifier, x_train, y_train)
+        self.n_data = len(y_train)
+
+    def evaluate_defence(self, is_clean, **kwargs):
```
lcadalzo, line 19: I'm lacking context on `evaluate_defence()`...is this a part of the ART API that all poisoning defenses must have this method?

If no, should we just omit it? I like raising a NotImplementedError if we expect someone might try using the method, but it's not actually important for our sake to implement it

davidslater: It's a `@abc.abstractmethod` in ART, so I think we need to implement it. The NotImplementedError seems right.

scenario_configs/poisoning_baselines/cifar10/dlbd/watermark/cifar10_dlbd_watermark_activation_defense.json
```diff
@@ -0,0 +1,98 @@
+{
```
lcadalzo, line 1: should the `poisoning_baselines` dir go in the `eval5` dir? At its current dir level, not sure it quite fits to sit alongside `eval5` and `eval1-4`

armory/art_experimental/poison_detection/random_filter.py
```diff
@@ -0,0 +1,49 @@
...
+    def detect_poison(self, **kwargs):
+        """
+        Selects data at random to label as poison.
+        :return: (report, is_clean_lst):
+                where report is None (for future ART compatibility)
+                where is_clean is a list, where a 1 at index i indicates that x_train[i] is clean,
+                    and a 0 indicates that x_train[i] is detected as poison.
+        """
+
+        expected_pp_poison = kwargs.get("expected_pp_poison", None)
```
davidslater, line 33: Can you make this a keyword argument?  and remove the **kwargs? 

```
def detect_poison(self, expected_pp_poison=None):
    ...
```

armory/art_experimental/poison_detection/random_filter.py
lcadalzo: Got it. In that case, I agree the NotImplementedError makes sense

As an aside, not sure why it didn't err then when I ran the config after deleting the method

armory/scenarios/poison.py
swsuggs: Look at [5747c50](https://github.com/twosixlabs/armory/pull/1484/commits/5747c5041f548e4a94af58511514c4406ad78280) to see what I had before

lcadalzo: if we do keep the check, I think
```
if defense_config["kwargs"].get("perfect_filter", False):
```
would be cleaner than 

```
 if (
                "perfect_filter" in defense_config["kwargs"]
                and defense_config["kwargs"]["perfect_filter"] is True
            )
```

armory/utils/config_schema.json
```diff
@@ -87,8 +87,6 @@
             },
             "required": [
                 "kwargs",
-                "module",
-                "name",
```
davidslater, line 91: Looking at this again, I think that it might be better to leave these as required fields, and simply set them to `null` in the PerfectFiltering configs.

armory/art_experimental/poison_detection/random_filter.py
swsuggs: It did throw some kind of error for me before I implemented it

armory/scenarios/poison.py
davidslater: The False is unnecessary, as `.get("perfect_filter")` will default to None, which is false.

armory/utils/config_schema.json
swsuggs: There's some regex matching that doesn't allow `null`.  We could rewrite to allow null as an option?

armory/scenarios/poison.py
lcadalzo: I see, @swsuggs disregard my comment then. I suppose there's not a great way to propagate that info to the defense class without adding another check for a specific value

armory/utils/config_schema.json
davidslater: Can you try the empty string `""` instead of `null`?

scenario_configs/poisoning_baselines/cifar10/dlbd/watermark/cifar10_dlbd_watermark_activation_defense.json
swsuggs: I can reason either way.  These could be useful in future evals or outside the gard evaluation program.  On the other hand, they're created especially for convenience in evaluating this specific round of threat models.   I'm happy to do whatever you think makes more sense

davidslater: If there are particular ones that would be good to link from the base directory, though, let's add a link or 2.

armory/art_experimental/poison_detection/random_filter.py
swsuggs: yep

armory/scenarios/poison.py
```diff
@@ -225,47 +225,59 @@ def filter_dataset(self):
...
+                "perfect_filter" in defense_config["kwargs"]
+                and defense_config["kwargs"]["perfect_filter"] is True
+            ):
```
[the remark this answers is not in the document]
swsuggs: much better

armory/utils/config_schema.json
swsuggs: It does not accept the empty string either

davidslater: That is weird. I guess do `"null"`. That should work. Technically, it could be a module and a name. :)

swsuggs: Check out the latest commit, where config validation allows `null` for module and name (although this affects the module field anywhere, not just under defense).  What are your thoughts?

swsuggs: Oops, your last message didn't show up before I posted the above

swsuggs: I'll switch it back and use `"null"`

davidslater: Yeah, let's use `"null"`. The modification of the type checking will have broader impact than I would like.

davidslater (approved):

PR #1486, opened 2022-05-03
WIP: DO NOT MERGE rudimentary example of using hub.record() to save images

lcadalzo:
See #1285. I've added temporary code to the baseline model in `cifar.py` which rotates the input and then `hub.records()` the rotated image. This is just meant as a starting point to determine how we can use `armory.instrument` and the export samples feature to allow for arbitrary saving of images


lcadalzo:
I think using probes and Exporters as sinks, as shown in #1488, likely makes more sense

lcadalzo:
Closing in favor of #1488

PR #1488, opened 2022-05-04
ExportMeters + DAPRICOT bug fixes

lcadalzo:
Addresses #1285, #1498, #1509, and #1510. This approach is an alternative to #1486 that I think could make more sense


davidslater:
You could do this with a Meter like so:
```
from armory.utils.export import ImageClassificationExporter                              
                                                                                         
class ImageExportMeter(Meter):                                                           
    def __init__(self, name, metric_arg_name, base_output_dir):                          
        super().__init__(self, name, lambda x: x, metric_arg_name)                       
        self.base_output_dir = base_output_dir                                           
        self.exporter = ImageClassificationExporter()                                    
                                                                                         
    def measure(self, clear_values=True):                                                
        self.is_ready(raise_error=True)                                                  
        batch_num, value = self.arg_batch_indices[0], self.values[0]                     
        probe_variable = self.arg_index.values()[0]                                      
        batch_size = value.shape[0]                                                      
        for idx in range(batch_size):                                                    
            sample = self.exporter.get_sample(value[idx])                                
            self.exporter.save_sample(                                                   
                fname=f"{self.base_output_dir}/{probe_variable}_batch_{batch_num}_ex_{idx}",
                sample=sample,                                                           
            )                                                                            
        if clear_values:                                                                 
            self.clear()                                                                 
        self.never_measured = False                                                      
                                                                                         
                                                                                         
hub.connect_meter(ImageExportMeter("exporter", "cifar_model.x_rotated", "tmp_dir"),      
        use_default_writers=False)
```

lcadalzo:
I'd like to refactor export.py to unify the way examples are exported in the following 2 cases:

1. the standard way, i.e. setting `"export_batches"` in the config file
2. using a probe/meter as is done in this PR

But before doing so, I'd like to nail down how we're accomplishing case 2

davidslater:
I like the ExportMeter. I suggest that we pass in an `Exporter` instantiated object (instead of a class) - that should simplify the logic a bit. Instead of creating subclasses, then, we can just pass in different `Exporter` objects. 

Take a look at the `Meter.set` method as well - that's where the hub passes in batch information - and we probably shouldn't have the hub internal to the meter (unless passed in as an arg to __init__). The stage information should be filtered by the arg names, and so you can get it that way.

lcadalzo:
One issue I'm hitting is parsing probe values to be passed in as `export()` kwargs. While `"scenario.x"` or `"scenario.x_adv"` is consistently passed as a positional arg, the scenarios with boxes also need to pass `y` and `y_pred` kwargs

Ideally the probe value names would align 1:1 with `export()` kwarg names, but this isn't the case. Both `"scenario.y_pred"` and `"scenario.y_pred_adv"` need to be passed to the `y_pred` kwarg of the export function, so `probe_value.split(".")[-1]` will work for the former but not the latter

The previous commit deals with this my mapping the probe name`"y_pred_adv"` to the export kwarg name "`y_pred"` but this feels a bit hacky

lcadalzo:
`ExportMeters` are always created by the scenario, but they aren't used if `export_batches` isn't set in the config (or if it is set, but to `False` or `0`). This currently results in a warning as such:

```
2022-05-13 15:05:49 51s WARNING  armory.instrument.instrument:finalize:549 Meter 'x_exporter' was never measured. The following args were never set: []
```

The list is empty because the args are indeed being set/probed, but there's specific logic in the `ExportMeter` to do nothing since `export_batches` is 0. I think it's undesirable to have this warning occurring in the default case.

We could modify `finalize()` of `ExportMeters` to not throw the warning, but I do think the warning would be helpful for someone integrating an `ExportMeter` into their model code e.g. Another alternative is to create a `self.warn` attribute that is by default `True` but set to `False` by the scenario code. This way, the default `ExportMeters` don't warn when not used but custom ones would. If we do add a warning, we should improve the message (i.e. make it specific to `ExportMeters`)

lcadalzo:
Current issues/drawbacks/questions that I still see here:

1. The basename of each outputted image file is created [here](https://github.com/lcadalzo/armory/blob/1285-export-preprocessed-data-via-probes/armory/utils/export.py#L589). If no stage filter is set for the meter (as is the case [here](https://github.com/lcadalzo/armory/blob/1285-export-preprocessed-data-via-probes/armory/baseline_models/pytorch/cifar.py#L27)), then `measure()` will be called multiple times during the `attack` stage for the same `batch_num`. Note that this also occurs if you set a stage filter of `attack`. Consequentially, the successive `measure()` calls are currently just overwriting the image repeatedly during the various attack iters since the string ` f"batch_{batch_num}_ex_{batch_idx}_{probe_variable}"` isn't changing. There are various ways we could avoid overwriting (e.g. have a class attribute that keeps track of how many times `measure()` has been called) but it's unclear what the desired behavior should be
2. The current code sets the rotated images to be outputted to `tmp_dir/`, see [here](https://github.com/lcadalzo/armory/blob/1285-export-preprocessed-data-via-probes/armory/baseline_models/pytorch/cifar.py#L23). It would be convenient if there was an easy way to feed in the scenario output directory here, but I don't know of a good way to do that. The only hacky way I can think to do so is access the `ExportMeters` from the hub and access their `.exporter.output_dir` as such: 
```
(Pdb) hub.meters[-2].exporter.output_dir
'/armory/outputs/2022-05-16T203248.015341/saved_samples_1652815295.4693105'
```

3. Do we want to move `ExportMeters` out of `armory.utils.export`?

davidslater:
Hmm, let start in reverse order:
3) Perhaps? I think that the entire `armory.utils.export` seems to fit better in instrumentation, anyhow. Thoughts on just moving the entire file to `armory.instrument.export`?

davidslater:
2) It seems like the hub would be a good place to store output directory info. We could add a "set_output_dir" and method to it, and have that set at the Scenario `__init__` or in `Scenario._set_output_dir()`. It could default, potentially, to `armory.paths.runtime_paths().output_dir` or something like that.

Then, when you set up the exporter, you can do something like:
```
probe = get_probe("cifar_model")
hub = get_hub()
exporter = ImageClassificationExporter(os.path.join(hub.get_output_dir(), "export"))
export_meter = ExportMeter(
    "rot_exporter",
    exporter,
    "cifar_model.x_rotated",
)
hub.connect_meter(export_meter, use_default_writers=False)
```


You could also have a convenience method in the Hub like:
```
def output_subdir(self, subdir):
    return os.path.join(self.output_dir, subdir)
```
So that your call above becomes:
```
exporter = ImageClassificationExporter(hub.output_subdir("export"))
```

davidslater:
Regarding 1), maybe we make it a bit more explicit. We could have a kwarg in the init that specifies how this is handled:
```
class ExportMeter(Meter):
    def __init__(
        self, name, exporter, x_probe, y_probe=None, y_pred_probe=None, max_batches=None, overwrite_mode="error",
    ):
    """
    overwrite_mode: if file exists on write, then:
        "error" - raise error
        "increment" - append an underscore and a number to existing files, incrementing the number until the target file does not exist
        "overwrite" - overwrite file
        "overwrite_warn" - overwrite file and log a warning
        "skip" - don't write
        "skip_warn" - don't write but log a warning
    """
```

That may be more options than we want, however.

davidslater:
Overall, though, I really do like this update.

lcadalzo:
CI tests are now all passing, including tests I've added for the new classes introduced in this PR. Still leaving as WIP, though, since there's temporary code in `cifar.py` that needs to eventually be removed and the docs/notebooks still need to be updated too.

lcadalzo:
@swsuggs do you mind testing that we're still correctly exporting what the poisoning team desires for the witches brew scenario?

lcadalzo:
> I'll revisit this and try and remove the dependency where load_export_meters() needed to be called before load_model()

I had actually already removed this dependency when I moved the logic for setting the export dir from the exporter to the scenario `__init__()` in [this](https://github.com/twosixlabs/armory/pull/1488/commits/43a99ad2b3a06989e62f7b27ecefd65b610179af#diff-84f46a19b799a62b1ee5e34c4de63be6e6fa1ed50888ab3eea1f07f45d429a4c) commit. Thus, in my previous commit, I reset the order of pieces in `load()` to the same order as 0.15.0

armory/instrument/instrument.py
```diff
@@ -311,6 +312,20 @@ def __init__(self):
...
+        self.output_dir = name
+
+    def get_output_dir(self):
+        return self.output_dir
+
+    def _set_export_dir(self, name):
+        self.export_dir = name
+
+    def get_export_dir(self):
+        return self.export_dir
```
davidslater, line 328: Does it make more sense to enforce export_dir to be a subdir of output_dir? Otherwise, if you're in the container, you're likely going to make an error in pathing that doesn't get saved.

Then this section would become:
```
        self._set_export_dir("saved_samples")

    def _set_export_dir(self, output_subdir):
        self.export_dir = os.path.join(armory.paths.runtime_paths().output_dir, output_subdir)

    def get_export_dir(self):
        return self.export_dir
```

lcadalzo: just thinking about whether there could be a situation where you want to export artifacts to somewhere other than `~/.armory/outputs`...but I think it's cleanest to have everything in one place so I like the change.

I do think the logic of setting export_dir to `f"saved_samples_{self.time_stamp}"` when `"saved_samples"` already exists needs to remain, so that outputs from subsequent runs don't get jumbled up / overwrite eachother

lcadalzo:
> or at least could replace batch_0_ex_0_scenario.x.png with non_trigger_batch_0_ex_0_scenario.x.png,
but if it's easier to be consistent with the output from other scenarios, then that's ok.

The format of ` f"batch_{batch_num}_ex_{batch_idx}_{probe_variable}"` is intended to create a ID of sorts that should be unique for each example being saved. This is defined in the `measure()` method of the `ExportMeter`. I'd prefer not to create another `ExportMeter`-derived class specific to witches brew

On the contrary, the witches brew scenario code directly uses an `Exporter` (as opposed to an `ExportMeter`) to save the image with basename `f"trigger_batch_{self.i}_ex_{batch_idx}"`. This would be much easier to modify, if we wanted. I opted to have the string start with `trigger*` so that all the trigger images would be next to eachother but you could still reference the batch/ex numbers

armory/instrument/instrument.py
davidslater: Yeah, I agree.

armory/instrument/instrument.py
```diff
@@ -312,17 +312,17 @@ def __init__(self):
...
     def get_output_dir(self):
         return self.output_dir
 
-    def _set_export_dir(self, name):
-        self.export_dir = name
+    def _set_export_dir(self, output_subdir):
+        self.export_dir = f"{self.output_dir}/{output_subdir}"
 
     def get_export_dir(self):
         return self.export_dir
```
davidslater, line 328: I see a potential edge case here: If you set the export directory and then set the output directory, the export directory won't be a subdir of the output dir.

davidslater: Maybe we internally store export_subdir, and then reapply it when output directory is changed.

lcadalzo:
The previous CI failure appears unrelated to this PR, re-running CI [here](https://github.com/lcadalzo/armory/actions/runs/2386622078)

PR #1490, opened 2022-05-04
Compute metrics

### LINKED ISSUE #881: Better documentation of using computational metrics

davidslater:
Fixes #881

- [x] Add tests
- [x] Figure out how to add `null` as a valid part of the `"profiler_type"` part of json config validation.


davidslater:
You can test by adding `"profiler_type": "Basic",` or `"profiler_type": "Deterministic",` to the metrics config, like:
```
    "metric": {
        "means": true,
        "perturbation": "linf",
        "profiler_type": "Basic",
        "record_metric_per_sample": false,
        "task": [
            "categorical_accuracy"
        ]
    },
```

Note that "Deterministic" outputs a LOT of content.

swsuggs:
This seems to work for the enum thing:
```
"profiler_type": {
          "enum": [
              "Basic",
              "Deterministic",
              null
          ],
          "type": ["string", "null"]
      },
```
not sure if it would adversely affect anything else

davidslater:
Thanks, @swsuggs , that worked

davidslater:
@swsuggs ready for a review

swsuggs:
maybe a naive question... but what use is all that output from the deterministic profiler?  Would it be helpful to also provide some simple tool for parsing and filtering and sorting it?  For example, if as a performer I wanted to break down attack time by my model's components vs other armory or art components, is there a way to do that?

davidslater:
I've used it for finding bottlenecks. Typical usage would be to just glance through the list and find the worst offenders, then dig into those specific functions to see what the cause was. I would consider it more of a "debug" mode type of tool. I've never actually parsed it before, just printed to screen.

davidslater:
It's python cProfile standard library output: https://docs.python.org/3/library/profile.html
So there's probably a standard tool for parsing if people want that. I don't think it makes sense to put it into our codebase unless we see a more specific use case.

swsuggs:
When I run `pytest tests/unit/test_compute_metrics.py`, I get 
```
tests/unit/test_compute_metrics.py:7: in <module>
    from armory.metrics import compute
E   ModuleNotFoundError: No module named 'armory.metrics'
```
But I can run the scenarios.  So I'm not sure if I'm doing something wrong there

swsuggs:
Building a new docker container on the current commit fixed that.
I can run tests and configs.
LGTM

Issue #993, opened 2021-02-19
partition utils.metrics for better discovery and composition


mwartell:
Per Slater:

Consider changing our metrics setup a bit and migrating armory.utils.metrics to its own top level space armory.metrics with modules for different types, like perturbation, task, and compute. It’s a bit of a messy module at the moment.

It would be nice to make metrics discoverable and addressable, and automatically enable batchwise or elementwise usage
Like `metrics.perturbation.element.l2` or `metrics.perturbation.batch.linf` or `metrics.task.batch.categorical_accuracy`.

davidslater:
This makes nearly all of the metrics element-wise, which I think will greatly simplify testing. Batchwise versions of metrics are automatically generated.

PR #1493, opened 2022-05-06
Perturbation metrics

davidslater:
Note that this builds off of the compute metrics PR.

Partially addresses #993 and #989 (will need to update `task` metrics similarly)

- [x] Migrate existing tests
- [x] Add tests to cover all metrics
- [x] Update docs


davidslater:
When writing tests, I had a hard time understanding and parsing `image_circle_patch_diameter`, so I rewrote it, fixed a minor bug (for rectangular images, if you transpose them in the original method, you can get different results), and expanded the docs. @lcadalzo

docs/metrics.md
```diff
@@ -30,42 +31,88 @@ When `means` is true, the average value for the given metric is also recorded.
...
+```
+perturbation.batch.l1([0, 0, 0], [1, 1, 1])
+# array([1., 1., 1.])
+perturbation.element.l1([0, 0, 0], [1, 1, 1])
+# 3.0
+```
+Metric outputs are numpy arrays or scalars.
+
+
+### Task Metrics
```
lcadalzo, line 101: maybe this is more relevant to the task metrics PR, but we're missing metrics in this table

tests/unit/test_perturbation_metrics.py
```diff
@@ -0,0 +1,181 @@
...
+
+import pytest
+import numpy as np
+
+from armory.metrics import perturbation
+
+pytestmark = pytest.mark.unit
+
+
+def test_lp_norms():
```
lcadalzo, line 15: I think it'd make sense to also test `perturbation.element` for lp norms as well

docs/metrics.md
davidslater: Planning to address those missing task metrics in that PR.

armory/metrics/perturbation.py
```diff
@@ -0,0 +1,340 @@
...
+def numpy(function):
+    """
+    Ensures args (but not kwargs) are passed in as numpy vectors
+    It casts them to to a common data type (complex, float, int) if possible
+    And it ensures that they have the same shape
+    """
+
+    @functools.wraps(function)
+    def wrapper(x, x_adv, **kwargs):
+        # TODO: cast from Torch tensor and TF tensor to numpy?
```
lcadalzo, line 117: is the TODO here bc we were worried that `np.asarray` might not work for torch and tf? If it does, should we remove the comment (otherwise, should we address it)?

docs/metrics.md
```diff
@@ -30,42 +31,88 @@ When `means` is true, the average value for the given metric is also recorded.
...
+from armory.metrics import perturbation
+print(peturbation.batch)
+# ['image_circle_patch_diameter', 'l0', 'l1', 'l2', 'linf', 'max_image_circle_patch_diameter', 'max_l0', 'max_l1', 'max_l2', 'max_linf', 'mean_image_circle_patch_diameter', 'mean_l0', 'mean_l1', 'mean_l2', 'mean_linf', 'snr', 'snr_db', 'snr_spectrogram', 'snr_spectrogram_db']
+print(perturbation.element)
+# ['image_circle_patch_diameter', 'l0', 'l1', 'l2', 'linf', 'max_image_circle_patch_diameter', 'max_l0', 'max_l1', 'max_l2', 'max_linf', 'mean_image_circle_patch_diameter', 'mean_l0', 'mean_l1', 'mean_l2', 'mean_linf', 'snr', 'snr_db', 'snr_spectrogram', 'snr_spectrogram_db']
+```
+Currently, all perturbation metrics have element-wise and batch-wise versions, though the config assumes that the batch version is intended.
+For instance:
+```
+perturbation.batch.l1([0, 0, 0], [1, 1, 1])
```
lcadalzo, line 93: should we mention in the docs that `perturbation.element.<metric>()` == `perturbation.<metric>()`?

lcadalzo: i.e.:
```
>>> perturbation.l2(x, x_adv) == perturbation.element.l2(x, x_adv)
True

```

tests/unit/test_perturbation_metrics.py
davidslater: Yeah, that makes sense. I think it's probably unnecessary for all the norms, as the batch norms just combine results from the element norms, but it would be good to have something that tests them directly.

armory/metrics/perturbation.py
davidslater: I just checked, and it looks like `np.asarray` works fine for torch and tf tensors. So I'm going to remove the TODO.

davidslater: Done.

tests/unit/test_perturbation_metrics.py
davidslater: Done.

docs/metrics.md
davidslater: I don't think so. That is the case for perturbation metrics, but is not the case for task metrics, which often require all inputs (and therefore only have a `batch` version). I wouldn't want that to be an implicit assumption people make.

lcadalzo:
LGTM

PR #1495, opened 2022-05-07
Task metrics

### LINKED ISSUE #989: metrics should handle a batch of numpy object array

### LINKED ISSUE #993: partition utils.metrics for better discovery and composition

### LINKED ISSUE #1494: Enable easier integration of user-defined metrics
Instead of a hard enumeration of metrics, as currently defined in the config_schema, allow additional metrics to use a standard import mechanism, such as "my.module.submodule.function". This could be handled in a fairly sane way. Basically attempt this in order:
```python
import importlib                                                                         
                                                                                         
                                                                                         
def get_metric(string):                                                                  
    tokens = string.split(".")                                                           
    if not all(token.isidentifier() for token in tokens):                                
        raise ValueError(f"{string} is not a valid '.'-separated set of identifiers")    
    if len(tokens) < 2:                                                                  
        raise ValueError(f"{string} not a valid module and function path")               
                                                                                         
    errors = []                                                                          
    for i in range(len(tokens) - 1, 0, -1):                                              
        module_name = ".".join(tokens[:i])                                               
        metric_name = ".".join(tokens[i:])                                               
        try:                                                                             
            module = importlib.import_module(module_name)                                
        except ImportError:                                                              
            errors.append(f"ImportError: import {module_name}")                          
            continue                                                                     
        try:                                                                             
            return getattr(module, metric_name)                                          
        except AttributeError:                                                           
            errors.append(                                                               
                f"AttributeError: module {module_name} has no attribute {metric_name}"   
            )                                                                            
                                                                                         
    error_string = "\n".join(errors)                                                     
    raise ValueError(                                                                    
        f"Could not import metric {string}. "                                            
        "The following errors occurred:\n"                                               
        f"{error_string}"                                                                
    )
```

### LINKED ISSUE #1496: Migrate poisoning metrics into `armory.metrics` from `armory.utils.poisoning` and refactor
Some of the functionality from `armory.utils.poisoning` should probably go more directly into poisoning scenarios.

davidslater:
Fixes #993
Fixes #989
Fixes #1494
Fixes #1496

This splits the remaining metrics in `armory.utils.metrics` into tasks and statistical metrics (the latter mostly used for poisoning), in `armory.metrics.task` and `armory.metrics.statistical`, respectively.

It also moves `armory.utils.poisoning` to `armory.metrics.poisoning` and modifies it to fit the current paradigm.

Note that this builds off of the perturbation metrics PR #1493.

TODO:
- [x] Ensure parity
- [x] Expand test suite
- [x] Improve loading functionality
- [x] Update docs
- [x] Refactor poisoning to fit new model


davidslater:
@swsuggs I've refactored most of the poisoning fairness work. I've moved `get_majority_mask` to `armory.metrics.statistical` (which still needs tests for everything - I've stubbed them out in `tests/unit/test_statistical_metrics.py`).

I've refactored a lot of `armory.metrics.poisoning` - I now have most of the ExplanatoryModel functionality wrapped up in a single class. The main thing that needs some thought and work is refactoring `FairnessMetrics.add_cluster_metrics(...)`. If there is a good way to separate the two metric results, that would be ideal. Also, I think it makes sense to move the hub and log recording into the poisoning scenarios (and probably the ExplanatoryModel predictions as well). Your thoughts here would be helpful.

armory/metrics/poisoning.py
```diff
@@ -0,0 +1,263 @@
...
+        # get majority ceilings on unpoisoned part of train set
+        poisoned_mask = np.zeros_like(y_poison, dtype=bool)
+        poisoned_mask[poison_index.astype(np.int64)] = True
+        x_unpoisoned = x_poison[~poisoned_mask]
+        y_unpoisoned = y_poison[~poisoned_mask]
+
+        activations = self.explanatory_model.get_activations(x_unpoisoned)
+        majority_mask_unpoisoned, majority_ceilings = get_majority_mask(
+            activations,
+            y_unpoisoned,  # TODO: check
```
swsuggs, line 203: what needs to be checked here and at 214?

davidslater: I'm not sure if I'm passing in the right value. Or if it needs to be structured differently.

davidslater:
Yeah, go ahead and make a PR into my branch with your fixes for `statistical.py` as well.

davidslater:
@swsuggs if you can fill out the tests in `tests/unit/test_poisoning_metrics.py`, that would be helpful as well.

armory/metrics/poisoning.py
```diff
@@ -0,0 +1,263 @@
...
+            x = self.preprocess(x)
+            activation, _ = self.explanatory_model(x)
+            activations.append(activation.detach().cpu().numpy())
+        return np.concatenate(activations)
+
+    @staticmethod
+    def _preprocess(
+        x, resize_image=True, size=(224, 224), resample=Image.BILINEAR, device=DEVICE
+    ):
+        if isinstance(x.dtype, np.floating):
```
swsuggs, line 109: Not sure exactly what the input is expected to be at this point, but this returns false when x.dtype == np.float32 or np.float64.

It appears that `x.dtype == np.floating` returns true if `x.dtype == np.float64` but not if `x.dtype == np.float32`.

I believe armory data is loaded as np.float32.

armory/metrics/poisoning.py
```diff
@@ -0,0 +1,263 @@
...
+        activations = []
+        with torch.no_grad():
+            x = self.preprocess(x)
+            activation, _ = self.explanatory_model(x)
+            activations.append(activation.detach().cpu().numpy())
+        return np.concatenate(activations)
+
+    @staticmethod
+    def _preprocess(
+        x, resize_image=True, size=(224, 224), resample=Image.BILINEAR, device=DEVICE
```
swsuggs, line 107: why do we want to resize by default?

armory/metrics/poisoning.py
davidslater: The input should either be `np.uint8` or any numpy floating type `np.float32`, `np.float64`, etc.

armory/scenarios/poison.py
```diff
@@ -205,14 +206,13 @@ def poison_dataset(self):
         self.record_poison_and_data_info()
 
     def record_poison_and_data_info(self):
-
         self.n_poisoned = int(len(self.poison_index))
         self.n_clean = (
-            len(self.y_clean) - self.n_poisoned
+            len(self.y_poison) - self.n_poisoned
         )  # self.y_clean is the whole pre-poison train set
```
swsuggs, line 212: The comment is no longer relevant given you're using self.y_poison instead of self.y_clean

swsuggs:
I think there are additional simplifications to be made in some other variables.  For example, 295 - 297:
```
is_dirty_mask = np.ones_like(self.y_clean)
is_dirty_mask[self.indices_to_keep] = 0
self.is_dirty_mask = is_dirty_mask
```
can become `self.is_dirty_mask = ~self.indices_to_keep`.  (Unless it's critical that is_dirty_mask is 1/0 while indices_to_keep is Boolean, but I don't think it is.)

I realize this wasn't part of this PR.  But I wouldn't complain if you took a moment to make the variable names more consistent/self-explanatory.  For instance, `is_dirty_mask` is the complement of `indices_to_keep`.  Perhaps calling the former `indices_to_remove` or the latter `is_clean_mask` would reflect that better.  And better than "is_clean..." or "is_dirty..." might be "detected_clean..." or "detected_dirty..." so it's clear that's according to the defense, and not ground truth.

armory/scenarios/poison.py
```diff
@@ -518,28 +532,116 @@ def evaluate_current(self):
...
+        return self.majority_mask_test_set
+
+    def record_bias_results(self, chi2_spd, class_ids, record_prefix, log_prefix):
+        for class_id in class_ids:
+            chi2, spd = chi2_spd[class_id]
+            self.hub.record(
+                f"{record_prefix}bias_chi^2_p_value_{str(class_id).zfill(2)}", chi2
+            )
+            self.hub.record(f"{record_prefix}bias_spd_{str(class_id).zfill(2)}", spd)
+        log.info(
```
swsuggs, line 600: looks like these log lines should be indented under the for loop

armory/scenarios/poison.py
```diff
@@ -518,28 +532,116 @@ def evaluate_current(self):
...
+            self.test_y,
+            self.majority_mask_test_set,
+            self.test_y == self.test_y_pred_class,
+            self.test_set_class_labels,
+        )
+
+        self.record_bias_results(
+            chi2_spd, self.test_set_class_labels, "model_", "Model "
+        )
+        if self.is_filtering_defense:
```
swsuggs, line 630: self.is_filtering_defense has not been defined.

swsuggs: should be `self.use_filtering_defense`

armory/scenarios/poison.py
davidslater: Removed

armory/scenarios/poison.py
```diff
@@ -234,7 +234,7 @@ def filter_dataset(self):
 
             if defense_config["kwargs"].get("perfect_filter"):
                 log.info("Filtering all poisoned samples out of training data")
-                indices_to_keep = np.ones_like(self.y_poison, dtype=np.bool_)
+                indices_to_keep = np.ones_like(self.y_poison, dtype=bool)
```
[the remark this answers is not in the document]
davidslater: Done.

armory/scenarios/poison.py
davidslater: Removed.

armory/scenarios/poison.py
davidslater: Done.

davidslater:
@swsuggs ready for another review

swsuggs:
lgtm

davidslater:
Done

davidslater:
@swsuggs can you do a quick review of `poison.py` and `multimodal_so2sat_scenario.py`? I had to manually merge those two.

armory/scenarios/poison.py
```diff
@@ -508,28 +506,119 @@ def evaluate_current(self):
...
-                self.poison_index,
-                self.indices_to_keep,
-                # Get unpoisoned test set
-                config_loading.load_dataset(
-                    self.config["dataset"],
-                    split="test",
-                    num_batches=None,
-                    **self.dataset_kwargs,
-                ),
+        if self.num_export_batches > self.sample_exporter.saved_batches:
```
swsuggs, line 509: looks like lucas's code removed this block (through line 517).  self.sample_exporter has no attribute 'saved_batches'

swsuggs: But when I remove it (locally) and run, it exports all the images, not just up to export_batches from the config.  so there's something else missing

swsuggs: @lcadalzo may be able to find the issue faster than me, I'm not yet very familiar with the new exporting code

lcadalzo: @swsuggs after removing this block and running with `"export_batches": 2`, my output looks right:

```
I have no name!@d16b7467e2a5:/workspace$ ls /armory/outputs/2022-05-26T191142.588735/saved_samples_1653592697.6484163/
batch_0_ex_0_scenario.x.png  batch_0_ex_0_scenario.x_adv.png  batch_1_ex_0_scenario.x.png  batch_1_ex_0_scenario.x_adv.png  predictions.pkl
```

Which config are you using? What did you set `"export_batches"` to?

lcadalzo: I also changed batch_size to 1

swsuggs: It was my error.  I ran a gtsrb dlbd config, but I was thinking of witches' brew which automatically uses a test batch_size of 1.  My batch size was still 512 after all.

swsuggs: Curious what config you used though.  When I set batch_size to 1 for gtsrb dlbd, I get an error:
`ValueError: Expected more than 1 value per channel when training, got input size torch.Size([1, 300])`

lcadalzo: I was also using gtsrb dlbd, but for speed's sake I commented out the `fit()` call in the scenario code. When I uncomment it, I see the error. Weird bc the error goes away if I set batch_size to 2. I'll create an issue for it

lcadalzo:
[This](https://github.com/davidslater/armory/blob/task-metrics/armory/scenarios/poison.py#L509-L517) block of code should be removed from `poison.py`. `multimodal_so2sat_scenario.py` looks ok and runs without error

davidslater:
Thanks guys! Made the fix, and I'll merge once the tests pass.

PR #1499, opened 2022-05-09
Writer updates

### LINKED ISSUE #1485: Add type checking to Writer classes?
Say a user adds a `hub.record("name", x_preprocessed)` , perhaps naively, to their code where `x_preprocessed` is a numpy array. This gets passed along to a `ResultsWriter` object, an element of `self.default_writers`, which then ultimately writes each  entire image array to output JSON.

To give an example of how unwieldy this gets: using the xView dataset with `--skip-attack --num-eval-batches 10`, this takes the config output length from ~100 lines (8KB) to ~ 6.5 million (230MB). 

Perhaps a related but separate aspect of this is whether we want `record()` to always push all records to all default writers.

davidslater:
Fixes #1485

### DIFF
--- a/armory/instrument/config.py
+++ b/armory/instrument/config.py
@@ -29,6 +29,7 @@ class MetricsLogger:
         include_adversarial=True,
         include_targeted=True,
         record_metric_per_sample=False,
+        max_record_size=2**20,
     ):
         """
         task - single metric or list of metrics
@@ -39,6 +40,7 @@ class MetricsLogger:
         include_benign - whether to include benign task metrics
         include_adversarial - whether to include adversarial task metrics
         include_targeted - whether to include targeted task metrics
+        max_record_size - maximum number of bytes in a record (for ResultsWriter)
         """
         self.task = task
         self.task_kwargs = task_kwargs
@@ -70,7 +72,9 @@ class MetricsLogger:
                 perturbation, use_mean=means, record_final_only=self.record_final_only
             )
 
-        self.results_writer = ResultsWriter(sink=self._sink)
+        self.results_writer = ResultsWriter(
+            sink=self._sink, max_record_size=max_record_size
+        )
         get_hub().connect_writer(self.results_writer, default=True)
 
         self.metric_results = None
--- a/armory/instrument/instrument.py
+++ b/armory/instrument/instrument.py
@@ -660,18 +660,37 @@ class ResultsWriter(Writer):
     Write results to dictionary
     """
 
-    def __init__(self, sink=None):
+    def __init__(self, sink=None, max_record_size=None):
         """
         sink is a callable that takes the output results dict as input
             if sink is None, call get_output after close to get results dict
+        max_record_size - if not None, the maximum size in bytes for a single record
+            Records exceeding that size log a warning and are dropped
         """
         super().__init__()
         self.sink = sink
         self.records = []
         self.output = None
+        if max_record_size is not None:
+            if max_record_size < -1:
+                raise ValueError(f"max_record_size {max_record_size} cannot be < -1")
+            if json_utils is None:
+                raise ValueError("Cannot import numpy. Set max_record_size to None")
+            max_record_size = int(max_record_size)
+        self.max_record_size = max_record_size
 
     def _write(self, name, batch, result):
-        self.records.append((name, batch, result))
+        record = (name, batch, result)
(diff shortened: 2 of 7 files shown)


armory/instrument/instrument.py
```diff
@@ -660,18 +680,37 @@ class ResultsWriter(Writer):
...
 
     def _write(self, name, batch, result):
-        self.records.append((name, batch, result))
+        record = (name, batch, result)
+        if self.max_record_size is not None:
+            try:
+                json_utils.check_size(record, self.max_record_size)
+            except ValueError:
+                log.warning(
+                    f"record ({name}, {batch}, ...) size > "
```
lcadalzo, line 709: might not be super clear that the integer in the middle is `batch` when it's logged:

```
2022-05-10 14:54:24 13s WARNING  armory.instrument.instrument:_write:708 record (x_preprocessed, 9, ...) size > max_record_size 1048576 Dropping.

```

maybe consider `f"batch={batch}"` ?

armory/instrument/config.py
```diff
@@ -29,6 +29,7 @@ def __init__(
         include_adversarial=True,
         include_targeted=True,
         record_metric_per_sample=False,
+        max_record_size=2**20,
```
lcadalzo, line 32: my method of testing this PR was to do the same thing that brought the issue to my attention: add a `hub.record()` in the model's `forward()` that records an image and see whether the output JSON file size balloons.

When I ran this test (using CIFAR), I noticed that the output JSON was still blowing up in size. I realized that this was simply because a `(1, 3, 32, 32)` array is still below the threshold. So for other scenarios with bigger inputs, it'll ignore the arrays (since I imagine they'll exceed `2 ** 20` bytes), but it may be worth considering a stricter threshold. Not sure how arbitrary `2 ** 20` was

davidslater: I put something that was about 4x lower than the `(200, 200, 3)` images. My main concern with pushing too small is making sure that complex records don't get unintentionally removed.

I figured that 1 MB per record would mean that even if you pushed in a bunch of 1 MB records, you'd probably wouldn't get too large of a file (unless you did ~10k records at the 1 MB limit). For CIFAR, the json would still be only ~300 MB if you output the entire test set.

armory/instrument/instrument.py
davidslater: Updated.

armory/instrument/config.py
davidslater: If you have a different number that you think works better, happy to add it. But I'm not sure about going so small as needed to catch the CIFAR (let alone MNIST) images. A metric record that's outputting something complex like confusion matrices or a sweep of (tpr, fpr, threshold) values for a ROC Curve could easily go above 3x32x32 = 3072 values.

lcadalzo: I'll point out that, in the scenario when `hub.record()` is placed inside the model's `forward()`, this gets called multiple times per attack so the file size will grow well beyond 300MB which I've tested/confirmed (hence why it's better to accomplish this user goal with a meter and stage filters, but the point here is that there's nothing stopping a user from using `record()`.

Nonetheless, I agree with your point about being careful not to filter out things that, as you said, could easily exceed 3072 values. Do you think it could make sense to add a constraint on the total size of `self.records` as opposed to just the individual `record`?

davidslater: I'm less confident on what a good value would be here. And it would be rough to go through an entire long scenario run to only figure out at the end that the full record was too large. I would rather have it successfully write the giant thing to disk.

What do you think about a warning that runs at the end that says that the results dictionary is very large - check to ensure large arrays are not being written?

Alternatively, we could keep running track of the size of the records thus far, and have it drop records once it exceeds a certain large size. (But what size would that be? 100 MB? Would it keep running? Error Out?)

lcadalzo: I'm good with a warning. It does feel a bit arbitrary picking a threshold, so at least the consequences (of a "false positive") we can live with moreso if it's a warning. 100MB seems reasonable

davidslater: I put the warning in `scenario.py`, since that is what is writing results to disk, and set it at 128 MB. Since it is a warning, I did not make it configurable.

lcadalzo (approved):
LGTM

Issue #1492, opened 2022-05-05
Poisoning documentation and cleanup


swsuggs:
Create a new `docs/poisoning.md` and include:
- Threat model descriptions (clean label, dirty label, witches' brew)
- Poisoning-specific config information (adhoc, attack kwargs)
- Poisoning and fairness metrics

Also---add a config kwarg to witches' brew to overwrite the presaved dataset (if it exists) instead of raising an error.

swsuggs:
finished in #1503

PR #1503, opened 2022-05-10
Poisoning doc and cleanup

swsuggs:
For #1492 
In addition to what is described in the linked issue, this PR trims and consolidates a lot of the poisoning code.  GTSRB preprocessing is moved to the dataset file instead of the GTSRB scenario.  The GTSRB scenarios are consolidated into poison.py leaving simply a DLBD scenario and a CLBD scenario.  Config, scenario, and attack names are clarified.  The spectral_signature defense is removed because it exists in ART.


docs/poisoning.md
```diff
@@ -0,0 +1,167 @@
...
+The trigger is usually a small (but not imperceptible) image superposed on the data, and the adversary's goal is to force the model to misclassify test images that have the trigger applied.  Armory includes several trigger images under `utils/triggers/`.
+
+
+In poisoning attacks, the term _source class_ refers to the label of the image(s) that the adversary hopes to misclassify.  In the case of a targeted attack, the _target class_ is the desired misclassification label.  Neither of these terms describes which class gets poisoned; that depends on the threat model.  All of Armory's poisoning scenarios perform targeted attacks.  For simplicity, most Armory scenarios assume a single source class (all images to misclassify are from the same class) and a single target class (all misclassifications are aiming for the same label).  The exception is Witches' Brew, which accepts images from arbitrary source classes that can all have distinct targets.
```
reed-twosixlabs, line 14: > Neither of these terms describes which class gets poisoned; that depends on the threat model.

My understanding is that _source class_ is only meaningful for DLBD attacks since it's not a useful concept for CLBD and WB attacks. Is that still the case for our implementation?

docs/poisoning.md
```diff
@@ -0,0 +1,167 @@
...
+### Clean-label backdoor
+
+In a [Clean-label Backdoor (CLBD) Attack](https://people.csail.mit.edu/madry/lab/cleanlabel.pdf), the triggers are applied to target-class images during training.  The poisoned samples also undergo some imperceptible gradient-based modifications to weaken the natural features, thus strengthening the association of the trigger with the target class label.
+At test time, the adversary applies the trigger to source-class images in order to get them to misclassify as `target`.
+
+
+
+### Witches' brew
+
+[Witches' Brew](https://arxiv.org/abs/2009.02276) is a clean-label attack but there is no backdoor or trigger involved.  The adversary selects individual `source` images from the test set; these are the images that the adversary wants to misclassify as `target` and are called _triggers_, not to be confused with the backdoor trigger described before.  The attack uses a gradient-matching algorithm to modify a portion of the train-set target class, such that the unmodified test-set triggers will be misclassified.
```
reed-twosixlabs, line 32: > The attack uses a gradient-matching algorithm to modify a portion of the train-set target class, such that the unmodified test-set triggers will be misclassified.

Is poisoning only members of the target class part of the attack or just our implementation?

swsuggs: It's Taesung's (ART's) implementation.  Not sure whether the original formulation required it.

docs/poisoning.md
```diff
@@ -0,0 +1,167 @@
...
+## Metrics
+
+The four primary poisoning metrics are:
+- `accuracy_on_benign_test_data_all_classes`
+- `accuracy_on_benign_test_data_source_class`
+- `accuracy_on_poisoned_test_data_all_classes`
+- `attack_success_rate`
+
+These are computed after the model is trained on a poisoned dataset.  First, all of the test data is evaluated with no poison.  This gives us the first two metrics.  Next, data from the source class is poisoned and evaluated again.  The third metric, `accuracy_on_poisoned_test_data_all_classes`, is the total classification accuracy when the source class is poisoned.  The fourth metric, `attack_success_rate`, only measures the percentage of source-class examples that are misclassified as `target`.  In a well-executed attack, all these metrics will have high values: the first indicating that we have an effective, well-trained model;
+the second confirming that the model is accurate on the source class; the third showing that total accuracy does not decrease substantially when one class is poisoned, and the fourth demonstrating that we can induce targeted misclassification in the source class.
```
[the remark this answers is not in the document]
swsuggs: It is that, yes.  But I don't think it should be low...  Ideally, our attack leaves accuracy on all other classes unchanged, right?  So we don't want the victim to notice accuracy dropping when we start poisoning some images and toss the model out.  Reducing the accuracy on only one class won't affect overall accuracy too much, although of course this partly depends on how many classes there are

armory/scenarios/poisoning_dlbd.py
```diff
@@ -69,7 +70,11 @@ def poison_dataset(self, x, y, return_index=False, fraction=None):
         return poison_x, poison_y
 
 
-class Poison(Scenario):
```
davidslater, line 72: Let's keep the name of this `Poison`, so that it's clear that it's the base class that other poisoning scenarios inherit from. You can keep the "dirty label" as part of the docstring, though.

davidslater (approved):
LGTM

PR #1514, opened 2022-05-13
WIP: Method hooking

### LINKED ISSUE #1512: Add method hooking to probe functionality
This would make it much easier to grab values from various functions, such as in https://github.com/twosixlabs/armory/issues/1462, where we could simply hook this method: https://github.com/twosixlabs/armory/blob/388edde7d85f96dac6a96c13854b955f1bb5c3c3/armory/art_experimental/attacks/carla_obj_det_patch.py#L262

Something like our eval patch (but hooking instead of replacing):
```
import types                                                                              
                                                                                          
                                                                                          
def patch_method(obj):                                                                    
    """                                                                                   
    Patch method for given class or object instance.                                      
                                                                                          
    If a class is passed in, patches ALL instances of class.                              
    If an object is passed in, only patches the given instance.                           
    """                                                                                   
                                                                                          
    def decorator(method):                                                                
        if not isinstance(obj, object):                                                   
            raise ValueError(f"patch_method input {obj} is not a class or object")        
        if isinstance(obj, type):                                                         
            cls = obj                                                                     
            setattr(cls, method.__name__, method)                                         
        else:                                                                             
            setattr(obj, method.__name__, types.MethodType(method, obj))                  
        return method                                                                     
                                                                                          
    return decorator
```
Or something similar to:
https://stackoverflow.com/questions/35758323/hook-python-module-function

This could involve pre-hooking (hooking the inputs to the method) and post-hooking (hooking the return value).

davidslater:
Fixes #1512

### DIFF
--- a/armory/instrument/hooking.py
+++ b/armory/instrument/hooking.py
@@ -1,0 +1,52 @@
+import types
+
+
+def hook_wrapper(method, pre_method_hook=None, post_method_hook=None):
+    def wrapped(*args, **kwargs):
+        if pre_method_hook is not None:
+            pre_method_hook(*args, **kwargs)
+        return_value = method(*args, **kwargs)
+        if post_method_hook is not None:
+            post_method_hook(return_value)
+        return return_value
+
+    return wrapped
+
+
+def method_hook(obj, method_name, pre_method_hook=None, post_method_hook=None):
+    """
+    Hook target method and return the original method
+
+    If a class is passed in, hooks ALL instances of class.
+    If an object is passed in, only hooks the given instance.
+    """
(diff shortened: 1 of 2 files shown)


lcadalzo:
See [this](https://github.com/lcadalzo/armory/commit/c36c6be113192bfff1ef64ae36cd107825e174fa) commit for an example that combines hooking with ExportMeters to save off just the patch.

```
I have no name!@dc7a6c888c8c:/workspace$ ls /armory/outputs/2022-05-23T203114.906345/saved_samples/
batch_0_ex_0_hooked_method.patch.png
```

PR #1516, opened 2022-05-15
Eval5 update carla video attack

yusong-tan:
Add code to initialize the patch to the correct size for Eval5 data


lcadalzo:
@yusong-tan what effect(s) (qualitative and/or quantitative) did you observe that this change has on the results?

lcadalzo:
I'm observing 17.05% `adversarial_video_tracking_mean_iou` on the undefended model, slightly below the current baseline of 18%

PR #1519, opened 2022-05-18
add hub disconnect_meter() method

### LINKED ISSUE #1511: Add hub.remove_meter() method
An example of where this could be useful (after #1488 is merged): the witches brew scenario's parent creates export meters for `x` and `x_adv`, but we don't an exporter for the latter since its probe value is never set. In the witches brew scenario code, this could look something like:

```
def load_export_meters(self):
    super().load_export_meters()
    self.hub.remove_meter("x_adv_exporter")
```

If there is no meter in `self.meters` for which `meter.name == "x_adv_exporter"`, ideally this wouldn't break but just be a no-op

lcadalzo:
Closes #1511

### DIFF
--- a/armory/instrument/instrument.py
+++ b/armory/instrument/instrument.py
@@ -353,15 +353,10 @@ class Hub:
         self.meters.append(meter)
         self.mapper.connect_meter(meter)
 
-    def disconnect_meter(self, meter_name):
-        """
-        :param meter_name: string, name of meter
-        """
-        for i, meter in enumerate(self.meters):
-            if meter.name == meter_name:
-                removed_meter = self.meters.pop(i)
-                self.mapper.disconnect_meter(removed_meter)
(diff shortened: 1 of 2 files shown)


lcadalzo:
to test this for the cifar10 scenario, I (locally only) temporarily added 
```
pert_meter = get_hub().meters[-1]
get_hub().disconnect_meter(pert_meter)
```

in the scenario `load()` method after metrics are loaded, then observing in the output JSON that (as expected) there are no perturbation results

```
"results": {
        "adversarial_mean_categorical_accuracy": [
            0.0
        ],
        "benign_mean_categorical_accuracy": [
            0.0
        ],
        "compute": {}
    },
```

lcadalzo:
The initial commit had `disconnect_meter()` take a string, but I found it much simpler to have the method take a meter

PR #1521, opened 2022-05-20
improve import time

davidslater:
This PR attempts to reduce the import time of armory. It does this in two ways:
1) Generates the `__version__` only if asked for.
2) Delays import of torch, tensorflow, and art until a submodule needing them (`art_experimental`, `baseline_models`, `data`, `scenarios`) is loaded. Also adds a logging message to that effect.

Once `tiga` is merged in, the time lag for version generation should be fairly minimal, but this will reduce things for now.

### DIFF
--- a/armory/__init__.py
+++ b/armory/__init__.py
@@ -1,131 +1,20 @@
-"""Adversarial Robustness Evaluation Test Bed
-
-ARMORY Versions use "Semantic Version" scheme where stable releases will have versions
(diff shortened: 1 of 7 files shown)


mwartell:
This is a nice modularization of this mechanism.  SCM is nice in principle, but the native python support for it is still poorly defined, so I think this is a fine stopgap.

PR #1522, opened 2022-05-23
use version string from armory version for base image tag

mwartell:
In #1521 we backed off (slightly) from a firm dependence on the setuptools_scm module.  The `armory.__version__` attribute is now the authoritative armory version, so use that to create the base tag.


davidslater (approved):
Tested and LGTM.

Issue #1472, opened 2022-05-02
Error running librispeech_asr_imperceptible_undefended.json


lcadalzo:
```
File "/opt/conda/lib/python3.8/site-packages/art/attacks/evasion/imperceptible_asr/imperceptible_asr_pytorch.py", line 567, in _attack_2nd_stage[best_loss_2nd_stage[local_batch_size_idx] = loss_2nd_stage[local_batch_size_idx].numpy()
		RuntimeError^[[0m:^[[1m Can't call numpy() on Tensor that requires grad. Use tensor.detach().numpy() instead
```

```
python -m armory run scenario_configs/librispeech_asr_imperceptible_undefended.json
```

lcadalzo:
Note: I'm unable to reproduce this error manually after observing it from the logs of the batch of runs

davidslater:
I think this error only shows up when using GPUs. I was able to reproduce.

It looks like it's an ART error in 
https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/art/attacks/evasion/imperceptible_asr/imperceptible_asr_pytorch.py#L567

```
best_loss_2nd_stage[local_batch_size_idx] = loss_2nd_stage[local_batch_size_idx].numpy()
```
should be
```
best_loss_2nd_stage[local_batch_size_idx] = loss_2nd_stage[local_batch_size_idx].detach().cpu().numpy()
```

Looks like there could be a similar issue here at https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/art/attacks/evasion/imperceptible_asr/imperceptible_asr_pytorch.py#L412 , where 
```
                        max_local_delta = np.max(np.abs(local_delta[local_batch_size_idx].detach().numpy()))
```
should be
```
                        max_local_delta = np.max(np.abs(local_delta[local_batch_size_idx].detach().cpu().numpy()))
```

davidslater:
PR submitted to ART: https://github.com/Trusted-AI/adversarial-robustness-toolbox/pull/1677

Issue #1468, opened 2022-05-02
Error running resisc45_baseline_densenet121_patch.json


lcadalzo:
```
x_adv = self.attack.generate(x=x, y=y_target, **self.generate_kwargs)
        │    │      │          │    │           │    └ {}
        │    │      │          │    │           └ <armory.scenarios.image_classification.ImageClassificationTask object at 0x7ff9794165e0>
        │    │      │          │    └ None
        │    │      │          └ array([[[[0.31764707, 0.3254902 , 0.27450982],
        │    │      │                     [0.32156864, 0.32941177, 0.2784314 ],
        │    │      │                     [0.32156864, 0.3294117...
        │    │      └ <function AttackWrapper.generate at 0x7ff83e38e550>
        │    └ <armory.art_experimental.attacks.patch.AttackWrapper object at 0x7ff83449fd30>
        └ <armory.scenarios.image_classification.ImageClassificationTask object at 0x7ff9794165e0>
File "/workspace/armory/art_experimental/attacks/patch.py", line 16, in generate
self._attack.generate(x, y=y, **kwargs)
│    │       │        │    │    └ {}
│    │       │        │    └ None
│    │       │        └ array([[[[0.31764707, 0.3254902 , 0.27450982],
│    │       │                   [0.32156864, 0.32941177, 0.2784314 ],
│    │       │                   [0.32156864, 0.3294117...
│    │       └ <function AdversarialPatch.generate at 0x7ff844c91310>
│    └ <art.attacks.evasion.adversarial_patch.adversarial_patch.AdversarialPatch object at 0x7ff8371785b0>
└ <armory.art_experimental.attacks.patch.AttackWrapper object at 0x7ff83449fd30>
File "/opt/conda/lib/python3.8/site-packages/art/attacks/evasion/adversarial_patch/adversarial_patch.py", line 162, in generate
raise ValueError("Adversarial Patch attack requires target values `y`.")
ValueError: Adversarial Patch attack requires target values `y`.

```

To reproduce: 
```
python -m armory run scenario_configs/eval1-4/resisc45/resisc45_baseline_densenet121_patch.json --check
```

lcadalzo:
[...]

Issue #1467, opened 2022-05-02
Error running resisc45_baseline_densenet121_cascade.json


lcadalzo:
```
attack = config_loading.load_attack(attack_config, self.model)
         │              │           │              │    └ art.estimators.classification.keras.KerasClassifier(model=<keras.engine.functional.Functional object at 0x7f4b36fc4910>, use_...
         │              │           │              └ <armory.scenarios.image_classification.ImageClassificationTask object at 0x7f4c793af5e0>
         │              │           └ {'knowledge': 'white', 'kwargs': {'eps': 0.2, 'inner_configs': [{'kwargs': {'eps': 0.2, 'eps_step': 0.1, 'max_iter': 100, 'nu...
         │              └ <function load_attack at 0x7f4b3d58e1f0>
         └ <module 'armory.utils.config_loading'
from '/workspace/armory/utils/config_loading.py'>
File "/workspace/armory/utils/config_loading.py", line 159, in load_attack
attack = attack_fn(classifier, **attack_config["kwargs"])
         │         │             └ {'knowledge': 'white', 'kwargs': {'eps': 0.2, 'inner_configs': [{'kwargs': {'eps': 0.2, 'eps_step': 0.1, 'max_iter': 100, 'nu...
         │         └ art.estimators.classification.keras.KerasClassifier(model=<keras.engine.functional.Functional object at 0x7f4b36fc4910>, use_...
         └ <class 'armory.art_experimental.attacks.cascading_attack.CascadingAttack'>
File "/workspace/armory/art_experimental/attacks/cascading_attack.py", line 12, in __init__
self.attacks.append(load_attack(inner_config, estimator))
│    │       │      │           │             └ art.estimators.classification.keras.KerasClassifier(model=<keras.engine.functional.Functional object at 0x7f4b36fc4910>, use_...
│    │       │      │           └ {'kwargs': {'batch_size': 1, 'confidence': 0.9, 'eps': 0.2, 'max_iter': 10, 'targeted': False}, 'module': 'art.attacks.evasio...
│    │       │      └ <function load_attack at 0x7f4b3d58e1f0>
│    │       └ <method 'append' of 'list' objects>
│    └ [<art.attacks.evasion.projected_gradient_descent.projected_gradient_descent.ProjectedGradientDescent object at 0x7f4b34420e20>]
└ <armory.art_experimental.attacks.cascading_attack.CascadingAttack object at 0x7f4b36fc4eb0>
[...]

PR #1525, opened 2022-05-24
update to art 1.10.2

### LINKED ISSUE #1465: Error running librispeech_asr_imperceptible_defended.json
```
File "/opt/conda/lib/python3.8/site-packages/art/defences/preprocessor/mp3_compression.py", line 119, in __call__
raise ValueError("Mp3 compression can only be applied to temporal data across at least one channel.")
ValueError: Mp3 compression can only be applied to temporal data across at least one channel.
```

To reproduce: 

```
python -m armory run scenario_configs/librispeech_asr_imperceptible_defended.json --check
```

### LINKED ISSUE #1523: Upgrade to ART 1.10.2

davidslater:
Fixes #1523

This may fix
#1465 
#1467 - I'll upload new scenario config:
    `scenario_configs/eval1-4/resisc45/resisc45_baseline_densenet121_cascade.json`
#1472 

But that needs to be tested.

The fix for #1468 is slated for a later version of ART.

### DIFF
--- a/armory-base-requirements.txt
+++ b/armory-base-requirements.txt
@@ -34,7 +34,7 @@ scipy==1.7.*
 matplotlib==3.*
 
 # ART
-adversarial-robustness-toolbox==1.10.0
+adversarial-robustness-toolbox==1.10.2
 
 # Data dependencies
 Pillow==9.0.*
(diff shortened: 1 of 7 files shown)


lcadalzo:
1465: this reached the same error as before, but when I change the defense config to `"Mp3CompressionPyTorch"` it works.
1467: confirmed that this now works
1472: even with ART 1.10.1, I'm still unable to reproduce the issue (with or without GPUs). That being said, I did also try running the config with 1.10.2 and this worked too

I'm still building docker images (I tested the configs above with `--interactive` and pip installing inside the container), but I don't envision any issues with that. Seems like we should update the config for 1465 but otherwise lgtm

davidslater:
@lcadalzo config updated

lcadalzo:
Cool, will merge after I see the images all built ok

PR #1535, opened 2022-05-26
Release fix

### LINKED ISSUE #1531: Update release.yml to ensure pip installed or build.py to remove that dependency.

davidslater:
Fixes #1531

This updates the build script to not require pip installation. We already require it to be run from the root of the armory repo, so it seems weird to require a pip install to simply import armory.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,7 +41,6 @@ jobs:
         run: |
           rm -rf /usr/share/dotnet &
           pip install -r requirements.txt
-          pip install -e .
           version=$(python -m armory --version)
           docker login --username ${{ secrets.DOCKER_USERNAME }} --password ${{ secrets.DOCKER_PASSWORD }}
           docker pull twosixarmory/tf2:latest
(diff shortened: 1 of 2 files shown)


lcadalzo:
lgtm, just one minor comment unrelated to this PR: we currently allow doing either a single build or all the builds. If we were to add `nargs='+'` to the `parser.add_argument()` call, this would allow for an arbitary number e.g.
```
python docker/build.py pytorch tf2
```
to build the tf2 and pytorch images. If we make this change, we'd need to modify the code below
```
else:
    frameworks = [args.framework]
```
since argparse would then already parse the input(s) as a list
`

lcadalzo:
lgtm, just one minor comment unrelated to this PR: we currently allow doing either a single build or all the builds. If we were to add `nargs='+'` to the `parser.add_argument()` call, this would allow for an arbitary number e.g.
```
python docker/build.py pytorch tf2
```
to build the tf2 and pytorch images. If we make this change, we'd need to modify the code below
```
else:
    frameworks = [args.framework]
```
since argparse would then already parse the input(s) as a list
`

davidslater:
@lcadalzo regarding the `nargs` comment, I think maybe we can revisit that if we expand the number of docker images we create. However, for now, I would like to reserve the possibility of adding more non-keyword args without having to remove the list functionality.

PR #1536, opened 2022-05-26
WIP: remove weird split methods

davidslater:
Takes the useful part out of #1345

### DIFF
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -8,19 +8,16 @@ data is found at '<dataset_dir>/cifar10'
 The 'downloads' subdirectory under <dataset_dir> is reserved for caching.
 """
 
-import json
 import os
-import re
 from typing import Callable, Union, Tuple, List
 
 import numpy as np
-from armory.logs import log
 from PIL import ImageOps, Image
-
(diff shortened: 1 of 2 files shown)


lcadalzo:
I might be missing some context but this PR appears to remove split string parsing functionality. Was there something added in a previous PR that obviated the need for this functionality?

davidslater:
Is all filtering functionality done in TFDS? When you call `--index=1,3,45`, or `--classes=5`, is that actually done in TFDS, not in the Data Generator?

lcadalzo:
To rephrase my earlier comment: this PR removes some code, but it doesn't add any. Is the code removed by this PR not needed at all? I don't see an issue this PR is tied to, and I'm unfamilar with 1345, so just trying to gain context

davidslater:
Let me take another look at it. I thought that this functionality was duplicative with the `--index` and `--class` material, but I could be wrong - it could be the underlying mechanism.

This is related to some of the dataset cleanup, but seemed independent, so I pulled it out separately. Let me double check that the other functionality still works, and if not, I'll hold off until the full dataset revamp.

Issue #492, opened 2020-04-28
Per-Class Metrics


davidslater:
Per-class statistics (precision, recall, confusion matrices, etc.).

Rationale: Helps diagnose where models/defenses/etc are improving/failing when compared to the baseline.

davidslater:
Intel/Georgia Tech request.

davidslater:
We currently provide object detection per class and per-class accuracy. Confusion matrices would be nice, though I will take this off the current milestone for now.

PR #1538, opened 2022-05-27
WIP per-class metrics

swsuggs:
For #492

### DIFF
--- a/armory/instrument/config.py
+++ b/armory/instrument/config.py
@@ -216,6 +216,8 @@ class ResultsLogWriter(LogWriter):
         elif any(m in name for m in QUANTITY_METRICS):
             # Don't include % symbol
             f_result = f"{np.mean(result):.2}"
+        elif isinstance(result, dict):
+            f_result = f"{result}"
         else:
             f_result = f"{np.mean(result):.2%}"
         log.success(
@@ -253,6 +255,11 @@ def _task_metric(
     elif name == "word_error_rate":
         final = metrics.get("total_wer")
         final_suffix = "total_word_error_rate"
+    elif name == "per_class_mean_accuracy":
+        metric = metrics.get("identity_unzip")
+        func = metrics.get("per_class_mean_accuracy")
+        final = lambda x: func(*metrics.task.identity_zip(x))
+        final_suffix = name
     elif use_mean:
         final = np.mean
         final_suffix = f"mean_{name}"
--- a/armory/utils/config_schema.json
+++ b/armory/utils/config_schema.json
@@ -261,7 +261,8 @@
                 "carla_od_disappearance_rate",
                 "carla_od_hallucinations_per_image",
                 "carla_od_misclassification_rate",
-                "carla_od_true_positive_rate"
+                "carla_od_true_positive_rate",
+                "per_class_mean_accuracy"
             ]
         },
         "sysconfig": {


swsuggs:
@davidslater The poisoning scenario adds this metric automatically, right in the code.  Is this an acceptable way to make it available to other scenarios through the config/metric/task section?

armory/instrument/config.py
```diff
@@ -253,6 +255,11 @@ def _task_metric(
     elif name == "word_error_rate":
         final = metrics.get("total_wer")
         final_suffix = "total_word_error_rate"
+    elif name == "per_class_mean_accuracy":
+        metric = metrics.get("identity_unzip")
+        func = metrics.get("per_class_mean_accuracy")
+        final = lambda x: func(*metrics.task.identity_zip(x))
```
swsuggs, line 261: flake8 didn't like that I used a lambda here.  It told me to "use a def".  Is that really better for this situation?

davidslater: A lambda one-liner is a bit odd with assignment. It's essentially the same as:
```
def final(x):
    return func(*metrics.task.identity_zip(x))
```
I don't see any issues with using that.

davidslater:
I like adding it to the config part of instrumentation.

swsuggs:
@davidslater So far I've added (beside per-class accuracy) a confusion matrix, and precision and recall for each class.  Two questions: Will you check that I'm computing the right thing (I described my understanding in the comments), and that the output is in a sufficiently useful format (dict, array, etc)?  Can you think of any other metrics or statistics that would be nice to have?

davidslater (changes requested):
Let's move the metrics to `task`, as they take in the paired `y` and `y_pred` as inputs.

armory/metrics/statistical.py
```diff
@@ -21,6 +22,54 @@ def register(metric):
...
+        D[class_] = (precision, recall)
+
+    return D
+
+
+@populationwise
+def confusion_matrix(y, y_pred, normalize=True):
+    """
+    Produce a matrix C such that C[i,j] is the percentage of class i that was classified as j
+    If normalize is False, C[i,j] is the actual number of such elements, rather than the percentage
```
davidslater, line 55: Make it clear what you're normalizing with respect to.

armory/metrics/statistical.py
```diff
@@ -21,6 +22,54 @@ def register(metric):
...
+    """
+    # Assumes that every class is represented in y
+    y = np.array(y)
+    y_pred = np.array(y_pred)
+    if y_pred.ndim == 2:
+        y_pred = np.argmax(y_pred, axis=1)
+    N = len(np.unique(y))
+    C = np.zeros((N, N))
```
davidslater, line 63: Numpy array is a good form for this.

armory/metrics/statistical.py
```diff
@@ -21,6 +22,54 @@ def register(metric):
...
+    """
+    Produce a matrix C such that C[i,j] is the percentage of class i that was classified as j
+    If normalize is False, C[i,j] is the actual number of such elements, rather than the percentage
+    """
+    # Assumes that every class is represented in y
+    y = np.array(y)
+    y_pred = np.array(y_pred)
+    if y_pred.ndim == 2:
+        y_pred = np.argmax(y_pred, axis=1)
+    N = len(np.unique(y))
```
davidslater, line 62: If `y_pred` is 2D, you can use that to derive N. (Or at least check to ensure that they match).

armory/metrics/statistical.py
```diff
@@ -21,6 +22,54 @@ def register(metric):
     return set_namespace(registered, metric)
 
 
+@populationwise
+def precision_and_recall(y, y_pred):
```
davidslater, line 26: I would rename as `per_class_precision_and_recall`.

armory/instrument/config.py
```diff
@@ -253,6 +257,31 @@ def _task_metric(
...
+        final_suffix = name
+    elif name == "precision_and_recall":
+        metric = metrics.get("identity_unzip")
+        func = metrics.get("precision_and_recall")
+
+        def final(x):
+            return func(*metrics.task.identity_zip(x))
+
+        final_suffix = name
+
```
davidslater, line 284: Can you combine these into a single block? E.g., 
```python
    elif name in ("per_class_mean_accuracy", "confusion_matrix", "precision_and_recall"):
        metric = metrics.get("identity_unzip")
        func = metrics.get(name)

        def final(x):
            return func(*metrics.task.identity_zip(x))

        final_suffix = name
```

armory/metrics/statistical.py
swsuggs: I may be misunderstanding, but N is the number of classes, not the total number of items.  Hence length of np.unique(y) and not length of y.  I don't think we can assume every class will show up in y_pred.  For that matter it seems a little risky to assume they will all be present in y.

davidslater: I think this implicitly assumes that the classes are all integers from 0 to N - 1. However, if y has missing classes, then there will be some misalignment.

swsuggs: Oh right, of course.  Is it true that Armory scenarios will always have a 2D y_pred?  Or it just depends on how the meter and probes are set up?  So far the only source of a 1D y_pred I've encountered is my own unit tests, but I can expand those to 2D and then get N the way you described.

davidslater: Right now it's dependent on the underlying model, unfortunately.

swsuggs: Well after all, if a class is totally absent from y, its row in the matrix would be all zeros since it was never classified as anything.  So maybe what I need to do is make this a dictionary after all, and key it with class labels, so if one is missing, then at least it will be clear what rows are what class.  Alternatively, I could add a row of zeros at the index of missing class labels, but this would only be possible for missing labels less than the greatest non-missing label.

armory/instrument/config.py
swsuggs: Yep that was on my todo list

armory/metrics/statistical.py
```diff
@@ -21,6 +22,54 @@ def register(metric):
...
+    # breakpoint()
+    N = C.shape[0]
+    D = {}
+    for class_ in range(N):
+        # precision: true positives / number of items identified as class_
+        tp = C[class_, class_]
+        total_selected = C[:, class_].sum()
+        precision = tp / total_selected
+
+        # recall: true positives / number of actual items in class_
```
swsuggs, line 42: per-class recall is the exact same as per-class accuracy, which I didn't realize till now.  Is it still useful to have two separate `per_class_accuracy` and `per_class_precision_and_recall` functions?

davidslater: Sure, let's keep both.

armory/metrics/statistical.py
davidslater: Let's just have the function assume that `y_pred` is 2D (and add that to the docstring). Other things can be handled by the user.

davidslater:
See my two recent comments. Beyond that, I think what needs to be done is:
A) remove the changes to `armory/instrument/config.py`, and `armory/utils/config_schema.json`.
B) move the new functions from `statistical` to `task`, since we're registering them as task populationwise metrics anyhow.
C) Probably easiest to rebase off of develop.

Issue #1540, opened 2022-05-30
Add attack success rate metric to witches' brew


swsuggs:
Witches' brew reports how many trigger images are misclassified, but not how many are classified as their target, as the other poisoning scenarios do.

PR #1542, opened 2022-05-30
add attack success rate for witches' brew

swsuggs:
Addresses #1540


swsuggs:
Will re-open on the dev_0.15.2 branch

PR #1544, opened 2022-06-01
Remove unneeded scripts

davidslater:
I don't think we need these scripts. The `release.yml` pushes docker images.


davidslater:
@mwartell can you take a look at this when you get a chance? (low priority)

mwartell (approved):

PR #1550, opened 2022-06-03
updating to ART 1.10.3

### LINKED ISSUE #1549: Update to ART 1.10.3
1.10.3 will include a bug fix for Adversarial Texture that speeds up attack convergence

lcadalzo:
Closes #1549. Leaving as `WIP` until 1.10.3 is released, at which point I'll manually trigger CI

### DIFF
--- a/armory-base-requirements.txt
+++ b/armory-base-requirements.txt
@@ -34,7 +34,7 @@ scipy==1.7.*
 matplotlib==3.*
 
 # ART
-adversarial-robustness-toolbox==1.10.2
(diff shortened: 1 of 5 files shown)


lcadalzo:
tests have all passed

mwartell (changes requested):
There is an ART version in host-requirements.txt and armory-base-requirements.txt which should match.

Alternatively, they could be scrapped altogether, but if they are present, they should agree in case someone tries to use them.

mwartell (approved):

PR #1558, opened 2022-06-08
Witches brew proxy model fix

swsuggs:
Taesung confirmed that the classifier passed to the attack should be already trained.


lcadalzo:
@swsuggs you'd need to trigger CI manually, but I don't think this will pass formatting test. When I run black on the file, it reformats

swsuggs:
@lcadalzo Thanks, looks like I ran black and flake in the wrong directory.  By the way, how do I run CI manually?

swsuggs:
Cool, [here](https://github.com/swsuggs/armory/actions/runs/2463651074/) go the tests

PR #1575, opened 2022-06-25
add Eval 5 carla object detection test data

yusong-tan:


lcadalzo:
I think it makes the most sense to have these changes applied on top of `r0.15.3`. I've created a new branch `dev_0.15.4` which was branched off of `r0.15.3`.

When I redirected this PR to `dev_0.15.4`, the diff was messy because `yusong-tan:eval5_carla_od_test` isn't branched off of `r0.15.3` but rather `develop`. There have been changes to `develop` recently that I don't think are in an official state and did not go into `0.15.3`. 

The cleanest thing to do here, I think, is to create a new branch off of `r0.15.3` (e.g. `carla-test-sets`), apply the changes in this PR to that branch, and then make a PR from `carla-test-sets` into `dev_0.15.4`. Then `dev_0.15.4` can be merged into `master` and a 0.15.4 release cut from master

lcadalzo:
> Should I resubmit branching off r0.15.3?

It looks like @swsuggs has done so already. I'll be reviewing #1579

PR #1579, opened 2022-06-29
Eval5 carla test sets

swsuggs:
includes video tracking test 2.0.0 and object detection test 2.0.0, rebased from #1575 and #1573

### DIFF
--- a/armory/data/adversarial/carla_obj_det_test.py
+++ b/armory/data/adversarial/carla_obj_det_test.py
@@ -4,13 +4,10 @@ import collections
 import json
 import os
 from copy import deepcopy
-from PIL import Image
 import numpy as np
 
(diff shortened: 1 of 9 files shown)


swsuggs:
@lcadalzo 
1. Done, please let me know if that looks right
2. CI run passed: https://github.com/swsuggs/armory/actions/runs/2585169534.  Did you mean for me to update the tests for the new shape too?

lcadalzo:
I think end-to-end tests aren't being run currently, although we intend to get those back running soon. That explains why we don't see an hour.

[These](https://github.com/swsuggs/armory/blob/eval5-carla-test-sets/tests/end_to_end/test_e2e_datasets.py#L612-L614) lines should be updated for the new shape

PR #1585, opened 2022-07-06
Updated Scenario.md with Eval 5 CARLA Test Results

yusong-tan:
…est results


swsuggs:
I have replicated the results.  Thanks Mike!

PR #1587, opened 2022-07-08
Adding a ResNet-18 model optimized with SGD.

TS-Lee:
We see some attacks perform differently on a model trained on different optimizers. To test that, I want to add a ResNet-18 model optimized with SGD.

### DIFF
--- a/armory/baseline_models/pytorch/resnet18.py
+++ b/armory/baseline_models/pytorch/resnet18.py
@@ -60,3 +60,25 @@ def get_art_model(
         clip_values=(0.0, 1.0),
     )
     return wrapped_model
+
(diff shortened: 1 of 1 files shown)


TS-Lee:
Yes. We want to see if an attack or a defense is sensitive to a certain parameter, and hence it becomes inapplicable to a certain model.

swsuggs:
Would you please run `black armory/baseline_models/pytorch/resnet18.py`, making sure you have black version 22.3.0?

TS-Lee:
Done. It looks like it passed all tests.

swsuggs:
Great, thanks!

PR #1590, opened 2022-07-11
Update release.yml

mwartell:
temporary stop of push to pypi


mwartell:
defective attempt, closing

PR #1602, opened 2022-08-03
bring v0.15.4 master back into develop

mwartell:
I made a trial merge for this PR and git found only one conflict in `poisoning_witches_brew.py`. I think the conflict resolution is obvious, but I wanted @swsuggs to review since the changes came in on his commit c2242a72. 

~The `armory --version command` in my trial merge reads `0.15.2.g5b4196e4` so I'm holding this as draft until I find why.~
This was cleared by commiting (but not pushing) my test merge. This PR is still draft waiting for a merge resolution on the one conflicted file. This is in fact how setuptools_scm should work, the most resent committed tag on `develop` was v0.15.1; after I committed the merge, the tag advanced to 0.15.4+


swsuggs (approved):
```
<<<<<<< HEAD
        self.hub.connect_meter(
            Meter(
                "attack_success_rate",  # percent of triggers classified as target
                metrics.get_supported_metric("categorical_accuracy"),
                "scenario.target[trigger]",
                "scenario.y_pred[trigger]",
                final=np.mean,
                final_name="attack_success_rate",
                record_final_only=True,
            )
        )

        per_class_mean_accuracy = metrics.get_supported_metric(
            "per_class_mean_accuracy"
        )
=======
        per_class_mean_accuracy = metrics.get("per_class_mean_accuracy")
>>>>>>> develop

```
If the obvious resolution is to keep HEAD, then yes that is what we should do

PR #1605, opened 2022-09-06
Eval6 - CARLA Object Detection Dev Data

yusong-tan:
Adding databuilder and checksum for Eval6 CARLA object detection dev data


armory/data/url_checksums/carla_over_obj_det_dev.txt
```diff
@@ -0,0 +1 @@
+armory-public-data carla/carla_over_od_dev_1.0.0.tar.gz 61751299 ee6fbd0f00cf88aaef3ae097e9d661d52b0d97e5688098d342dad4780f931c8a
```
jprokos26, line 1: Should this be changed to the following? I was unsure why the old version did not have the full url and why the cached versions never do either.
```suggestion
https://armory-public-data.s3.us-east-2.amazonaws.com/carla/carla_over_od_dev_1.0.0.tar.gz 61751299 ee6fbd0f00cf88aaef3ae097e9d661d52b0d97e5688098d342dad4780f931c8a
```

yusong-tan: I don't have the answer to this question because it's always done by someone at TwoSix.

jprokos26: Should have put this tag here before, @davidslater thoughts?

armory/data/adversarial/carla_over_obj_det_dev.py
```diff
@@ -0,0 +1,224 @@
...
+                )
+            ),
+            # both modalities share the same objects
+            "objects": tfds.features.Sequence(
+                {
+                    "id": tf.int64,
+                    "image_id": tf.int64,
+                    "area": tf.int64,  # un-normalized area
+                    "boxes": tfds.features.BBoxFeature(),  # normalized bounding box [ymin, xmin, ymax, xmax]
+                    "labels": tfds.features.ClassLabel(num_classes=5),
```
jprokos26, line 76: This dataset no longer has `trafficlight` correct?
As such `num_classes` should be 4 to represent pedestrian, vehicle, green screen, background correct?

yusong-tan: @jprokos26 comment for line 63 is no longer correct. It should say {'pedestrian':1, 'vehicle':2}. `num_classes` is now 3, representing pedestrian, vehicle, background.

The baseline model and configs reflect `num_classes = 3`

armory/data/url_checksums/carla_over_obj_det_dev.txt
davidslater: There are two types of checksums:
`url_checksums` are meant for the original raw data, and have the full `https://...` URL. These are what TFDS uses to build the datasets.
`cached_s3_checksums` have the fully processed TFDS data, and have the `armory-public-data...` format.

Since this one references the raw dataset, it should have the `https://...` format.

armory/data/adversarial/carla_over_obj_det_dev.py
jprokos26:
```suggestion
                        "id": tf.int64,  # {'pedstrian':1, 'vehicles':2}
                        "name": tfds.features.Text(),
                        "supercategory": tfds.features.Text(),
                    }
                )
            ),
            # both modalities share the same objects
            "objects": tfds.features.Sequence(
                {
                    "id": tf.int64,
                    "image_id": tf.int64,
                    "area": tf.int64,  # un-normalized area
                    "boxes": tfds.features.BBoxFeature(),  # normalized bounding box [ymin, xmin, ymax, xmax]
                    "labels": tfds.features.ClassLabel(num_classes=3),
```

PR #1611, opened 2022-09-10
speech commands dataset

swsuggs:
~~Mini~~ speech commands dataset to use for audio poisoning scenario in eval6


davidslater (changes requested):
If `pad_data is False` and `batch_size > 1`, then the preprocessing fails:
```>>> ds = datasets.speech_commands("test", batch_size=8)
>>> x, y = next(ds)
  File "/workspace/armory/data/datasets.py", line 166, in __next__
    return self.get_batch()
  File "/workspace/armory/data/datasets.py", line 159, in get_batch
    x = self.preprocessing_fn(x)
  File "/workspace/armory/data/datasets.py", line 559, in wrapped
    x = function(x)
  File "/workspace/armory/data/datasets.py", line 825, in speech_commands_canonical_preprocessing
    return canonical_audio_preprocess(speech_commands_context, batch)
  File "/workspace/armory/data/datasets.py", line 787, in canonical_audio_preprocess
    assert x.dtype == context.output_type
AssertionError
```

I'm not sure if this is an issue with the standard preprocessing or with your method.

Otherwise, this looks good.

swsuggs:
@davidslater looks like it's the preprocessing.  It worked when I tested it.  But it fails if all items in the batch happen to be the same size.  The problem is [here](https://github.com/twosixlabs/armory/blob/67afab29c3c4bd42dc1584d4f4483692a359f556/armory/data/datasets.py#L781-L783), where for some reason, casting batch as type 'object' will cause all elements to be also type 'object' if they were the same length. 

Same length---will result in error:
```
>>> batch =[np.array([1,2,3]),np.array([4,5,6])]
>>> newbatch = np.array([x.astype(np.float32) for x in batch], dtype=object)
>>> newbatch[0].dtype
dtype('O') 
```
Different lengths---does what we want:
```
>>> batch =[np.array([1,2,3]),np.array([4,5,6,7])]
>>> newbatch = np.array([x.astype(np.float32) for x in batch], dtype=object)
>>> newbatch[0].dtype
dtype('float32') 

```

swsuggs:
I can just check if the elements are the same length or not.  Or is there a better dtype or array structure that could work for both cases?

davidslater:
I would recommend adding an if statement in the canonical audio preprocess to check for same size batch. If that is true, then you would need to do this to have it match the input:
```
new_batch = np.empty(batch.shape, dtype=object)
new_batch[:] = [x.astype(context.output_type) / context.quantization for x in batch]
batch = new_batch
```

The problem is that `dtype=object` can be ambiguous regarding structure.

armory/data/datasets.py
```diff
@@ -778,10 +778,18 @@ def canonical_audio_preprocess(context, batch):
             assert x.min() >= context.input_min
             assert x.max() <= context.input_max
 
-        batch = np.array(
-            [x.astype(context.output_type) / context.quantization for x in batch],
-            dtype=object,
-        )
+        items_are_different_lengths = np.any([len(i)-len(batch[0]) for i in batch])
+
+        if items_are_different_lengths or len(batch) == 1:
```
davidslater, line 783: When `len(batch) == 1`, this will fail the later assertion of `dtype`, as the input type is `np.object`. I would remove the `or len(batch) == 1` portion.

If a batch of size 1 is sent in that is not an object array, it won't hit this code.

davidslater (approved):
LGTM

PR #1615, opened 2022-09-15
Audio poisoning

swsuggs:
Baseline model, attack, config, triggers

PR #1621, opened 2022-09-29
Swsuggs audio poisoning dms

davidslater:


davidslater:
See the commits I added. I can add them to your PR directly, or we can do something different, but hopefully this will make it easier to see and review.

With this, I do still hit an error here:
```

swsuggs:
Thanks!!  This looks great and I'll test it out.  I'm fine with you adding them to my PR.  

I did start getting that error yesterday too.  I'm not yet sure what from since it didn't happen every time, but I was also changing epochs and batch size and poison percentage.

davidslater:
Added to #1615

PR #1628, opened 2022-10-03
Eval6 Object Detection Integration

jprokos26:
# Eval6 dataset & depth attack integration
Resolves https://github.com/twosixlabs/armory/pull/1604
Resolves https://github.com/twosixlabs/armory/pull/1605
Resolves https://github.com/twosixlabs/armory/pull/1606
Resolves https://github.com/twosixlabs/armory/pull/1607
Does not resolve MOT PRs.

### Target values:
Benign mAP values for provided weights files uploaded to s3 matching those received from Mitre. To reproduce run:
```
armory run scenario_configs/eval6/carla_overhead_object_detection/<config_file> --skip-attack
```
| weight | [dev](https://github.com/twosixlabs/armory/pull/1628/files#diff-3a61c78ff13286639831651ce03ff799c4ed1cca9367a760b5adc89ea94d7caeR31) | [val](https://github.com/twosixlabs/armory/pull/1628/files#diff-fa6db5b9676c08e2409f5ae6329189eeef3f88471988bd8edb1c9ebf8919eee8R955) | configs |
|--------|-----|----|-------|
| carla_rgb_weights_eval6.pt | 0.64 | 0.7 | [dev](https://github.com/jprokos26/armory/blob/eval6-integration-scenarios/scenario_configs/eval6/carla_overhead_object_detection/carla_obj_det_adversarialpatch_undefended.json), [val](https://github.com/jprokos26/armory/blob/eval6-integration-scenarios/scenario_configs/eval6/carla_overhead_object_detection/carla_obj_det_val_adversarialpatch_undefended.json)
| carla_depth_weights_eval6.pt | 0.605 | 0.715 | [dev](https://github.com/jprokos26/armory/blob/eval6-integration-scenarios/scenario_configs/eval6/carla_overhead_object_detection/carla_obj_det_depth_adversarialpatch_undefended.json), [val](https://github.com/jprokos26/armory/blob/eval6-integration-scenarios/scenario_configs/eval6/carla_overhead_object_detection/carla_obj_det_depth_val_adversarialpatch_undefended.json)
| carla_multimodal_naive_weights_eval6.pt | 0.63 | 0.675 | [dev](https://github.com/jprokos26/armory/blob/eval6-integration-scenarios/scenario_configs/eval6/carla_overhead_object_detection/carla_obj_det_multimodal_adversarialpatch_undefended.json), [val](https://github.com/jprokos26/armory/blob/eval6-integration-scenarios/scenario_configs/eval6/carla_overhead_object_detection/carla_obj_det_val_multimodal_adversarialpatch_undefended.json)
| carla_multimodal_robust_weights_eval6.pt | 0.665 | 0.72 | [dev](https://github.com/jprokos26/armory/blob/eval6-integration-scenarios/scenario_configs/eval6/carla_overhead_object_detection/carla_obj_det_multimodal_adversarialpatch_defended.json), [val](https://github.com/jprokos26/armory/blob/eval6-integration-scenarios/scenario_configs/eval6/carla_overhead_object_detection/carla_obj_det_val_multimodal_adversarialpatch_defended.json)

All weights files uploaded to `armory-public-data/model-weights/`

@davidslater


swsuggs:
Can you clarify what the target values in your top comment correspond to, as far as undefended/defended, dataset split, attack?  And are those just the numbers you get, or that Mike got?  (or both?)

jprokos26:
> Can you clarify what the target values in your top comment correspond to, as far as undefended/defended, dataset split, attack? And are those just the numbers you get, or that Mike got? (or both?)

Updated description to address these issues.

armory/baseline_models/pytorch/carla_single_modality_object_detection_frcnn.py
```diff
@@ -30,6 +26,16 @@ def get_art_model(
...
+        assert (
+            "roi_heads.box_predictor.cls_score.bias" in checkpoint
+        ), "invalid checkpoint for current model, layers do no match."
+        assert (
+            model.roi_heads.box_predictor.cls_score.out_features
+            == checkpoint["roi_heads.box_predictor.cls_score.bias"].shape[0]
+        ), (
+            f"provided model checkpoint does not match supplied model_kwargs['num_classes']: "
+            f"{model_kwargs['num_classes']} != {checkpoint['roi_heads.box_predictor.cls_score.bias'].shape[0]}"
+        )
```
jprokos26, line 38: Is this too strict? Will the underlying model ever change its structure? If so, there does not seem to be a good way to grab the output shape for a checkpoint without knowing the names of its layers.

davidslater: If it changes structure, we can update the file.

swsuggs:
I'm not finding the model weights in s3, did you upload them to `armory-public-data/model-weights/`? @jprokos26

jprokos26:
> I'm not finding the model weights in s3, did you upload them to `armory-public-data/model-weights/`? @jprokos26

No, they are there now.

swsuggs:
Thanks.  Have you heard from Mike lately about the carla_multimodal_robust_weights_eval6.pt, or should we remind him?

jprokos26:
> Thanks. Have you heard from Mike lately about the carla_multimodal_robust_weights_eval6.pt, or should we remind him?

No I haven't, I will ping him about it now.

davidslater:
I think he said yesterday at the meeting that he was still working on the multimodal one.

armory/data/adversarial/carla_over_obj_det_dev.py
```diff
@@ -0,0 +1,224 @@
...
+                )
+            ),
+            # both modalities share the same objects
+            "objects": tfds.features.Sequence(
+                {
+                    "id": tf.int64,
+                    "image_id": tf.int64,
+                    "area": tf.int64,  # un-normalized area
+                    "boxes": tfds.features.BBoxFeature(),  # normalized bounding box [ymin, xmin, ymax, xmax]
+                    "labels": tfds.features.ClassLabel(num_classes=5),
```
swsuggs, line 76: Likewise, I think num_classes should be 4... pedestrian, vehicle, green screen, background.  But maybe we should double check with Mike

swsuggs:
I haven't run the dpatch configs but the others run and I get the same numbers so that's good.

swsuggs:
But something seems off either with the depth modality or just the way it is exported.
batch_1_ex_0_scenario x_depth

armory/data/adversarial/carla_over_obj_det_dev.py
```diff
@@ -0,0 +1,224 @@
...
+                        "id": tf.int64,
+                    },
+                ),
+                length=2,
+            ),
+            # both modalities share the same categories
+            "categories": tfds.features.Sequence(
+                tfds.features.FeaturesDict(
+                    {
+                        "id": tf.int64,  # {'pedstrian':1, 'vehicles':2, 'trafficlight':3}
```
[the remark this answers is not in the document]
jprokos26: Yes, according to the `kwcoco_annotations.json` there is only pedestrian and vehicle; this should be changed to
```suggestion
                        "id": tf.int64,  # {'pedstrian':1, 'vehicles':2}
```

armory/data/adversarial/carla_over_obj_det_dev.py
jprokos26: I agree, for some reason this was not reflected in his initial PR but it might have been a minor oversight.

swsuggs:
I also expected that this [line](https://github.com/twosixlabs/armory/blob/67afab29c3c4bd42dc1584d4f4483692a359f556/armory/scenarios/carla_object_detection.py#L83) of the carla scenario would need to be updated since this dataset has fewer classes.

`default_export_kwargs={"with_boxes": True, "classes_to_skip": [4]},`

This is supposed to tell the exporter not to draw the boxes for the green screen.  But, no green screen boxes are being drawn anyway.  Does this dataset not include them?

swsuggs:
For exporting the depth image, we just need to convert from this rgb depth encoding to log grayscale which is what we were using last time.  [convert to meters](https://carla.readthedocs.io/en/latest/ref_sensors/#depth-camera).  [convert to log scale](https://github.com/twosixlabs/armory/blob/67afab29c3c4bd42dc1584d4f4483692a359f556/armory/art_experimental/attacks/carla_obj_det_utils.py#L46)

dxoigmn:
Can you export both? Would be nice to not lose any precision in the depth modality. I agree log depth is useful for visual inspection by humans.

armory/data/adversarial/carla_over_obj_det_dev.py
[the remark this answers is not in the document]
jprokos26: Addressed by c5bd795f235f1d3275ba11467ebc51e92c09cdfd

armory/data/adversarial/carla_over_obj_det_dev.py
jprokos26: Addressed by c5bd795f235f1d3275ba11467ebc51e92c09cdfd

jprokos26:
> I also expected that this [line](https://github.com/twosixlabs/armory/blob/67afab29c3c4bd42dc1584d4f4483692a359f556/armory/scenarios/carla_object_detection.py#L83) of the carla scenario would need to be updated since this dataset has fewer classes.
> 
> `default_export_kwargs={"with_boxes": True, "classes_to_skip": [4]},`
> 
> This is supposed to tell the exporter not to draw the boxes for the green screen. But, no green screen boxes are being drawn anyway. Does this dataset not include them?

You are correct, this dataset does not, see https://github.com/twosixlabs/armory/pull/1605#discussion_r988332459. I am assuming `classes_to_skip` be removed entirely then as such?
```
            default_export_kwargs={"with_boxes": True},
```

swsuggs:
I think we'd like to maintain the behavior for the other carla datasets.  Leaving it as is seems to cause no harm as this dataset has only three classes anyway.  But it would be preferable to infer which class to skip or whether to do so.  Can you see some way to do that?

jprokos26:
> I think we'd like to maintain the behavior for the other carla datasets. Leaving it as is seems to cause no harm as this dataset has only three classes anyway. But it would be preferable to infer which class to skip or whether to do so. Can you see some way to do that?

There doesn't seem to be a clean way to extract this information from `CarlaObjectDetectionTask`. Since the dataset is a `ArmoryDataGenerator` I don't see a way to extract the class names and even if that were obtainable, I'm not sure how to decide which ones need to be skipped based on the name. The easiest way to implement this would be to have a field in the config for `classes_to_skip`, but this may be worth a separate issue.

swsuggs:
Let's leave it for now and create an issue.

jprokos26:
> Did mike create all of these configs? I'm not sure if we really need `carla_obj_det_val_multimodal_adversarialpatch_defended.json` since you can't attack the train set anyway (it doesn't have green screens).

No, mike only created the configs listed in https://github.com/twosixlabs/armory/pull/1607/files but I was unsure if I should include the configs I used to generate the numbers mentioned in this PRs description - I agree it is unecessary. Should I remove all but Mike's configs?

jprokos26:
> It is a lot of configs. Looks like in eval5 we didn't have any example `val` configs. Maybe let's follow the pattern there

Addressed in c11d07e6afc588aef0552d34b4f6165bff3a6643

swsuggs (approved):
Thanks, looks great

PR #1629, opened 2022-10-03
black + yamllint + flake

davidslater:


davidslater:
Here is a fix for ci testing. I commented out the `tool.black` section of `pyproject.toml`. I am keen to merge this in quickly to enable things to work.

Two things on your end:
1) make sure you add in the pre-commit hooks (I found flake and yamllint errors in addition to the black ones)
2) If you want to add in `tool.black`, I think that would be fine, so long as you can find a configuration that doesn't change the current set of files. If we want to make changes to how we format with black, we can do that all at once but I try to minimize that to about once a year, when the new black version comes out.

christopherwoodall (approved):
That looks great!

PR #1630, opened 2022-10-03
Build update

davidslater:
1. restore docs and datafiles to docker build
2. validate version hash syntax
3. calculate fallback version if different.
4. use mamba instead of conda
5. add mamba environment.yml


christopherwoodall (approved):
Looks good!

PR #1631, opened 2022-10-03
Build Updates - Backwards Compatibility

christopherwoodall:
Remove the local `cwd` and `local_git_dir` bind mounts in order to run old container versions. Removing those mounts, by default, would allow for this update to maintain backwards compatibility with older versions. That said, containers version 0.15.4 and prior rely on the git directory to be present for versioning- they mirror the version of the local codebase- and therefore will be labeled as being version "0.0.0".

To actively develop code w/out rebuilding the container, developers can use the `ARMORY_DEV_MODE` and `ARMORY_PRETEND_VERSION` flags. Example:

```bash
$ docker tag 7c65e3e1a61d twosixarmory/pytorch:0.15.4.dev
$ ARMORY_DEV_MODE=1 ARMORY_PRETEND_VERSION="0.15.4.dev" armory run scenario_configs/cifar10_baseline.json --check --use-gpu
```

image


mwartell:
Related to #1630

mwartell:
Per discussion today we should attempt to install [ART extra dependencies](https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/43bfd26874eb6124cc030805c9746077c8a6b901/setup.py#L60) for that which ART needs but armory doesn't

mwartell:
Recap from gard sync today:

1. #1633 
2. reconcile ART and Armory dependencies 
3. document the developer mode and its semantics.

davidslater:
I think the ones that we need are listed in Dockerfile-base. pydub and librosa, but not sure what else we need or if that matches an ART extras line.

christopherwoodall:
Re-examine at a later point

PR #1632, opened 2022-10-04
Baseline poison configs and bug fix

### LINKED ISSUE #1619: Out of memory when running poisoning fairness model
On the develop branch, memory explodes when the poisoning explanatory model runs for the fairness metrics.  So far I've seen this on Cifar10 DLBD configs; haven't tested others yet.
```
File "/workspace/armory/scenarios/poison.py", line 582, in compute_explanatory
self.get_train_majority_mask_and_ceilings()

File "/workspace/armory/scenarios/poison.py", line 532, in get_train_majority_mask_and_ceilings
activations = self.explanatory_model.get_activations(

File "/workspace/armory/metrics/poisoning.py", line 98, in get_activations
x = self.preprocess(x)

File "/workspace/armory/metrics/poisoning.py", line 136, in preprocess
return type(self)._preprocess(

File "/workspace/armory/metrics/poisoning.py", line 130, in _preprocess
return torch.tensor(x).to(device)

RuntimeError: CUDA out of memory. Tried to allocate 27.76 GiB
```
This does not happen on the dev_0.15.4 branch, so the problem should have been introduced sometime after that one split off.

swsuggs:
Adds configs for the baseline defenses for the new poison scenarios.

Also fixes #1619


davidslater (commented):
It looks good, though I think the "backdoor_path" may be brittle and fail if you aren't launching with a matching armory repo. However, I see that's how we're doing it for the backdoors.

Question: is it just librosa that you need the `pytorch-deepspeech` image for?

davidslater:
Gotcha. Then let's move the docker container in the configs to `pytorch`, as librosa is now installed in base container.

The issue with the triggers is that it is assuming that it is being run from armory base directory, and will use the triggers on the host machine, not those in the container.

PR #1636, opened 2022-10-04
Launch / Exec - better defaults

davidslater:
This makes two small changes:
1) Instead of "launch" defaulting to a no-op and shutting down without `--interactive` or `--jupyter` provided, it now defaults to `--interactive`:
```
armory launch pytorch
```
2) Instead of "exec" erroring out when no command line args are given, it defaults to a no-op.
```
armory exec pytorch --
```
Exec will still error if the `--` delimiter is not present.

### DIFF
--- a/armory/__main__.py
+++ b/armory/__main__.py
@@ -668,6 +668,8 @@ def launch(command_args, prog, description):
     (config, args) = arguments.merge_config_and_args(config, args)
 
     rig = Evaluator(config, root=args.root)
+    if not args.interactive and not args.jupyter:
+        args.interactive = True
     exit_code = rig.run(
         interactive=args.interactive,
         jupyter=args.jupyter,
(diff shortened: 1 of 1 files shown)


armory/__main__.py
```diff
@@ -668,6 +668,8 @@ def launch(command_args, prog, description):
     (config, args) = arguments.merge_config_and_args(config, args)
 
     rig = Evaluator(config, root=args.root)
+    if not args.interactive and not args.jupyter:
```
mwartell, line 671: In a common case case, `args.jupyter` is `False`  so this conditional degrades to `args.interactive = True`. That is, it forces interactive mode regardless of what was put on the command line.  Is this the intent?

mwartell (changes requested):

armory/__main__.py
mwartell: Yes, that's exactly the stated intent.

mwartell (approved):

PR #1640, opened 2022-10-05
buildable docker containers

davidslater:
Bare minimum fixes to make `tf2` and `pytorch-deepspeech` containers build properly. Goal is to enable armory release on Friday, as the bigger updates will likely need to wait (e.g., removing the `tf2` and `pytorch-deepspeech` containers).

This removes the TF Object Detection API dependency from both the `tf2` dockerfile and `pyproject.toml`, as it is not actively being supported and causes dependency hell.

`librosa` is now in base, so I removed its install from the `pytorch-deepspeech` dockerfile.

The only other changes were to remove version information from `pyproject.toml` except where explicitly needed.

### DIFF
--- a/docker/Dockerfile-base
+++ b/docker/Dockerfile-base
@@ -22,9 +22,10 @@ RUN apt-get -y -qq update && \
     build-essential \
     git \
     curl \
-#    libgl1-mesa-glx \
+    libgl1-mesa-glx \
 #    libglib2.0-0 \
     && rm -rf /var/lib/apt/lists/*
+# libgl1-mesa-glx is needed for cv2 (opencv-python)
 
 # Install Conda
 RUN wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ~/miniconda.sh && \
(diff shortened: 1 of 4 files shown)


mwartell (commented):
These changes are self-consistent in static review. I am currently testing their behavior.

mwartell:
The CI tests are hitting a satisfiability problem with the unpinned versions causing pip to spend 2+ hours trying to find a solution. Pip itself is complaining that it is taking too long and suggesting that some versions be pinned to help the solver.

The first step is to take the `[all]` target out of the build, because I was able to solve `[developer,pytorch,engine]` locally where `all` was NP-hard. That is running on @christopherwoodall local action simulator to confirm that armory runs with without all.

It might be worthwhile to pin some major versions (e.g. `>=1.*`) to help the solver.

mwartell:
Per discussion in slack I'm going to axe the `[all]` target and update the CI to use the more focused targets as validated by Chris' ad hoc runs. I'm gating this PR on that next to be written PR #TBD

mwartell (approved):

PR #1642, opened 2022-10-05
remove [all] target from pyproject and CI because it lags CI badly

mwartell:
The `[all]` target causes 2+ hour pip installations because it is computationally hard when versions are unpinned. Also, we don't need such an installation.


christopherwoodall (approved):
While I still think the "all" target is valid, I see no issue in removing it.

PR #1643, opened 2022-10-05
Total wer fix

davidslater:
Fixes #1639 

Problem was due to an overly restrictive check, which I have relaxed.

PR #1644, opened 2022-10-05
Hubert ASR - replacement for deep speech without the dependencies

davidslater:
This will replace the Deep Speech 2 model with Hubert. It is more modern, better performing, and just uses basic `torchaudio`.


davidslater:
Should be able to test with:
```
armory run scenario_configs/eval6/asr_librispeech/hubert_untargeted_snr_pgd.json --check
```

davidslater:
Note: this depends on #1643 to work.

swsuggs (approved):
Runs for me.

PR #1646, opened 2022-10-06
remove some version warts from CI

### LINKED ISSUE #1645: CI failure - version yields "2202-10-06"
In https://github.com/twosixlabs/armory/actions/runs/3193160010 and others we get 

    docker: Error response from daemon: manifest for twosixarmory/pytorch:2022-10-05 not found: manifest unknown: manifest unknown.

This is trying to parse a log message as a version.

mwartell:
While investigating #1645 I uncovered some flakiness in CI which we've been stumbling on. These changes clean up
`3-test-docker.yml` adds the `armory --show-docker-version-tag` command and better defines an error return in `version.py`.

Because these are fiddly, I wanted review of these on their own. 

~Fixes #1645~

### DIFF
--- a/.github/workflows/3-test-docker.yml
+++ b/.github/workflows/3-test-docker.yml
@@ -19,15 +19,15 @@ jobs:
 
       # Get history and tags for SCM versioning to work
       # yamllint disable rule:indentation
-      - name: 📩 Checkout the Source
-        run: |
-          git fetch --prune --unshallow || true
-          git fetch --depth=1 origin +refs/tags/*:refs/tags/*
-          # Verify .git folder
-          if [ ! -d "./.git" ]; then
-            echo "Expected ./.git folder to exist"
-            exit 1
-          fi
+    #   - name: 📩 Checkout the Source
+    #     run: |
+    #       git fetch --prune --unshallow || true
+    #       git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+    #       # Verify .git folder
+    #       if [ ! -d "./.git" ]; then
+    #         echo "Expected ./.git folder to exist"
+    #         exit 1
+    #       fi
       # yamllint enable rule:indentation
 
       - name: 🐍 setup python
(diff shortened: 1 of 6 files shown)


.github/workflows/3-test-docker.yml
```diff
@@ -54,7 +51,8 @@ jobs:
       # yamllint disable rule:indentation
       - name: 🤞 Run Image tests
         run: |
-          ARMORY_VERSION=`armory --version | sed -r 's/\+/\./g'`
+          ARMORY_VERSION=`python -m armory.utils.version`
```
davidslater, line 54: Why can't you just do `python -m armory --version`?

mwartell: That was a WIP hack where I added an custom `if __main__` to version.py. It is not going into this PR when ready for review.

mwartell: Do you have an opinion on using `python -m armory.utils.get_docker_version` to factor out the postprocessing? Where get_docker_version.py is
```
# docs…
if __name__…
    print(to_docker_tag(get_version()))
```

davidslater: It would be nice to use the same entrypoint in `__main__`, but I like having a `get_docker_version` function.

Also nice not to have:
```
/home/anaconda/anaconda/conda_envs/david.slater/armory39/lib/python3.9/runpy.py:127: RuntimeWarning: 'armory.utils.version' found in sys.modules after import of package 'armory.utils', but prior to execution of 'armory.utils.version'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))
```

It would only be a couple of lines in main to add the docker version using the get_docker_version function.
https://github.com/twosixlabs/armory/blob/develop/armory/__main__.py#L747-L763

mwartell: This is now implemented as `armory --show-docker-version-tag` where the long, descriptive name is intentional. I am not documenting this in `Usage`

christopherwoodall (approved):

mwartell:
Closed by #1653

PR #1648, opened 2022-10-06
Carla mutli-object tracking

swsuggs:
Pulling together Mike's PRs #1624 #1625 #1626 #1627.

The config `carla_mot_adversarialpatch_undefended.json` is runnable.

Todo:
- [x] scenario run_benign() and run_attack(), at least
- [x] dataset preprocessing (annotation format)
- [ ] dataset tests
- [ ] dataset docs
- [x]  metric functions (metrics/task.py)
- [ ] revise metrics to use instrument framework
- [ ] export functions to accommodate new y format


swsuggs:
Scenario runs and calculates metrics (but does not use the instrument framework).  

Really the only reason we need a new scenario is that `carla_video_tracking.py` assumes that `model.predict()` has a `y_init` kwarg, which the ByteTrack FRCNN does not.

swsuggs:
@davidslater One reason we talked about changing the format of the `y` data objects was in case it was easier to get `y_init` for the model.  This model is based on an object detector with a Kalman filter and does not have `y_init` like the goturn model.  I'm not inclined to modify the format of `y` without a clearly better alternative.  The current format is what the baseline model uses and it seems reasonable that performers can build their models to accommodate that.  

As far as exporting, I imagine the scenario could preprocess the `y` so that the old video exporter can be used.  Object ID and class do not matter for exporting since all the boxes will look the same.  Unless, maybe it would be nice to use a different color of box for each tracked object?

davidslater (approved):
If this runs end-to-end, let's merge in, and then have separate issues/PRs for the remaining items.

swsuggs:
Created issues #1655, #1656, #1657 for the remaining tasks.

PR #1647, opened 2022-10-06
Armory Release Workflow 

christopherwoodall:
Updates Armory's release workflow.

PR #1635, opened 2022-10-04
version take 2

davidslater:

PR #1653, opened 2022-10-07
version and ci omnibus merge

mwartell:
This is an omnibus merge from three separate workbranches

  1. #1635 was merged to `develop` 
  2. I then merged `develop` into #1647
  3. I merged #1646 merged into that

yielding this `mwartell/version-merges-by-three` branch. The purpose of each of these were to correct various build and release issues.
 
There were some minor conflicts but I believe I captured the proper intent from each. 

Closes #1646, closes #1647 because it incorporates all their changes.


christopherwoodall (commented):
LGTM!

armory/__main__.py
```diff
@@ -760,6 +761,10 @@ def main():
     elif sys.argv[1] in ("-v", "--version", "version"):
         print(f"{armory.__version__}")
         sys.exit(0)
+    elif sys.argv[1] == "--show-docker-version-tag":
+        version = armory.utils.version.get_version()
+        print(armory.utils.version.to_docker_tag(version))
```
davidslater, line 766: Do we want to make this use the same pattern as `--version`, replacing this with
```
print(armory.utils.version.to_docker_tag(armory.__version__))
```
?

davidslater (approved):
One comment, but I'm fine with it either way.

armory/__main__.py
mwartell: https://github.com/twosixlabs/armory/issues/1654

Issue #1655, opened 2022-10-07
Carla MOT -- add docs and tests for new dataset


swsuggs:

swsuggs:
Completed #1658.

PR #1658, opened 2022-10-07
Carla mot dataset

swsuggs:
For #1655


davidslater (changes requested):
Test is good. 

Can you write a description of the `y` format somewhere in the docs? The 9 different values is super hard to grok.

davidslater (changes requested):
I'm looking for more information for the user - maybe a line per item. How are multiple detections in a single timestep handled? Are they all time ordered? What is the differenece between object_id and class_id? Etc.

davidslater (approved):
LGTM

PR #1663, opened 2022-10-09
Update pyproject.toml

christopherwoodall:
Scenario `/eval6/carla_mot/carla_mot_dpatch_defended.json` imports [`carla_mot_frcnn_byte.py`](https://github.com/christopherwoodall/armory/actions/runs/3215159957/jobs/5256064280#step:4:78) which in turn imports `thop`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,6 +93,7 @@ pytorch = [
     "torch",
     "torchvision",
     "torchaudio",
+    "thop",
 ]
 
 tensorflow = [


mwartell (approved):
Adds `thop` as described.

davidslater:
That file actually has numerous issues, and requires a number of installs. I would rather keep it out of the main requirements for now and have a separate `flavor` (`mot`?) that is just mot focused, that installs all of those. @swsuggs can tell you the full list of requirements. I know if involves cython

swsuggs:
The carla MOT scenario requires `thop`, `lap`, `cython`, and `cython-bbox`, as well as git repos `JonathonLuiten/TrackEval`, `colour-science/colour`, and `ifzhang/ByteTrack`

davidslater:
Requirements are:
```
thop
lap
Cython
git+https://github.com/ifzhang/ByteTrack.git
cython-bbox
```

Note: it seems like installing cython-bbox and Cython in the same pip call results in an error, as the former relies on the latter. I'm not sure if there is a way to specify order of installation in the pip call, but that would probably fix that problem.

davidslater:
Let's pass on `JonathonLuiten/TrackEval`, unless we can get pip to ignore its overly specific requirements:
https://github.com/JonathonLuiten/TrackEval/blob/master/requirements.txt

For `colour-science/colour`, we have been using tag `@v0.3.16`, so `colour-science/colour@v0.3.16`.

However, neither of those are truly needed, as they work fine as git downloads in the sysconfig.

davidslater:
Same issue for `colour`, though it is FAR worse:
https://github.com/colour-science/colour/blob/v0.3.16/requirements.txt

I think we just don't pip install them, and rely on git download in scenarios.

mwartell:
Possibly obsoleted by #1667

mwartell:
I've merged #1667 into this branch am am now adding the needed CI.

mwartell:
Need to patch-up

- [ ] docker/build.py
- [ ] workflows/3-test-docker.yml gets a copy-paste from pytorch build
- [ ] workflows/release.yml gets a copy-paste from pytorch release
- [ ] armory/docker/images.py
- [ ] `armory/__main_.py`  argument support
- [ ] `test/unit/test_docker_build_script.py`

mwartell:
@swsuggs I have tagged you for review on this but am only seeking your awareness as you may need these changes for your MOT work.

The changes are:

   1. add the `twosixarmoy/carla-mot` image to build.py
   2. register `carla-mot` with images.py
   3. update `scenario_configs/eval6/carla_mot/*` to match the blessed `twosixlabs/carla-mot` name
   4. update build, release, test CI jobs for new image

docker/Dockerfile-carla-mot
```diff
@@ -0,0 +1,35 @@
+ARG base_image_tag
+
+FROM twosixarmory/pytorch:${base_image_tag} AS armory-bytetrack
```
swsuggs, line 3: Should `armory-bytetrack` be changed to `carla-mot` here as well?  not sure if it matters

mwartell: Good catch. I'm mostly sure it doesn't matter, but fixing that will stop the next person from having to wonder.

mwartell: The `AS` clause is only needed for multi-stage builds, and was not needed here.

[Dockerfile reference](https://docs.docker.com/engine/reference/builder/#from:~:text=Optionally%20a%20name%20can%20be%20given%20to%20a%20new%20build%20stage%20by%20adding%20AS%20name%20to%20the%20FROM%20instruction.%20The%20name%20can%20be%20used%20in%20subsequent%20FROM%20and%20COPY%20%2D%2Dfrom%3D%3Cname%3E%20instructions%20to%20refer%20to%20the%20image%20built%20in%20this%20stage.)

docker/Dockerfile-carla-mot
```diff
@@ -0,0 +1,35 @@
...
+# NOTE: This COPY command is filtered using the `.dockerignore` file
+#       in the root of the repo.
+COPY ./ /armory-repo
+
+RUN echo "Building Armory from local source" && \
+    pip install --no-compile --no-cache-dir --editable . && \
+    echo "Configuring Armory..." && \
+    armory configure --use-default && \
+    echo "Cleaning up..." && \
+    rm -rf /armory-repo/.git
```
davidslater, line 33: We're building from `pytorch`, not `base`, so these lines should be unnecessary.

davidslater: (lines 16-33)

davidslater (changes requested):
Just one change.

docker/Dockerfile-carla-mot
christopherwoodall: They are necessary until the updated images are uploaded into dockerhub.

christopherwoodall: See commits [6e17af8](https://github.com/twosixlabs/armory/pull/1663/commits/6e17af8f66787ee787e60ada28efadb90ca32f83) [c8cff4c](https://github.com/twosixlabs/armory/pull/1663/commits/c8cff4c3dfbb46de2810ca120f4b580ab96f3bf7) [9d16176](https://github.com/twosixlabs/armory/pull/1663/commits/9d161766b04b38ad62e2e8e11eb05fd46d4b4686)

mwartell: A clarification for my sake: because the carla-mot dockerfile starts with `FROM twosixlabs/pytorch` the CI build gets the latest `pytorch` which has version 0.15.4. This apparently redundant "Building Armory from local source" ensures that code and version are refreshed.

If you run `docker build pytorch` followed by `docker build carla-mot` on your local, *stateful* machine, this step is not needed. Because the CI including release is *stateless* between runs this is needed.  There was a [CI failure](https://github.com/twosixlabs/armory/actions/runs/3221627521/jobs/5270015780) earlier which shows the fault.

mwartell: Flagged for later build time optimization in #1671

davidslater: Hmmm. Maybe we should build from the `base` image, then, instead of `pytorch`.

Issue #1657, opened 2022-10-07
Carla MOT -- export videos


swsuggs:
We either need a new MOT exporter to support the format of multi-object predictions, or perhaps the scenario can process the predictions so that the regular carla video exporter can be used

PR #1665, opened 2022-10-10
Export carla MOT

swsuggs:
For #1657: update video exporter to allow for multi-object predictions


davidslater (changes requested):
The export looks good.

I would like you to change how the expand_dims is handed. In particular, `y` and `y_pred` should always have the batch dimension, and the model should always take a batch `x` or `x_adv`.

Can you modify the _model_ file so that it is the one that does the indexing `x[0]` and `np.expand_dims` at the end? Also, modify the task metric so that it does the indexing `y[0], y_pred[0]` in the metric?

davidslater (approved):
Looks good, but can you get someone else to test?

PR #1667, opened 2022-10-10
dockerfile for MOT

davidslater:


scenario_configs/eval6/carla_mot/carla_mot_dpatch_undefended.json
```diff
@@ -62,7 +62,6 @@
     "sysconfig": {
         "docker_image": "twosixsarmory/armory-bytetrack",
         "external_github_repo": [
-            "ifzhang/ByteTrack",
```
swsuggs, line 65: Could you do this on the other two mot configs too?

swsuggs (changes requested):

scenario_configs/eval6/carla_mot/carla_mot_dpatch_undefended.json
davidslater: Done.

swsuggs (approved):

PR #1669, opened 2022-10-10
WIP: MOT: update task results

davidslater:
This does *not* get things into the meter format. However, it should now produce the correct set of results to the output dict and print the correct info to screen.

### DIFF
--- a/armory/scenarios/carla_mot.py
+++ b/armory/scenarios/carla_mot.py
@@ -6,6 +6,7 @@ CARLA Multi-Object Tracking Scenario
 from armory.scenarios.carla_video_tracking import CarlaVideoTracking
 
 from armory.metrics.task import HOTA_metrics
+from armory.logs import log
 
 
 class CarlaMOT(CarlaVideoTracking):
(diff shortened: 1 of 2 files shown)


armory/scenarios/carla_mot.py
```diff
@@ -6,20 +6,29 @@
...
             "tracked_classes", ["pedestrian"]
         )
+        super().__init__(config, **kwargs)
+
+    def load_metrics(self):
+        all_tasks = self.config["metric"]["task"]
+        hotas = ["hota", "deta", "assa"]
+        self.hota_tasks = [t for t in all_tasks if t in hotas]
+        self.config["metric"]["task"] = [t for t in all_tasks if t not in hotas]
+        super().load_metrics()
```
swsuggs, line 24: This now throws an error because `config["metrics"]["task"]` is empty and that case is not currently considered in `instrument.config`.  It kind of prevents you from working outside the framework like this.
```
File "/workspace/armory/instrument/config.py", line 60, in __init__
construct_meters_for_task_metrics(
└ <function construct_meters_for_task_metrics at 0x7fc7ff4b0430>
File "/workspace/armory/instrument/config.py", line 345, in construct_meters_for_task_metrics
benign, adversarial, targeted = zip(
*tuples)
                                     └ []
ValueError: not enough values to unpack (expected 3, got 0)
```
(That was one reason I had those little placeholder task metric functions.)

swsuggs:
When I restore the placeholders, the rest of the code runs and the metrics are stored in the json file.  Do you hope to get this into the meter framework before the release?  If not, we could modify instrument/config.py to allow an empty task metric list (less desirable since it's a one-off situation), or just make some trivial placeholder task metric function.

PR #1672, opened 2022-10-11
Mot metrics meter

davidslater:
Should supersede #1669


armory/metrics/task.py
```diff
@@ -1715,3 +1697,60 @@ def get_per_class_per_video_metrics(self):
...
+        self.means = bool(means)
+        self.record_metric_per_sample = bool(record_metric_per_sample)
+
+        self.hota_metrics = HOTA_metrics(**kwargs)
+
+    def __call__(self, y_list, y_pred_list):
+        for i, (y, y_pred) in enumerate(zip(y_list, y_pred_list)):
+            for tracked_class in self.hota_metrics.tracked_classes:
+                self.hota_metrics.calculate_hota_metrics_per_class_per_video(
+                    y[0], y_pred[0], tracked_class, i
```
swsuggs, line 1726: y and y_pred should not be indexed.  They have a batch dimension but you are already iterating over that.

armory/metrics/task.py
```diff
@@ -1715,3 +1697,60 @@ def get_per_class_per_video_metrics(self):
...
+                )
+
+        for tracked_class in self.hota_metrics.tracked_classes:
+            self.hota_metrics.calculate_hota_metrics_per_class_all_videos(tracked_class)
+
+        results = {}
```
swsuggs, line 1732: Inconsistency with `results` vs `self.results`

armory/metrics/task.py
```diff
@@ -1715,3 +1697,60 @@ def get_per_class_per_video_metrics(self):
...
+        if self.record_metric_per_sample:
+            per_class_per_video_metrics = (
+                self.hota_metrics.get_per_class_per_video_metrics()
+            )
+            for tracked_class in self.hota_metrics.tracked_classes:
+                for k in ["hota", "deta", "assa"]:
+                    self.results[f"benign_{k}"] = []
```
swsuggs: Hard coded for "benign" but this function is also called for adversarial metrics.

swsuggs:
It may be desirable to log at least the mean metric values to the terminal.

armory/metrics/task.py
davidslater: Fixed

swsuggs (approved):
Awesome, everything is working for me.

PR #1682, opened 2022-10-12
document hyrda-lightning install for no-docker users

### LINKED ISSUE #1677: breaking change to deepspeech in v0.16.0
In v0.16.0 we had moved all dependency declarations into pyproject.toml. This works fine for `pip install`. However, PyPI.org does not accept the declaration for [romesco/hydra-lightning](https://github.com/romesco/hydra-lightning).

The romesco dependency is installed with

    pip install git+https://github.com/romesco/hydra-lightning/#subdirectory=hydra-configs-pytorch-lightning

But when publishing the wheel to PyPI that "direct reference" is not acceptable.  To fix that we created #1676 which moves that dependency to the armory deepspeech container and removes it from pyproject. 

This may be a problem for no-docker users.

mwartell:
Fixes #1677

### DIFF
--- a/docs/no_docker_mode.md
+++ b/docs/no_docker_mode.md
@@ -2,7 +2,7 @@ Armory No-Docker Mode
 =======================
 In order to run armory in `--no-docker` mode, you will need a properly
 setup environment.  Generally folks have used conda in the past, however this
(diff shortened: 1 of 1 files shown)


christopherwoodall (approved):
LGTM!

PR #1692, opened 2022-10-13
use armory.__version__ as the standard access

### LINKED ISSUE #1654: increase parallelism of calls to version mechanism
Do we want to make this use the same pattern as `--version`, replacing this with
```
print(armory.utils.version.to_docker_tag(armory.__version__))
```
?

_Originally posted by @davidslater in https://github.com/twosixlabs/armory/pull/1653#discussion_r990444602_
 
Yes, I should do that. And will happily do so outside the context of this pull request.

mwartell:
Fixes #1654


christopherwoodall (commented):
Would it be beneficial to wrap the output of the version methods in `version.py` so that the output is uniform both when calling an entry-point and when importing?

PR #1703, opened 2022-10-17
WIP: Adapt

davidslater:


lcadalzo:
I'm running the following code:

```
model = cifar10_model()
xy_batches, classes = cifar10_data()
attack = PGD_Linf(model)

for batch in xy_batches:
    x, y = batch
    x_adv = attack(x)
```

and getting the following error:
```
  File "run.py", line 9, in <module>
    x_adv = attack(x)
  File "/home/lucas/Desktop/gard/git/twosixlabs/armory/armory/adapt/pytorch.py", line 644, in __call__
    self.gradient()
  File "/home/lucas/Desktop/gard/git/twosixlabs/armory/armory/adapt/pytorch.py", line 582, in gradient
    self.loss = self.loss_fn(self.y_pred, self.y_target)
  File "/home/lucas/miniconda3/envs/armory-adapt/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
  File "/home/lucas/Desktop/gard/git/twosixlabs/armory/armory/adapt/pytorch.py", line 511, in forward
    return super().forward(torch.log(input), target)
  File "/home/lucas/miniconda3/envs/armory-adapt/lib/python3.8/site-packages/torch/nn/modules/loss.py", line 211, in forward
    return F.nll_loss(input, target, weight=self.weight, ignore_index=self.ignore_index, reduction=self.reduction)
  File "/home/lucas/miniconda3/envs/armory-adapt/lib/python3.8/site-packages/torch/nn/functional.py", line 2689, in nll_loss
    return torch._C._nn.nll_loss_nd(input, target, weight, _Reduction.get_enum(reduction), ignore_index)
RuntimeError: 0D or 1D target tensor expected, multi-target not supported

```

which I think occurs because `self.y_target` is an array of shape `torch.Size([1, 10])` (the same one as `self.y_pred`) rather than an integer

davidslater:
Thanks, @lcadalzo, I hadn't tested with model predictions.

armory/adapt/pytorch.py
```diff
@@ -0,0 +1,1126 @@
...
+
+Design goals:
+    1) Easy to understand
+    2) Easy to modify
+    3) Easy to compose
+    4) Fast and efficient
+    5) Minimal dependencies (framework-specific)
+
+User stories:
+    a) Initialize an attack with an interesting input
```
lcadalzo, line 12: For the user story `Initialize an attack with an interesting input`, not sure I'm following exactly what's meant since you can pass whatever to ART's `generate()` or in the case here,`__call__()`. Or do you mean the init within the epsilon ball that gets added to the input? If so, it looks like this code restricts you to either random or None.

davidslater: Yes, the desire is to enable the init within the epsilon ball. The code needs to be modified to enable passing in an init function instead of those two choices.

armory/adapt/pytorch.py
```diff
@@ -0,0 +1,1126 @@
...
+            "prob": ProbLoss(),
+        }
+        if output not in self.output_map:
+            raise ValueError(f"output {output} not in {list(self.output_map)}")
+        self.output = output
+        self.loss_fn = self.output_map[self.output]
+
+        # logits:
+        self.task_metric = lambda y_pred, y_true: (
+            y_pred.detach().argmax(dim=1) == y_true.detach()
```
lcadalzo, line 545: just pointing out a challenge here. While setting this task_metric is simple enough for classification, this would break for non-classification tasks and is much less straightforward to define in those cases

armory/adapt/pytorch.py
```diff
@@ -0,0 +1,1126 @@
...
+    if len(shape) != len(mask_shape):
+        raise ValueError(f"shape {shape} not consistent with mask_shape {mask_shape}")
+    for i, j in zip(shape, mask_shape):
+        if j > i:
+            raise ValueError(f"mask_shape {mask_shape} exceeds shape {shape}")
+        position.append(np.random.randint(i - j + 1))
+    return tuple(position)
+
+
+class PGD_RandomPatch(PGD_Patch2):
```
lcadalzo, line 924: is it necessary to have an additional class for patch pgd _and_ random patch pgd, as opposed to this just being something set or defined in the former?

armory/adapt/pytorch.py
```diff
@@ -0,0 +1,1126 @@
...
+# Create graph of input/output variables and connections?
+
+# attack = Attack([
+#     Input(x_orig),
+#     X_init(x_orig,
+#     Output(x_adv),
+# ])(epsilon)
+
+
+class EoTPGD_Linf(PGD_Linf):
```
lcadalzo, line 1017: it seems like the number of distinct classes needed is multiplying rapidly. Would you also need a class for EoT L2, EoT Patch, EoT random patch, etc.?

davidslater: I should have added a note on this one: this was meant to demonstrate how to extend a class to use EoT. I would remove this class, actually. Or, if EoT can be added as a wrapper around an existing class, then that would prevent that sort of class explosion.

I need a better way of handling / representing random components - that would enable specification of what you're performing the expectation over.

armory/adapt/pytorch.py
davidslater: Point taken

PR #1705, opened 2022-10-21
merge rc0.16.1 changes into develop

mwartell:
Chris and I would like to merge the release candidate changes into develop now rather than waiting until we release v0.16.1 to merge. We will, of course also merge that branch into develop upon release.

The rc0.16.1 is quiet while waiting on some known work, but we'd like to not base new workbranches on a maintenance branch. I believe this to be a zero risk operation but wanted to apprise others before doing so.


christopherwoodall (approved):
LGTM!

davidslater (approved):

PR #1710, opened 2022-10-25
testing cifar stem

### LINKED ISSUE #1689: [BUG]: ResNet Stem on CIFAR Models
### Description of the bug

The model architectures used for evaluation in eval 5 and eval 6 for poisoning with CIFAR-10 are not optimal. If correct, then this might be the reason why these models perform less well than expected in benign and robust accuracy.

Configurations in `https://github.com/twosixlabs/armory/tree/master/scenario_configs/eval5/poisoning/baseline_defenses/cifar10`, for example `https://github.com/twosixlabs/armory/blob/master/scenario_configs/eval5/poisoning/cifar10_witches_brew.json` describe the model architecture as `armory.baseline_models.pytorch.resnet18` with input shape `[32,32,3]`. However, these ResNet18s are defined in `https://github.com/twosixlabs/armory/blob/master/armory/baseline_models/pytorch/resnet18.py` and draw directly from the pytorch ImageNet implementations for ResNet18 models.

The PyTorch ResNets are not well-suited to inputs of shape [32,32,3] as they contain a downsampling stem (with a 7x7 convolution of stride 2, and a 3x3 maxpool). A better stem would be to have a single 3x3 convolution without stride instead. This type of CIFAR-stem is common in literature when using ResNet-18 models on CIFAR-10 and should reduce the error rate for classification from 25% to about 5%. It is possible that the stem is modified somewhere else in the code (in which case this issue is wrong), but I could not find it in the code.

This issue also applies to the models described for eval6.

### Steps To Reproduce

Refer to `armory.baseline_models.pytorch.resnet18`.

### Additional Information

This issue could be fixed by loading the ResNet model as done currently, but replacing the `.conv1` module with a 3x3 convolution without stride, and the `.maxpool` module with a `torch.nn.Identity` for model configured for inputs with shape [32,32,3].

davidslater:
Fixes #1689


lcadalzo:
@davidslater does this need a review? If so, is there a corresponding issue?

davidslater:
I was in progress on this, but hadn't reviewed results. It works, it's just a question of impact on poisoning.

swsuggs:
Looks great to me.

PR #1711, opened 2022-10-25
cleanup mechanism for sourcing repo local files

mwartell:
This is a collection of three related commits 0451b347 3f809552 ee0e1abf that were made to branch develop but that we'd like in rc0.16.1. I have cherry-picked them into this work branch for separate review.

The changes create an `armory.SRC_ROOT` variable that is of general utility to code that wants to make root-relative references. It also adds a `armory.utils.triggers.TRIGGERS_DIR` pathlib.Path which contains binary files used mostly by poisoning scenarios. These replace hardcoded paths containing confusing `../../..` style paths.

### DIFF
--- a/armory/__init__.py
+++ b/armory/__init__.py
@@ -1,11 +1,17 @@
 """
 Adversarial Robustness Evaluation Test Bed
 """
+
+from pathlib import Path
+
 from armory.logs import log
 from armory.utils import version, typedef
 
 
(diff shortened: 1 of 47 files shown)


mwartell:
I don't like this being a new commit, because it means duplicate code when we merge back into develop. Will git do the right thing? I'm going to make a serial merge PR to make it easier to resolve.

PR #1713, opened 2022-10-25
change output dir of CARLA video tracking exporter

### LINKED ISSUE #1712: CARLA Video Tracking exporting boxes to multiple locations
There are two `ExportMeters` set up for the CARLA video tracking scenario: one that outputs images with boxes and one that doesn't. The latter `ExporMeter` is outputting to, e.g., `/armory/outputs/2022-10-25T183719.308044/` whereas the former is (correctly) outputting to `/armory/outputs/2022-10-25T183719.308044/saved_samples`. Both sets of boxes should be outputted to the same subdirectory

lcadalzo:
Closes #1712


davidslater (approved):
LGTM

PR #1714, opened 2022-10-25
create explicit paths for repo-relative file access

mwartell:
This is a collection of three related commits 0451b347 3f809552 ee0e1abf that were made to branch develop but that we'd like in rc0.16.1. I have merged them into this work branch for separate review. Because 183ecc9 and 3833096 were present in develop prior to the merge, they came along for the ride. The only difference is the addition of a documentation line from #1706.

The changes create an `armory.SRC_ROOT` variable that is of general utility to code that wants to make root-relative references. It also adds a `armory.utils.triggers.TRIGGERS_DIR` pathlib.Path which contains binary files used mostly by poisoning scenarios. These replace hardcoded paths containing confusing `../../..` style paths.


mwartell:
Closing in favor of just merging develop at ee0e1ab. Interestingly, github saw that the net effect was the same and showed the CI results from yesterday. https://github.com/twosixlabs/armory/actions/runs/3315542723

PR #1715, opened 2022-10-25
Mot coco

### LINKED ISSUE #1699: Reformat labels for multiobject tracking?

davidslater:
Fixes #1699

Currently, it only provides functions for mapping to and from mot / coco format.

To be fully featured, it needs to enable metrics to be produced from coco format.


davidslater:
Currently, issues are:
1. ClipVideoTrackingLabels is wrong - it assumes one label per frame
2. The HOTA metrics make an assertion that every frame has a prediction
3. CARLA label preprocessing needs to take the tuple into account

lcadalzo:
If I naively run `scenario_configs/eval6/carla_mot/carla_mot_adversarialpatch_undefended.json` from the `develop` branch it runs ok and outputs sensible metrics results. I assume this was the case for you too, and that each of the issues in that list was only apparent upon closer inspection? If there's any nontrivial steps (i.e. beyond just pdb'ing through the code) needed to reproduce the issues you saw initially, any description of that would be helpful

davidslater:
Some issues I noted as issues without seeing them pop up (ClipVideoTrackingLabels, HOTA metrics assertion), so I never triggered those as bugs. Others were due to issues I had to fix in A) adding the COCO format as an option and B) switching from 1-indexing to 0-indexing.

If it's helpful, I can talk through the changes I made.

lcadalzo:
Not quite sure why this didn't come up when I tested earlier, but when I run `scenario_configs/eval6/carla_mot/carla_mot_adversarialpatch_undefended.json` from `develop`, I get the following:

```
  File "/opt/conda/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/opt/conda/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/workspace/armory/scenarios/main.py", line 228, in <module>
    run_config(
  File "/workspace/armory/scenarios/main.py", line 142, in run_config
    scenario = get(*args, **kwargs)
  File "/workspace/armory/scenarios/main.py", line 134, in get
    scenario = ScenarioClass(config, **kwargs)
  File "/workspace/armory/scenarios/carla_mot.py", line 15, in __init__
    self.tracked_classes = config.get("scenario", {}).get(
  File "/workspace/armory/scenarios/carla_video_tracking.py", line 16, in __init__
    super().__init__(*args, **kwargs)
TypeError: __init__() got an unexpected keyword argument 'tracked_classes'

```

Although it looks like you have a fix in place, so I'll copy your updated `__init__()` over to develop so I can reach the other issues

armory/data/adversarial_datasets.py
```diff
@@ -830,6 +830,29 @@ def __call__(self, x, labels):
         return boxes, patch_metadata_dict
 
 
+class ClipMOTVideoTrackingLabels(ClipVideoTrackingLabels):
```
lcadalzo, line 833: How/where is `ClipMOTVideoTrackingLabels` used?

```
(armory) lucas@ubu0953:~/Desktop/gard/git/twosixlabs/armory$ grep -r "ClipMOTVideoTrack" .
Binary file ./armory/data/__pycache__/adversarial_datasets.cpython-38.pyc matches
./armory/data/adversarial_datasets.py:class ClipMOTVideoTrackingLabels(ClipVideoTrackingLabels):
(armory) lucas@ubu0953:~/Desktop/gard/git/twosixlabs/armory$ 
```

lcadalzo: I also think its __init__() is identical to that of the parent class it overrides

lcadalzo: if `y` has shape `(batch_size, num_detections, 9)`, I'm not seeing how `(y[y[:, :, 0] < self.max_frames]` makes sense as it looks like it's only indexing the first of the 9

lcadalzo: I'd think [here](https://github.com/davidslater/armory/blob/mot-coco/armory/data/adversarial_datasets.py#L1115) is the place the class should be referenced

davidslater: Good catch on all of the above, will change.

davidslater: It indexes the first of the 9 because that's the frame_id field, so it's using it to create an index.

davidslater:
Okay, should be good to go now.

scenario_configs/eval6/carla_mot/carla_mot_adversarialpatch_undefended_coco.json
```diff
@@ -0,0 +1,82 @@
+{
```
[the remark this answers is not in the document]
lcadalzo: perhaps going a step further, is there a need for having the `coco_format` parameter to toggle between in the first place? As opposed to just having things always be in coco format. This would simplify things a bit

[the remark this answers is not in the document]
davidslater: That is true. I kept them in because the `coco_format` flag goes so many different places it's easy to mess up.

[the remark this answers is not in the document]
davidslater: There is so much hardcoded MOT format stuff, that it's probably useful to have ways of toggling that. What about having `"coco_format": true` for all the configs and removing the non-coco format ones? Then, if someone really wants to use MOT format, they know exactly which things to switch.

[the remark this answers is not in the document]
lcadalzo: yea i'm good with that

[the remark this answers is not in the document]
davidslater: Done

armory/metrics/task.py
```diff
@@ -1645,32 +1652,34 @@ def preprocess(self, gt_data, tracker_data, tracked_class):
...
-        if len(unique_tracker_ids) > 0:
-            unique_tracker_ids = np.unique(unique_tracker_ids)
-            tracker_id_map = np.nan * np.ones((np.max(unique_tracker_ids) + 1))
-            tracker_id_map[unique_tracker_ids] = np.arange(len(unique_tracker_ids))
-            for t in range(num_timesteps):
-                if len(data["tracker_ids"][t]) > 0:
-                    data["tracker_ids"][t] = tracker_id_map[
-                        data["tracker_ids"][t]
-                    ].astype(int)
+        def relabel(unique_ids, num_timesteps, sub_data):
```
lcadalzo, line 1655: it's a little difficult to grok what's going on with this function. If there are any comments or variable renaming that you think could help with this, that'd be nice to add but otherwise it appears the function is doing what it's supposed to be

lcadalzo:
one very minor comment lingering above but otherwise LGTM

armory/metrics/task.py
davidslater: I think the issue was the excessive use of numpy when unnecessary. Take a look now.

armory/metrics/task.py
```diff
@@ -1587,10 +1587,10 @@ def _check_and_format(self, data):
...
     def relabel(unique_ids, list_of_id_arrays):
         """
-        Re-label int IDs to 1-indexed sequential IDs
+        Re-label int IDs to 0-indexed sequential IDs
         """
         unique_ids = sorted(set(unique_ids))
-        new_unique_ids = list(range(1, len(unique_ids) + 1))
+        new_unique_ids = list(range(len(unique_ids)))
         id_map = {k: v for k, v in zip(unique_ids, new_unique_ids)}
         new_list = [np.array([id_map[x] for x in array]) for array in list_of_id_arrays]
```
lcadalzo, line 1595: is it important that the `new_list` ids are in sequential order? Or just that the range of ids is a consecutive span of ints? When I look at this variable, I notice it's not quite in order and not sure if your use of "sequential" in the comment implies it should be

```
(Pdb) new_list[2]
array([ 0, 21,  1,  3,  4,  5,  6,  7,  8, 24,  9, 10, 25, 11, 12, 13, 14,
       15, 22, 16, 17, 26, 18, 19, 20])
```

armory/metrics/task.py
lcadalzo: looks a lot more readable now

armory/metrics/task.py
davidslater: Not for each array in the list. Just that the range of ids is consecutive ints (over the entire list of arrays). Reordering them in the arrays would require reordering them in the other data structures, which I would like to avoid.

PR #1725, opened 2022-10-27
Adds eval 6 to the scenario runner

christopherwoodall:
- Adds the `eval6` directory
- Fixes path error


mwartell (approved):
corrects some minor errors in the scenario-runner

mwartell (approved):

Issue #1666, opened 2022-10-10
Add ability to export depth image in both RGB and Grayscale Log format


jprokos26:
See https://github.com/twosixlabs/armory/pull/1628#issuecomment-1268878568 for context.

Within the `ObjectDetectionExporter`'s [_export()](https://github.com/twosixlabs/armory/blob/develop/armory/instrument/export.py#L111) method, there is no way to tell if `x` is a depth image or not and therefore we cannot tell if we should export the image in both standard RGB as well as in grayscale log format. This is not an issue when using `both` as modality, but is when only using `depth`.

An alternative to outputting both by default would be to have a post-processing function to convert the exported images for human viewing.

lcadalzo:
This is my understanding of the current situation:

- if the export code [sees 6 channels,](https://github.com/twosixlabs/armory/blob/develop/armory/instrument/export.py#L67) it will split the input into two and export an RGB and depth image
- the export code currently makes no attempt to convert between the 2 kinds of depth shown [here](https://carla.readthedocs.io/en/latest/ref_sensors/#depth-camera): rgb-depth (_Original_, in the linked image) and grayscale depth (_Converted_, in the linked image)
- the CARLA OD dataset used in eval 5 has the depth data in grayscale already, so when we exported depth data then it looked like the image on the right
- the CARLA overhead OD dataset used in eval 6 instead has depth data in the rgb-depth form (like the image on the left)

And potential issues with this:
- As Jonathan said, if you're _only_ using depth channels, there'll be just 3 channels and a depth image will be exported without `depth` anywhere in the filename (since the export code doesn't know it's a depth image)
- the mismatch between the eval 5 datasets's depth format and that of the overhead dataset in eval 6 (the former is grayscale, the latter rgb-depth). How do we want to handle this?

@jprokos26 does this all align with your assessment of the situation? Am I misinterpreting anything?

davidslater:
@swsuggs this one is open, if you want to take it

PR #1737, opened 2022-11-02
Eval6 export tweaks

swsuggs:
For #1666.

Some depth images were not being exported with "depth" in their name.  This should fix that.

It also converts rgb-depth images to grayscale and exports both (no options to do one or the other as of yet).  There are some assumptions made in this process, but I think they should hold.

Also I noticed that depth images with boxes are not exported in the multimodal case (they are for depth-only models).  I've enabled that.  So now multimodal OD will export 12 images per sample: `boxes/no boxes` X `adv/benign` X `rgb/rgb-depth/gray-depth`.


swsuggs:
Still testing but @davidslater @lcadalzo take a look if you want

lcadalzo:
I agree there may not be a super clean way to handle this. That being said, l do think this PR would add a lot of complexity to the export process, and I'm wondering if it's worth the benefit.

To recap, there are two main issues:

1.  if you're using only depth channels (what I would assume to be a rare case), the outputted filename won't include "depth" in it. I would argue this isn't particularly a big problem, since if the user is only using depth channels, they know/expect the output to be depth. The presence of `depth` in the filename I think is most beneficial for distinguishing between the depth and rgb, but if there's no rgb, this distinction doesn't need to be made
2. different datasets have different depth formats (log-grayscale vs linear-rgb). While this inconsistency may be inconvenient, I actually think it may be more transparent to the user to only output whichever format the dataset is in. I can imagine a user being confused at seeing both outputs and not knowing which the model is actually seeing vs. which is just postprocessing for exporting purposes. Since there are users who may want access to both types of depth image, I think we should provide convenient conversion functions between the two formats, and we should also call out the depth format in our dataset documentation

For these reasons, and again primarily because of the complexity it adds to the export + scenario code, I suggest we simply add documentation highlighting the differences in depth format and pointing to conversion functions (which I think already exist?). If we feel strongly about keeping the changes here, one comment I'd have is to not output separate images with boxes for rgb, `depth`, _and_ `depth_grayscale`, as the number of output images per sample is getting a bit unwieldy:
```
I have no name!@cf366e417daf:/workspace$ ls /armory/outputs/2022-11-03T180355.846554/saved_samples_1667509112.1430664/
batch_0_ex_0_scenario.x.png                  batch_0_ex_0_scenario.x_depth_with_boxes.png            benign_predicted_boxes_coco_format.json
batch_0_ex_0_scenario.x_depth.png            batch_0_ex_0_scenario.x_depth_with_boxes_grayscale.png  ground_truth_boxes_coco_format.json
batch_0_ex_0_scenario.x_depth_grayscale.png  batch_0_ex_0_scenario.x_with_boxes.png                  predictions.pkl
```

swsuggs:
@lcadalzo I very much agree with both your points and I'm glad you brought them up.  The complexity this adds outweighs the benefit in my mind as well.  I prefer the idea of standalone conversion functions, but there are two further points to discuss/be aware of:

1. The existing functions are not sufficient to take an exported image, convert it, and save the result (even assuming we add `rgb_to_linear_depth()`).  How much of this additional process do we want to provide?  Would it be nice for the function to take an output directory and convert all depth images under it?  In that case, it would be useful for all depth images to have "depth" in their name.  It's simpler if we only provide a function that takes one png file and exports another, but this potentially leaves more work to the user.
2. It won't really work on images with boxes at all.  Drawing the box on the rgb-depth image destroys the actual depth information at those pixels, and the red and white boxes will be converted to something in the grayscale space but they certainly won't be red and white anymore.  I think it's fine as long as the docs are clear about why the tool is only suited for images with no boxes.  (It's possible to re-draw the boxes from the coco-format export data, but I'd prefer to wait and see if anyone really needs that.)

lcadalzo:
I think a simple function that converts either a single PIL image or numpy array is sufficient. I think it makes sense to keep things simple/clean in Armory and a user importing this function into their Python script can use it in whatever manner best for them

lcadalzo:
@swsuggs do you mind updating the docs to mention that users can convert between depth formats? We should also, beforehand, state which depth format each of the datasets uses. I think [this](https://github.com/swsuggs/armory/blob/eval6-export-tweaks/docs/adversarial_datasets.md#carla-object-detection) section is likely the best place

lcadalzo:
A few more comments:
1. could we add docstrings to [these](https://github.com/twosixlabs/armory/blob/3a8535c4b784ca6e03d828cc55c7c0d7bd1f93a0/armory/art_experimental/attacks/carla_obj_det_utils.py#L46-L91) functions indicating what the expected shape, dtype, and range for the inputs is?

2. the docs seem to imply there are 2 kinds of format, but we have > 2 conversion functions. So in addition to the logorithmic grayscale format and rgb encoding, there's a 3rd format linear depth in meters? If I want to go from rgb encoding to log grayscale, I need to call `rgb_depth_to_linear()` and then `linear_to_log()`?

3. a user may wonder why `carla_obj_det_dev` and `carla_obj_det_test` have 3 depth channels if it's log grayscale not rgb (the answer being that the second and third channels are copies of the first). We could state this in the docs
4. would be useful to add a test that converts rgb to linear, then back to rgb, ensuring the input was properly inverted

lcadalzo:
Changes to the docs look good, only thing is i'm hitting an error when trying the test:

```
  File "/workspace/test_conversion.py", line 39, in <module>
    test_carla_depth_format_conversion_utility_functions()
  File "/workspace/test_conversion.py", line 28, in test_carla_depth_format_conversion_utility_functions
    assert np.allclose(r, r_)
AssertionError

```
Looks like one value is normalized and the other not:
```
(Pdb) p r.mean()
192.17999999999884
(Pdb) p r_.mean()
0.7536470588235249
(Pdb) p g.mean()
202.0
(Pdb) p g_.mean()
0.7921568627450981
(Pdb) 202/255
0.792156862745098

```

lcadalzo:
Looks like I just needed to pull, sorry about that. LGTM

PR #1741, opened 2022-11-03
Add live demo to example Notebook and README

christopherwoodall:
- [x] Update notebook scenario path
- [x] Add scaffolding for Google Colab

Links resolve to `master`, but the branch can be changed in the URL. Below is an example URL pointing to the `rc0.16.1` branch:
```
https://colab.research.google.com/github/twosixlabs/armory/blob/rc0.16.1/notebooks/running_armory_scenarios_interactively.ipynb
```

Demo available here: 
[https://colab.research.google.com/drive/1cOb7xKnYrxsAs13w5J32BJFWEQ1nHX4r?usp=sharing](https://colab.research.google.com/drive/1cOb7xKnYrxsAs13w5J32BJFWEQ1nHX4r?usp=sharing)


README.md
```diff
@@ -33,7 +38,12 @@ evaluation outputs.
 
 To change these default directories simply run `armory configure` after installation.
 
-If installing from the git repo in editable mode, ensure that your pip version is 22+. 
+If installing from the git repo in editable mode, ensure that your pip version is 22+.
+
+## `docker`
```
lcadalzo, line 43: this docker section seems like it could either be deleted or fleshed out in more detail

lcadalzo:
if the main purpose of this PR is to demonstrate Armory use with Google Colab, can we simply tie that into the existing notebook rather than creating a new notebook that is 95% the same?

christopherwoodall:
> the Colab link is pointing to the version of the ipynb that's currently on master, not the version on this branch

This link will not work until these changes are merged into main.

> r.e. "Setup and Installation" , "Google Colab", [comment](https://github.com/twosixlabs/armory/pull/1741#issuecomment-1305879498)

See discourse in Slack

lcadalzo:
Is this in a place where I should be able to test things by actually running it in Colab? Does one need to first execute the steps you had added previously for getting Armory to run in Colab?

lcadalzo:
I can run the Colab cells without error. When I launch a scenario from my local machine with `--jupyter`, though, I run into the following error when running the cell below `"Loading the Scenario"`:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In [3], line 12
     10   print("Armory is running locally inside a container")
     11   # Search for the scenario config inside the `/armory/tmp` directory
---> 12   [scenario_config] = [f for f in sorted(Path("/tmp").iterdir()) if f.name == "interactive-config.json"]
     13 else:
     14   print("Armory is running remotely")

ValueError: not enough values to unpack (expected 1, got 0)
```

Note that I did not run the initial cell below:
```
if [ ! -d armory-repo ]; then
  git clone https://github.com/twosixlabs/armory.git armory-repo
fi

pip install --upgrade pip
pip install armory-testbed[developer,engine,datasets,pytorch,tensorflow]
```

christopherwoodall:
I had an error in the code cell below **Loading the Scenario**; searching `/tmp` instead of `/armory/tmp`. Should be fixed now.

image

mwartell:
Some notes on pedagogy.
1. there are too many choice points before the first code cell. The first time user doesn't know things that you do.
2. minimize the pip extras `pip install armory-testbed[pytorch]` entails engine, datasets, math, and developer is unneeded.
3. be consistent with naming; the phrase "Only execute this cell if you are running this notebook in a remote jupyter environment." is hard to answer if you don't know that colab is jupyter based.
4. Remove the choice at all when possible. Warning that the `pip install` takes a few minutes should be enough. If there is no harm to installing (aside from time) then don't ask the question.
5. The code in Loading the Scenario is crazy complicated.
6. `if ("remote", "local")[Path("/armory-repo").exists()] == "local"` is equivalent to `if Path("/armory-repo").exists():`
7.  `[scenario_config] = [f for f in sorted(Path("/armory/tmp").iterdir()) if f.name == "interactive-config.json"]` can be reduced to `config = Path("/armory/tmp").glob("**/interactive-config.json") [0]`
8. Better still would be to remove the complex conditional. Does the new `armory.SRC_ROOT` path help?
9. `from armory.scenarios.main import get as get_scenario` is needless cognitive load; simpler might be `import armory.scenarios.main` and then the call as `armory.scenarios.main.get()` if you think `get` is poorly named.
10. Is getting the absolute path necessary? If not that's one line I don't have to understand.

Pity the reader, they have to come into this knowing nothing, and every complicated thing detracts from the intention of a low-barrier introduction. Once past setup, the

lcadalzo:
A couple things about this comment:

```
# When launched interactively, or from the container, the scenario config will
# be located in a time-stamped sub-directory of `armory/tmp/`.
```

1. I'm not quite sure what's meant by "from the container" here. And perhaps, to be super explicit, we should say `"when launched with --interactive"`. And rather than `located` should we say `copied to`?
2. Do you mean `/armory/tmp`?

lcadalzo:
R.e.
```
If running armory locally, you can follow along with this tutorial by running the following command: armory run scenario_configs/eval5/carla_object_detection/carla_obj_det_dpatch_undefended.json --jupyter
```

that's one way, but another would be to use `--interactive` instead of `--jupyter`. Could it be helpful to be explicit and enumerate the 3 potential cases here (local with --interactive, local with --jupyter, or remote with Colab) and what a user needs to do in each case?

Also, since this may be someone's first glimpse at Armory (especially considering the addition of Colab use, which we may want to point new folks at), it may be helpful to add a 1 or 2 sentence preamble on how Armory _normally_ is run, i.e. when we say "interactive" what do we mean by non-interactive

PR #1742, opened 2022-11-04
Build datasets full pipeline

davidslater:


armory/datasets/README.md
```diff
@@ -0,0 +1,115 @@
...
+### TFDS standard dataset
+
+MNIST is a standard built-in TFDS dataset. To build this, do:
+```
+python -m armory.datasets.build mnist
+```
+
+To load it, you can do:
+```
+python -m armory.datasets.load mnist
```
lcadalzo, line 40: what's the intended use case in running `armory.datasets.load` as `__main__()`, simply to test whether it can load? I ask because running that script will load the dataset and do nothing. The interpreter example below seems more likely to resemble the use case I'd think, but wanted to check if I was missing something

armory/datasets/README.md
```diff
@@ -0,0 +1,115 @@
+# Setup
+
+This is designed to be run in an armory pytorch container.
+Recommended use:
+```
+armory launch pytorch
```
christopherwoodall, line 6: Should we add `build` and `load` as entry points in `pyproject.toml`? This way you can run this without installing armory; e.g. something like `docker run --rm -it --command armory-datasets build -v mount_volumes:if_needed twosixlabs/pytorch:latest`.

armory/datasets/package.py
```diff
@@ -0,0 +1,124 @@
...
+    file_size, file_sha256 = common.hash_file(filepath)
+
+    common.update_cached_datasets(name, version, subdir, file_size, file_sha256, url)
+
+
+def verify(name, data_dir: str = None):
+    info = common.cached_datasets()[name]
+    version = info["version"]
+
+    filepath = common.get_cache_dataset_path(name, version)
```
lcadalzo, line 69: should the `data_dir` kwarg be passed along to the `get_cache_dataset_path()` call?

armory/datasets/README.md
davidslater: I was thinking mainly for testing builds in a scripted fashion. I think that if you load a number of datasets sequentially in the same python instantiation, you may hit memory issues.

armory/datasets/download.py
```diff
@@ -0,0 +1,76 @@
+from pathlib import Path
+
+import boto3
+from botocore import UNSIGNED
+from botocore.client import Config
+from botocore.exceptions import ClientError
+
```
christopherwoodall, line 7: These should be added in `pyproject.toml` under `datasets`.

christopherwoodall: Both `boto3` and `botocore`.

armory/datasets/README.md
davidslater: Possibly, but let's hold off for now. We will want to think through most of the use cases beforehand. For most users, I don't think it will be needed, as loading a cached dataset within a scenario should work perfectly fine without its own entry point.

armory/datasets/package.py
davidslater: Yes, it should.

Also, I'm not the biggest fan of how I pass around `data_dir`, but not sure how we want to structure it.

armory/datasets/download.py
davidslater: Should we have them in both datasets and engine?

armory/datasets/README.md
```diff
@@ -0,0 +1,115 @@
...
+info, ds = load.load("digit")
+# Alternatively:
+info, ds = load.from_directory("/armory/datasets/new_builds/digit/1.0.8")
+```
+
+## Packaging and Uploading for Cache
+
+After a dataset has been successfully built and loaded (locally), it can be packaged and uploaded to the cache.
+
+First, it is recommended that you test the packaging and untarring process without upload/download.
```
lcadalzo, line 80: the order of operations here (create tar, update the json, etc) makes sense, but one minor comment I have is that I think the method names could be more descriptive of what they are doing

Or maybe what I'm arguing for is a separate method that wraps these, something like `cache()` or `add_to_cache()` that then downloads the tar _and_ updates the json _and_ verifies the file. This could be a more user-friendly way to complete these steps rather than one-by-one

armory/datasets/package.py
```diff
@@ -0,0 +1,124 @@
...
+) -> str:
+    """
+    Package a built dataset as .tar.gz, return path
+    """
+    version, data_dir, built_data_dir, subdir = build.build_info(
+        name, version=version, data_dir=data_dir
+    )
+
+    data_dir = Path(data_dir)
+    expected_dir = data_dir / name / version
```
lcadalzo, line 27: equivalently (I think):
```
expected_dir = data_dir / subdir
```

armory/datasets/package.py
```diff
@@ -0,0 +1,124 @@
...
+
+from armory.logs import log
+from armory.datasets import build, common
+
+
+def package(
+    name,
+    version: str = None,
+    data_dir: str = None,
+    cache_subdir=common.CACHE_SUBDIR,
```
lcadalzo, line 16: `cache_subdir` kwarg isn't being used

lcadalzo: i think the information it contains is captured in the `common.get_cache_dataset_path()` call

davidslater: Good catch, I refactored that into common

armory/datasets/README.md
davidslater: Added

armory/datasets/download.py
christopherwoodall: No issue there. Let `pip` take the wheel.

armory/datasets/package.py
davidslater: correct. I'm not sure if we actually need to return subdir in `build_info`. May be better to factor out, but I'm not sure - there may be datasets with weird pathing.

davidslater:
I added "carla_over_obj_det_dev" as an adversarial dataset.

armory/datasets/README.md
christopherwoodall: Makes sense, just note we can have multiple entrypoints to different modules; `armory`, `armory-engine`, `armory-datasets`, etc.

armory/datasets/common.py
```diff
@@ -0,0 +1,177 @@
...
+import json
+import hashlib
+from pathlib import Path
+
+import tensorflow_datasets as tfds
+
+from armory import paths
+from armory.logs import log
+
+SUBDIR = "new_builds"
```
lcadalzo, line 11: is the use of the `/armory/datasets/new_builds` subdir temporary and intended to separate out data that's been built in the past from regular armory use? If no, this might be an unexpected location to be storing future builds in the long-term

armory/datasets/download.py
davidslater: Done.

armory/datasets/common.py
davidslater: Yes, correct.

lcadalzo:
> I added "carla_over_obj_det_dev" as an adversarial dataset. I would be curious if a simple `load.load("carla_over_obj_det_dev")` works for you.

This did work for me. I thought it would fail since I hadn't built the dataset yet, but I see now that since `name in common.cached_datasets()` [here](https://github.com/davidslater/armory/blob/build-datasets-full-pipeline/armory/datasets/load.py#L72-L80), it will download it for me by default

the only reason this felt counter-intuitive to me is because, before checking, my assumption was that `cached_datasets()` was the set of datasets I had cached locally on my machine (which did not apply to `carla_over_obj_det_dev`). In reality, [cached_datasets](https://github.com/davidslater/armory/blob/build-datasets-full-pipeline/armory/datasets/cached_datasets.json) feels more like "officially supported Armory datasets", since it's the set of datasets included in the json file

davidslater:
I had a hard time with the name, because with this setup, we should be able to "support" datasets that we don't store in s3. So I wasn't sure what the right name was.

lcadalzo:
> I had a hard time with the name, because with this setup, we should be able to "support" datasets that we don't store in s3. So I wasn't sure what the right name was.

yea that makes sense. Could something like `armory_public_datasets()` or `armory_public_s3_datasets()` make sense?

armory/datasets/download.py
```diff
@@ -0,0 +1,76 @@
...
+        total = client.head_object(Bucket=bucket, Key=key)["ContentLength"]
+        with progress.ProgressPercentage(client, bucket, key, total) as Callback:
+            client.download_file(bucket, key, local_path, Callback=Callback)
+    except ClientError:
+        raise KeyError(f"File {key} not available in {bucket} bucket.")
+
+
+def download(
+    name,
+    data_dir: str = None,
```
lcadalzo, line 43: `data_dir` kwarg isn't being used in this function, although the `common.get_cache_dataset_path()` call below is by default using `get_root()`

davidslater: I should have passed that into `common.get_cache_dataset_path()`. Fixed.

davidslater:
Thoughts on merging this into the `tfdsv4` branch and iterating from there (on things like renaming)?

lcadalzo:
sounds good

PR #1746, opened 2022-11-08
Build datasets full pipeline generator

davidslater:


davidslater:
@lcadalzo made another set of changes / additions. If you look at the updates I made to the README, you can now patch in the new datasets into the current scenario loading. This should make it easier to test end-to-end consistency before we switch over. (It also shows how to wrap the armory data generator in an ART one)

davidslater:
For cleanliness, I moved the preprocessing to its own file.

For a new dataset (or for a custom one), a user would need to just register the preprocessing function with their dataset's name.

armory/datasets/config_load.py
```diff
@@ -0,0 +1,85 @@
...
+    else:
+        raise NotImplementedError
+
+    name = dataset_config.pop("name")
+    if ":" in name:
+        name, version = name.split(":")
+    else:
+        version = None
+
+    for k in "eval_split", "train_split":
```
lcadalzo, line 58: I'd need to look more closely at exactly how we're handling `"eval_split"` and `"train_split"` but this for loop seems a bit messy. I understand, though, that these keys need to be removed before `kwargs` are passed to `actual_loader()`

armory/datasets/config_load.py
```diff
@@ -0,0 +1,85 @@
...
+
+def actual_loader(
+    name,
+    version=None,
+    split="test",
+    shuffle_files=False,
+    framework="numpy",
+    epochs=1,
+    drop_remainder=False,
+    num_batches=None,
```
lcadalzo, line 18: `num_batches` should be passed along to the generator instantiation below

armory/datasets/config_load.py
```diff
@@ -0,0 +1,85 @@
...
+
+    for k in "eval_split", "train_split":
+        kwargs.pop(k, None)
+
+    kwargs.update(dataset_config)
+    if check_run:
+        kwargs["epochs"] = 1
+        kwargs["num_batches"] = 1
+    if kwargs.get("shuffle_files", False):
+        kwargs["shuffle_elements"] = True
```
lcadalzo, line 66: why is this `if` block necessary? If we changed the kwarg name in `actual_loader()` from `shuffle_elements` to `shuffle_files` would it remove the need for this?

armory/datasets/config_load.py
```diff
@@ -0,0 +1,85 @@
+"""
+Temporary file for testing loading from config without modifying armory.utils
+"""
+
+import copy
+
+from armory.datasets import load, preprocessing, generator
+
+
+def actual_loader(
```
lcadalzo, line 10: could the `ArmoryDataGenerator` object instead be instantiated inside the `load_dataset()` function, obviating the need for this extra function? I think this makes things a bit simpler if so

armory/datasets/preprocessing.py
```diff
@@ -0,0 +1,101 @@
...
+def list_registered():
+    return list(REGISTERED_PREPROCESSORS)
+
+
+def get(name, version=None):
+    if version is not None:
+        raise NotImplementedError("version is not None for preprocessing to_canon")
+    if name not in REGISTERED_PREPROCESSORS:
+        raise KeyError(
+            f"prepreprocessor {name} not registered. Use one of {list_registered()}"
```
lcadalzo, line 29: based on how things are currently set up, `name` will always simply be the name of the dataset. So let's say you're using a new (to Armory), say, image classification dataset e.g. `fashion_mnist`. I don't see an easy way to use your new dataset with an _existing_ preprocessing function, e.g. mnist's. A lot of these preprocessing functions will look the same (`(image_to_canon(element["image"]), element["label"])`) for different datasets, so it'd be a bit cumbersome to need to implement a new function:
```
def fashion_mnist(element):
    return (image_to_canon(element["image"]), element["label"])
```
that's a replica of existing ones.

armory/datasets/preprocessing.py
```diff
@@ -0,0 +1,101 @@
...
+    return function
+
+
+def list_registered():
+    return list(REGISTERED_PREPROCESSORS)
+
+
+def get(name, version=None):
+    if version is not None:
+        raise NotImplementedError("version is not None for preprocessing to_canon")
```
lcadalzo, line 26: why does `actual_loader()` pass along a `version` kwarg to `preprocessing.get()` if the latter shouldn't receive anything other than `None` for this? Is the idea that eventually `REGISTERED_PREPROCESSORS` will have keys for different versions of the same dataset?

armory/datasets/generator.py
```diff
@@ -0,0 +1,135 @@
...
+        Note: size computations will be wrong when filtering is applied
+    element_map - function that takes a dataset element (dict) and maps to new element
+    """
+
+    FRAMEWORKS = ("tf", "numpy", "torch")
+
+    def __init__(
+        self,
+        info,
+        ds_dict: dict,
```
lcadalzo, line 41: could be helpful to spell out in a docstring what they keys in `ds_dict` should map to, or how one constructs said dict

armory/datasets/config_load.py
```diff
@@ -0,0 +1,85 @@
...
+    name,
+    version=None,
+    split="test",
+    shuffle_files=False,
+    framework="numpy",
+    epochs=1,
+    drop_remainder=False,
```
lcadalzo, line 17: `drop_remainder` is also not being used

armory/datasets/config_load.py
```diff
@@ -0,0 +1,85 @@
...
+        batch_size=batch_size,
+        framework=framework,
+        epochs=epochs,
+        element_filter=None,
+        element_map=preprocessor,
+        shuffle_elements=shuffle_elements,
+    )
+
+
+def load_dataset(dataset_config, *args, check_run=False, **kwargs):
```
lcadalzo, line 37: what `args` should `load_dataset()` be receiving beyond `dataset_config`?

armory/datasets/config_load.py
```diff
@@ -0,0 +1,85 @@
...
+    NOTE: very ugly
+    """
+    dataset_config = copy.deepcopy(dataset_config)
+    module = dataset_config.pop("module")
+    if module == "armory.data.datasets":
+        pass
+    elif module == "armory.data.adversarial_datasets":
+        pass  # TODO: check
+    else:
+        raise NotImplementedError
```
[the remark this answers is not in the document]
davidslater: This was a temporary item. I expect the configs to change once we get the the new dataset material in there.

armory/datasets/config_load.py
davidslater: It's really weird. But it comes from scenarios passing in things. I don't think args is currently used, but this is the function signature for the existing `armory.utils.config_loading.load_dataset` method, so I duplicated that here.

armory/datasets/config_load.py
```diff
@@ -0,0 +1,85 @@
...
+    info, ds_dict = load.load(name, version=version, shuffle_files=shuffle_files)
+    preprocessor = preprocessing.get(name, version=version)
+    return generator.ArmoryDataGenerator(
+        info,
+        ds_dict,
+        split=split,
+        batch_size=batch_size,
+        framework=framework,
+        epochs=epochs,
+        element_filter=None,
```
lcadalzo, line 31: is setting `element_filter` to `None` temporary? One thing that appears to be still todo is incorporating `filter_by_class` and `filter_by_index`

armory/datasets/config_load.py
davidslater: Yes, exactly. The configs can look like this:
```
    "dataset": {
        "batch_size": 1,
        "eval_split": "test_clean",
        "framework": "numpy",
        "module": "armory.data.datasets",
        "name": "librispeech",
        "train_split": "train_clean100"
    },
```

And scenarios map those into the split needed at different points in time.

In contrast, I think we'll probably want to update the config structure so that the training dataset is a completely separate block, which should simplify some of that logic.

armory/datasets/config_load.py
davidslater: I wanted to separate them, because I think this function is much closer to what scenario configs should look like, not the current function. (But yes, having both is unneeded, ultimately.)

lcadalzo:
When running the mnist scenario, the results are succesfully logged and output JSON saved, but there is an ignored exception that gets logged at the very end:

```
2022-11-09 16:29:36 34s SUCCESS  armory.instrument.config:_write:224 benign_mean_categorical_accuracy on benign examples w.r.t. ground truth labels: 97.60%
2022-11-09 16:29:36 34s SUCCESS  armory.instrument.config:_write:224 adversarial_mean_categorical_accuracy on adversarial examples w.r.t. ground truth labels: 31.39%
2022-11-09 16:29:36 34s SUCCESS  armory.instrument.config:_write:224 benign_mean_categorical_accuracy on benign examples w.r.t. ground truth labels: 97.60%
2022-11-09 16:29:36 34s SUCCESS  armory.instrument.config:_write:224 adversarial_mean_categorical_accuracy on adversarial examples w.r.t. ground truth labels: 31.39%
2022-11-09 16:29:36 34s INFO     armory.scenarios.scenario:save:436 Saving evaluation results to path /armory/outputs/2022-11-09T141346.407543/ImageClassificationTask_1668011345.json inside container.
Exception ignored in: <generator object _graph_dataset_iterator at 0x7f608457ce40>
  File "/opt/conda/lib/python3.9/site-packages/tensorflow_datasets/core/dataset_utils.py", line 80, in _graph_dataset_iterator
AttributeError: 'NoneType' object has no attribute 'errors'
Exception ignored in: <generator object _graph_dataset_iterator at 0x7f606c5f1e40>
  File "/opt/conda/lib/python3.9/site-packages/tensorflow_datasets/core/dataset_utils.py", line 80, in _graph_dataset_iterator
AttributeError: 'NoneType' object has no attribute 'errors'
I have no name!@452ee2cecf43:/workspace$ 
```

the tf function this is referring to as follows:
```
def _graph_dataset_iterator(ds_iter, graph: tf.Graph) -> Iterator[NumpyElem]:
  """Constructs a Python generator from a tf.data.Iterator."""
  with graph.as_default():
    init = ds_iter.initializer
    ds_item = ds_iter.get_next()
    with utils.nogpu_session() as sess:
      sess.run(tf.compat.v1.tables_initializer())
      sess.run(init)
      while True:
        try:
          yield sess.run(ds_item)
        except tf.errors.OutOfRangeError:
          break

```

and the AttributeError is occurring at `except tf.errors.OutOfRangeError:`

davidslater:
It's probably some race condition when tearing down python. I've seen lots of similar errors in the past before involving NoneType in TF.

armory/datasets/generator.py
```diff
@@ -0,0 +1,135 @@
...
+                f"num_batches must be None or a positive integer, not {num_batches}"
+            )
+        if framework not in self.FRAMEWORKS:
+            raise ValueError(f"framework {framework} not in {self.FRAMEWORKS}")
+
+        size = info.splits[split].num_examples
+        batch_size = int(batch_size)
+        batches_per_epoch = size // batch_size
+        if not drop_remainder:
+            batches_per_epoch += bool(size % batch_size)
```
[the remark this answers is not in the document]
davidslater: The cast to a bool should only increment it by 1.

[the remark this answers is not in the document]
lcadalzo: somehow my eyes missed the bool, but yep makes sense

armory/datasets/config_load.py
```diff
@@ -0,0 +1,85 @@
...
+    name = dataset_config.pop("name")
+    if ":" in name:
+        name, version = name.split(":")
+    else:
+        version = None
+
+    for k in "eval_split", "train_split":
+        kwargs.pop(k, None)
+
+    kwargs.update(dataset_config)
```
lcadalzo, line 61: the 
```
kwargs.update(dataset_config)
```

I believe should occur before the `for k in "eval_split", "train_split":`, since the former can add keys such as `'eval_split"` to `kwargs` which is then subsequently erroneously passed to `actual_loader()`

armory/datasets/preprocessing.py
davidslater: That's a good call. I made an easy way to reference them.

armory/datasets/preprocessing.py
davidslater: Removed version from that.

armory/datasets/generator.py
davidslater: Added in docstring.

armory/datasets/config_load.py
davidslater: Made the whole thing a bit more readable.

davidslater:
@lcadalzo thanks again for the review. i'm working on the filtering operations to maintain parity with current codebase, and then will reach out again for another review.

davidslater:
Just tested with fashion mnist. You would need to do:
```
from armory.datasets import config_load, build
build.build('fashion_mnist')
gen = config_load.actual_loader('fashion_mnist', preprocessor_name='mnist')
x, y = next(gen)
x.max(), x.min()  # (1.0, 0.0)
```

davidslater:
Now I have some working filtering operations:
```
gen = config_load.actual_loader("fashion_mnist", preprocessor_name="mnist", class_ids=[3, 4], batch_size=100)
x, y = next(gen)
y
# array([4, 4, 4, 3, 4, 4, 4, 4, 4, 4, 4, 3, 3, 3, 4, 3, 4, 4, 4, 4, 4, 4,
       4, 4, 4, 3, 4, 3, 4, 4, 3, 3, 3, 4, 4, 3, 3, 3, 3, 4, 4, 3, 3, 3,
       3, 3, 4, 4, 4, 3, 3, 4, 3, 4, 3, 3, 3, 4, 4, 3, 3, 4, 4, 4, 3, 4,
       4, 3, 3, 4, 3, 4, 3, 3, 3, 3, 4, 3, 4, 4, 3, 3, 3, 4, 4, 3, 4, 4,
       4, 3, 3, 4, 3, 3, 3, 3, 4, 3, 3, 3]
```
Similarly,
```
gen = config_load.actual_loader("fashion_mnist", preprocessor_name="mnist", index="[4:45]", batch_size=100)
y
# array([5, 1, 0, 5, 7, 4, 0, 8, 2, 3, 9, 0, 7, 7, 2, 2, 0, 4, 4, 4, 2, 7,
       7, 4, 2, 4, 7, 5, 9, 5, 4, 4, 3, 3, 1, 7, 5, 3, 0, 0, 0])
```
and
```
gen = config_load.actual_loader("fashion_mnist", preprocessor_name="mnist", index=[0, 2, 5], batch_size=100)
y
# array([4, 9, 1])
```

armory/datasets/generator.py
```diff
@@ -0,0 +1,152 @@
...
+        if epochs > 1:
+            ds = ds.repeat(epochs)
+        if shuffle_elements:
+            # https://www.tensorflow.org/datasets/performances#caching_the_dataset
+            # for true random, set buffer_size to dataset size
+            #     for large datasets, this is too large, therefore use shard size
+            # set shuffle buffer to the number of elements in a shard
+            # shuffle_files will shuffle between shard, this will shuffle within shards
+            #     this won't be true random, but will be sufficiently close
+            examples_per_split = math.ceil(size / info.splits[split].num_shards)
```
lcadalzo, line 108: just noticed that `math` isn't being imported

lcadalzo: went ahead and added a commit to address

lcadalzo:
Looks like the error above is happening because `actual_loader()` is receiving `None` for the `preprocessor_name` kwarg. See the strange behavior below as a result:

```
(Pdb) type(x)
<class 'str'>
(Pdb) p x
'image'
(Pdb) p y
'label'
```

davidslater:
So the reason this happens is that the preprocessor is currently doing two things:
1) Performing data and label preprocessing
2) Mapping from a dict to a tuple

If it is None, the generator returns a dict (call it `z`), and so when you do `x, y = z`, it actually is taking the keys from the dict (and would have failed if there were more keys). Allowing the generator to return dicts could be helpful in many settings, I think, especially in places were we have `x`, `x_adv`, `y`, `y_patch_metadata`, etc. I would enable us to not have to wrap them into the tuple structure that we then have to unpack `((x, x_adv), (y, y_patch_metadata))`.

However, maybe I should focus on separating the preprocessing from the `dict -> tuple` mapping?

lcadalzo:
Got it. I think
```
x, y = next(dataset)

# in some cases:
y, y_patch_metadata = y
```

is preferable over something like

```
x_key = "image"
y_key = "label"
dict = next(dataset)
x = dict[x_key]
y = dict[y_key]
```

and think I agree that separating preprocessing (in the sense we traditionally think of it, e.g. normalizing) from mapping `dict` to `tuple` makes sense

davidslater:
I was thinking of something like mapping to some standard dict form `{"x": ..., "y": ..., }` instead of the original keys, so it's easy to check for fields of interest and you could even do something like:
```
dataset.as_tuple("x", "y", "y_patch_metadata")
x, y, y_patch_metadata = next(dataset)
```

lcadalzo:
perhaps it would be good to have a default mapping between datasets and preprocessing functions? I can get rid of the error by setting `"preprocessor_name"` in the `"dataset"` portion of the config, but I don't think we'd want to force users to always set this field

davidslater:
What about something like this:
```
def actual_loader(                                                                       
    name=None,                                                                           
    version=None,                                                                        
    shuffle_files=False,                                                                 
    preprocessor_name="DEFAULT", 
    ...
):
    ...
    if preprocessor_name is None:                                                        
        preprocessor = None                                                              
    elif preprocessor_name == "DEFAULT":                                                 
        # TODO: do something smart here                                                  
        raise NotImplementedError                                                        
    else:                                                                                
        preprocessor = preprocessing.get(preprocessor_name)
```

davidslater:
Something smart could do something like:
A) check for a preprocessor that matches the dataset name
B) use the info features to automatically detect what to use

lcadalzo:
Only downside to A is that I'd think many datasets could map to the same preprocessing fn (and also a developer would need to know/remember to name the fn the same thing as the dataset name). Would have to look more into B

What I was initially thinking was something like

```
def actual_loader(                                                                                                                               
    preprocessor_name="DEFAULT", 
    ...
):
    ...
    if preprocessor_name is None:                                                        
        preprocessor = None                                                              
    elif preprocessor_name == "DEFAULT":                                                 
        from armory.datasets.preprocessing import DATASET_TO_PREPROCESSOR
        preprocessor = DATASET_TO_PREPROCESSOR.get(name)                                                
    else:                                                                                
        preprocessor = preprocessing.get(preprocessor_name)
```

where `DATASET_TO_PREPROCESSOR` looks something like:
```
DATASET_TO_PREPROCESSOR= {"mnist": supervised_image_classification, ...}
```

lcadalzo:
> I was thinking of something like mapping to some standard dict form `{"x": ..., "y": ..., }` instead of the original keys, so it's easy to check for fields of interest and you could even do something like:
> 
> ```
> dataset.as_tuple("x", "y", "y_patch_metadata")
> x, y, y_patch_metadata = next(dataset)
> ```

I do like the benefit of being able to check for certain things (e.g. `"y_patch_metadata"`) rather than unexpectedly unraveling something of a different datatype/format than expected. Something feels nice and clean about doing `x, y = next(dataset)` but I suppose functionally it doesn't necessarily provide an advantage over returning dicts

davidslater:
`A` was mainly so that if we already had matching names, we would just smartly use those. However, `B` was where the main smarts would be. Having a `DATASET_TO_PREPROCESSOR` dict requires updating the dict to use a new dataset, which is not that different from:
```
mnist = register(supervised_image_classification, "mnist")
```

lcadalzo:
if you were using a new dataset, more likely you'd add the field to the config than update the dict in Armory, but I see your point. It's a downside to add another step each time a dataset is added. If `B` can be done in a way that feels relatively clean that seems like a good option

PR #1753, opened 2022-11-09
better global metrics from config

davidslater:
The `my_task` piece is for testing, and should be removed before merging.

You should be able to add `my_task` to the metrics list in a config and have it output in the final config.


davidslater:
@ppark-twosixtech ready for another review

davidslater:
Looks like I need to fix some more tests.

davidslater:
@ppark-twosixtech @mwartell this update should pass the tests, so ready for re-review

NOTE: I don't think we actually want to include this in the rc0.16.1 release, because there may be some regressions introduced that will take a while to find, so maybe once we're good with it, we wait until after release.

lcadalzo:
should we provide a pointer [here](https://github.com/davidslater/armory/blob/better-global-metrics/docs/configuration_files.md#specifying-kwargs-for-metric-functions) in `metrics.md` docs somewhere? Or even move that content to `metrics.md`. The example linked is a common user story, and a user who doesn't know how to get that functionality will probably look to `metrics.md` before `configuration_files.md`

davidslater:
Good plan. I forgot that was even a functionality.

lcadalzo:
I also think specifying `"kwargs"` in the config would feel more intuitive than `"task_kwargs"`

davidslater:
The functionality is only there for task metrics, I believe. We would need a separate one for perturbation kwargs, if you wanted to add those.

docs/metrics.md
```diff
@@ -192,6 +195,133 @@ This module mostly contains code to load explanatory models, generate activation
...
+```
+from armory import metrics
+def my_accuracy_metric(y_i, y_pred_i):
+    return y_i == np.argmax(y_pred_i)
+
+metrics.task.elementwise(my_accuracy_metric, "a_different_name")
+```
+
+Armory performs all built-in metric operations as batches, not as individual elements, so using the `elementwise` decorators will also produce a batchwise version of it that loops through the individual elements and provides a batchwise result.
+NOTE: when armory uses `get`, it will
```
lcadalzo, line 244: is there text missing here? `it will...`

davidslater:
@ppark-twosixtech can you look at whether this outputs CARLA MOT results properly?

armory/instrument/config.py
```diff
@@ -122,345 +195,112 @@ def results(self):
...
     record_final_only=True,
+    suffix="",
 ):
     """
-    Return list of meters generated for this specific task
+    Return meter generated for this specific task
     """
-    meters = []
     metric = metrics.get(name)
+    result_formatter = metrics.get_result_formatter(name)
```
ppark-twosixtech, line 234: Minor comment, not sure what best practices are here, but would suggest moving this closer to `return Meter` since `result_formatter` is only used there

armory/metrics/task.py
```diff
@@ -1901,3 +1956,47 @@ def __call__(self, y_list, y_pred_list):
...
+    return f"total={float(total):.3}, {num}/{denom}"
+
+
+@result_formatter("word_error_rate")
+def word_error_rate_formatter(result):
+    result = total_wer(result)
+    return total_wer_formatter(result)
+
+
+result_formatter("total_word_error_rate")(total_wer_formatter)
```
ppark-twosixtech, line 1976: Is there a reason to be able to refer to `total_wer_formatter` in `FORMATTERS` with two different keys?

armory/instrument/config.py
davidslater: I should have actually used in `GlobalMeter` too, fixing.

armory/metrics/task.py
davidslater: I meant it for backwards compatibility. But it's no longer needed, as it was only used in the old `instrument.config`

docs/metrics.md
davidslater: Fixed.

davidslater:
@lcadalzo @ppark-twosixtech I think I have addressed your comments, ready for re-review

ppark-twosixtech:
Was in the middle of testing and noticed that results weren't being shown on the log, only in the output file - is this intentional?

armory/instrument/instrument.py
```diff
@@ -709,13 +737,22 @@ def _write(self, name, batch, result):
 
 
 class LogWriter(Writer):
-    LOG_LEVELS = ("TRACE", "DEBUG", "INFO", "SUCCESS", "WARNING", "ERROR", "CRITICAL")
-
-    def __init__(self, log_level: str = "INFO"):
+    LOG_LEVELS = (
```
ppark-twosixtech, line 740: `PROGRESS` not included

ppark-twosixtech:
Not a critical issue per se, but setting `--log-level=TRACE` on `armory run` and setting `log-level: str ="TRACE"` for `ResultsLogWriter` does not show results in the log

ppark-twosixtech:
> 

resolved with 588c870

armory/instrument/instrument.py
ppark-twosixtech: Resolved with 588c870

ppark-twosixtech:
> Was in the middle of testing and noticed that results weren't being shown on the log, only in the output file - is this intentional?

Resolved with 588c870

ppark-twosixtech (approved):

ppark-twosixtech (approved):
LGTM

Issue #1755, opened 2022-11-10
Update scenario docs with eval6 baseline results


davidslater:

lcadalzo:
#1766 addressed audio poisoning, carla mot, and cifar dirty label. However, cifar sleeper agent (defenses) and gtsrb witches brew are still TODO

lcadalzo:
Still waiting on cifar sleeper agent results due to potential ART issue

PR #1766, opened 2022-11-18
Additional baseline results

swsuggs:
For #1755.

Remaining, in order of priority:
- [x]  audio poisoning
- [x]  carla mot (one result)
- [x]  cifar dirty label

Deferred till next release:
- [ ]  cifar sleeper agent (defenses)
- [ ]  gtsrb witches' brew


lcadalzo:
don't think it's related to this PR but while we're looking at these files, I notice a broken link in `carla_od_results.md` at the very bottom `Find reference baseline configurations [here]` (i.e. `here` appears in text and isn't a link)

docs/baseline_results/cifar10_witches_brew_results.md
```diff
@@ -1,3 +1,146 @@
 # Cifar10 Witches' Brew Baseline Evaluation
 
-Coming soon
\ No newline at end of file
+Results obtained using Armory 0.15.X (July 2022)
+
+Note: the baseline CIFAR resnet was updated in Armory 0.16.1.
```
lcadalzo, line 5: if the model was updated in 0.16.1 but the results are from 0.15, does this mean the results need to be re-computed?

swsuggs: Eventually I suppose that would be nice; it's certainly not urgent since witches' brew is not going to be a poisoning scenario we continue to use.  I think that labelling each result page with the armory version gives us some flexibility in when we update the results

lcadalzo:
LGTM

PR #1767, opened 2022-11-18
Oom error poisoning

### LINKED ISSUE #1761: OOM Error while running eval6 poisoning speech command audio_p10_undefended.json
While running the following test case : scenario_configs/eval6/poisoning/audio_dlbd/audio_p10_undefended.json, I run into an OOM issue.

Below is the exact error:
```
tensorflow.python.framework.errors_impl.ResourceExhaustedError: OOM when allocating tensor with shape[85511,16000] and type
 float on /job:localhost/replica:0/task:0/device:GPU:0 by allocator GPU_0_bfc [Op:Sub]
```

I am using a *NVIDIA 1080 Ti* with armory 0.16.0
Is the issue only with the fact that VRAM is not enough? I am able to run the same scenario on CPU clusters.

I tried to reduce the batch size from 64 all the way down to 2 and still gives the same error.

davidslater:
Fixes #1761

To test, you can set `fit_generator` to `true` in the `scenario` `kwargs` of the `audio_p01_undefended.json` config.

### DIFF
--- a/armory/data/datasets.py
+++ b/armory/data/datasets.py
@@ -258,7 +258,7 @@ class NumpyDataGenerator(DataGenerator):
         try:
             return next(self.generator)
         except StopIteration:
-            self.generator = self.single_epoch_generator()
+            self.generator = self._single_epoch_generator()
             return next(self.generator)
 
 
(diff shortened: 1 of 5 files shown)


davidslater:
Yeah, adding some documentation is a good idea.

davidslater:
I added this to the poisoning doc.

PR #1768, opened 2022-11-19
Activation bug resnet50

### LINKED ISSUE #1765: [BUG]: Poisoning activation defence gives error while extracting outputs of activation layers
### Description of the bug

This could be because the Resnet50 version of tensorflow from https://github.com/keras-team/keras/blob/b80dd12da9c0bc3f569eca3455e77762cf2ee8ef/keras/applications/resnet.py#L76 is not declared as tf.keras.models.sequential whereas the get_activations in https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/art/estimators/classification/tensorflow.py requires the model to be tf.keras.models.Sequential.

`ValueError: Method get_activations is not supported for non-Sequential models.`

Is there any quick fix?

### Steps To Reproduce

```
cd scenario_configs/eval6/poisoning/audio_dlbd/baseline_defenses
armory run audio_p10_activation_defense.json --no-docker --use-gpu
```

### Additional Information

_No response_

davidslater:
Fixes #1765

### DIFF
--- a/armory/baseline_models/tf_graph/audio_resnet50.py
+++ b/armory/baseline_models/tf_graph/audio_resnet50.py
@@ -33,6 +33,8 @@ def make_audio_resnet(**kwargs) -> tf.keras.Model:
     )
 
     model = keras.Model(resnet.inputs, resnet.outputs)
+    # ART's TensorFlowV2Classifier get_activations() requires a Sequential model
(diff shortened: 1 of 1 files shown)


davidslater:
An alternative solution would be to point the Activation Defense at a specific named layer of the resnet50.

Issue #1769, opened 2022-11-21
[BUG]: Error computing HOTA metrics in Carla MOT dpatch defended


swsuggs:
### Description of the bug

I get an error during the final metric calculations when running carla_mot_dpatch_defended.  I'm running on the rc0.16.1 branch.
```IndexError: arrays used as indices must be of integer (or boolean) type```

### Steps To Reproduce

armory run scenario_configs/eval6/carla_mot/carla_mot_dpatch_defended.json

### Additional Information

```
Traceback (most recent call last):
File "/opt/conda/lib/python3.8/runpy.py", line 194, in _run_module_as_main
return _run_code(code, main_globals, None,
│         │     └ {'__name__': '__main__', '__doc__': '\nMain script for running scenarios. Users will run a scenario by calling:\n    armory r...
│         └ <code object <module> at 0x7f5a91c72c90, file "/workspace/armory/scenarios/main.py", line 1>
└ <function _run_code at 0x7f5bd99b1d30>
File "/opt/conda/lib/python3.8/runpy.py", line 87, in _run_code
exec(code, run_globals)
│     └ {'__name__': '__main__', '__doc__': '\nMain script for running scenarios. Users will run a scenario by calling:\n    armory r...
└ <code object <module> at 0x7f5a91c72c90, file "/workspace/armory/scenarios/main.py", line 1>
File "/workspace/armory/scenarios/main.py", line 228, in <module>
run_config(
└ <function run_config at 0x7f5a90bfcee0>
File "/workspace/armory/scenarios/main.py", line 144, in run_config
scenario.evaluate()
│        └ <function Scenario.evaluate at 0x7f5a90849430>
└ <armory.scenarios.carla_mot.CarlaMOT object at 0x7f5a90bf3940>
> File "/workspace/armory/scenarios/scenario.py", line 402, in evaluate
self._evaluate()
│    └ <function Scenario._evaluate at 0x7f5a908493a0>
└ <armory.scenarios.carla_mot.CarlaMOT object at 0x7f5a90bf3940>
File "/workspace/armory/scenarios/scenario.py", line 393, in _evaluate
self.finalize_results()
│    └ <function Scenario.finalize_results at 0x7f5a90849310>
└ <armory.scenarios.carla_mot.CarlaMOT object at 0x7f5a90bf3940>
File "/workspace/armory/scenarios/scenario.py", line 381, in finalize_results
self.metric_results = self.metrics_logger.results()
[...]

PR #1772, opened 2022-11-22
Retain data type for arrays passed into "relable" function

jprokos26:
Addresses https://github.com/twosixlabs/armory/issues/1769
Relable function would return subarrays of type `float64` if they are empty which caused issues when passed into the HOTA function which expected it to be of type `int`. Change utilizes the empty array's dtype for consistency.


swsuggs (approved):
Thanks Jonathan, looks great

PR #1776, opened 2022-11-25
modify scenario load_dataset() and load_train_dataset()

lcadalzo:
NOTE: ignore the `v3` in the branch name, that's a typo

With the changes in this PR, the `load_dataset()` and `load_train_dataset()` methods are now always looking for `"test"` and "`train`" fields in the dataset config. To use an example where your test split is `test_clean`:

instead of having the config structured like:
```
"dataset": {
        "test_clean": {
            "batch_size": 64,
            "name": "my_dataset"
        }
    }

```

I think it seems more organized to have:

```
"dataset": {
        "test": {
            "batch_size": 64,
            "name": "my_dataset",
            "split": "test_clean"
        }
    }
```

This PR aligns the scenario code with the latter structure. This structure also aligns with the config schema modifications in #1774


lcadalzo:
Also added a commit that renames `eval_split_default` to `test_split_default`. In general, I think it only adds confusion to be using `eval` and `test` interchangeably throughout the codebase, so I'd like to get rid of `eval` where possible

davidslater:
`fit_batch_size` was an override we used for if the training loop needed a different batch size from test. That is unnecessary now that they are separately defined.

lcadalzo:
added the recent commit since I thought having `load_test_dataset()` and `load_train_dataset()` is more logical than `load_dataset()` and `load_train_dataset()`, especially given the direction we're going

davidslater (approved):
I like the design decisions

PR #1778, opened 2022-11-28
adding cifar preprocessing and updating config structure

lcadalzo:


lcadalzo:
One issue I've noticed come up here is that if the `"epochs"` you specify in the `"dataset"`-->`"train"` portion of the config is less than the epochs specified in `"model"`'s `"fit_kwargs"`, you get a `StopIteration` error bc the dataset fetched isn't long enough to do the training. I think it's confusing to have two very similar and interrelated params in different places, so I'd advocate for giving the `epochs"` specified in the `"dataset"`-->`"train"` sole control over the number of train epochs.

lcadalzo:
planning to address the above issue in a separate PR, so we should be able to merge this in quickly

davidslater (approved):

PR #1783, opened 2022-11-29
User init

davidslater:
Note: I targeted `master` for now, though this would go to `develop` once we merge `master` into `develop`.

### DIFF
--- a/armory/scenarios/scenario.py
+++ b/armory/scenarios/scenario.py
@@ -114,6 +114,8 @@ class Scenario:
                     kwargs_str = ""
                 log.info(f"Calling user_init function {module}.{name}({kwargs_str})")
                 target = getattr(mod, name, None)
+                if target is None:
+                    raise ValueError(f"user_init name {name} cannot be found")
                 if not callable(target):
                     raise ValueError(f"{module}.{name} is not callable")
                 target(**kwargs)
--- a/armory/scenarios/scenario.py
+++ b/armory/scenarios/scenario.py
@@ -3,6 +3,7 @@ Primary class for scenario
 """
 
 import copy
+import importlib
 import os
 import sys
 import time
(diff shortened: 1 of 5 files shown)


davidslater:
The loading functionality should probably go into `armory.utils.config_loading`, but that entire thing should probably be revamped.

davidslater:
@lcadalzo let me know if that works for you.

lcadalzo:
Haven't played around with it yet, but I like the re-wording of the docs

lcadalzo:
unrelated to this PR but can we get rid of [this](https://github.com/twosixlabs/armory/blob/388eec447d63959fa86def593201fe030da329cc/docs/scenarios.md#derivative-metrics) section, or is that still relevant?

davidslater:
Yes.

armory/scenarios/scenario.py
```diff
@@ -85,6 +86,40 @@ def __init__(
...
+            name = user_init.get("name")
+            kwargs = user_init.get("kwargs") or {}
+            if name:
+                if kwargs:
+                    kwargs_str = f"**{kwargs}"
+                else:
+                    kwargs_str = ""
+                log.info(f"Calling user_init function {module}.{name}({kwargs_str})")
+                target = getattr(mod, name, None)
+                if not callable(target):
```
lcadalzo, line 117: before we check if it's callable, should there be another check that it's not None?

If you specify a name that doesnt actually exist, you'll get 
```
ValueError <module.function_that_doesnt_exist> is not callable
```

rather than an error indicating the name can't even be found

docs/scenarios.md
```diff
@@ -19,6 +19,66 @@ Scenarios are implemented as subclasses of `Scenario`, and typically given their
...
+```python
+from armory.instrument import get_probe
+probe = get_probe("my_model")
+
+class MyModel(torch.nn.Module):
+    ...
+    def forward(x):
+        ...
+        table = np.array([[2, 3], [4, 6]])
+        probe.update(contingency_table=z)
```
lcadalzo, line 78: minor point but should this say

```
contingency_table=table
```

davidslater:
Should be ready for a re-review @lcadalzo

PR #1784, opened 2022-11-29
fix carla-mot build

davidslater:
This will pin carla-mot to explicitly use the pytorch container's version instead of "latest".

### DIFF
--- a/docker/build.py
+++ b/docker/build.py
@@ -82,6 +82,8 @@ def build_worker(framework, version, platform, base_tag, **kwargs):
     # Note: The replace is used to convert the version to a valid docker tag.
     version = version.replace("+", ".")
     dockerfile = script_dir / f"Dockerfile-{framework}"
+    if framework == "carla-mot":
+        base_tag = version
     build_command = [
         f"{platform}",
         "build",


davidslater:
Previously, if the current version pytorch container wasn't built yet but the previous version was and also tagged as "latest", carla-mot would build off of the previous version.

docker/build.py
```diff
@@ -82,6 +82,8 @@ def build_worker(framework, version, platform, base_tag, **kwargs):
     # Note: The replace is used to convert the version to a valid docker tag.
     version = version.replace("+", ".")
     dockerfile = script_dir / f"Dockerfile-{framework}"
+    if framework == "carla-mot":
+        base_tag = version
```
christopherwoodall, line 86: This will break the build and release workflows. We should explore making these changes in the Dockerfile. Possibly combining the two(`carla` + `pytorch`) and redirecting calls for `carla` in `armory/docker/images.py` to `pytorch`. This will also help to reduce the number of images that need to be maintained.

christopherwoodall (changes requested):

docker/build.py
christopherwoodall: An example approach:
  1. Move the following lines from `docker/Dockerfile-carla-mot` into `docker/Dockerfile-pytorch`:
```
  RUN pip install --no-cache-dir \
    git+https://github.com/ifzhang/ByteTrack.git \
    thop \
    lap \
    Cython

# Requires cython for install, so will fail if run in the same pip install as cython
RUN pip install --no-cache-dir \
    cython-bbox
```

  2. Delete `docker/Dockerfile-carla-mot`
 
  3. Modify the image mapping in `armory/docker/images.py`
```
PYTORCH = f"{DOCKER_REPOSITORY}/pytorch:{TAG}"
PYTORCH_DEEPSPEECH = f"{DOCKER_REPOSITORY}/pytorch-deepspeech:{TAG}"
TF2 = f"{DOCKER_REPOSITORY}/tf2:{TAG}"
CARLA_MOT = PYTORCH
ALL = (
    PYTORCH,
    PYTORCH_DEEPSPEECH,
    TF2,
    CARLA_MOT,
)
IMAGE_MAP = {
    "pytorch": PYTORCH,
    "pytorch-deepspeech": PYTORCH_DEEPSPEECH,
    "tf2": TF2,
    "carla-mot": PYTORCH,
}
```

davidslater: ...then we need to update build and release workflows. Will take a stab.

PR #1786, opened 2022-11-30
Merge pytorch, tf2, and carla mot images

christopherwoodall:
- [x] Consolidates Armory's `pytorch`, `tf2`, and `carla` images
- [x] Remove old images
- [x] Add new image to CI
- [x] Update hash map
- [x] Update `docker/build.py`
- [ ] Update dependencies in `pyproject.toml` and `environment.yaml`
- [x] Bump minimum Python version to 3.8 
- [x] Update documentation
- [ ] Fix `--interactive` displaying `i have no name` as the username (1)
- [ ] Expose ports  in Dockerfile (1)
- [ ] List volumes in Dockerfile (1)

1. These changes should eventually be moved into the base image, but that exceeds the scope of this PR.


davidslater:
If we are going to move in this direction, let's also remove `tf2` and `pytorch-deepspeech`, and just move directly to one container.

davidslater:
`TF2` can be mapped to `PYTORCH`. Actually, I would probably just map them all to `armory` or some generic container name.

the deep speech container will require some additional work:
a) removal of `armory/baseline_models/pytorch/deep_speech.py`
b) pinning of `eval1-4/asr_librispeech/`, `eval5/asr_librispeech/` to 0.16.1 version docker container (and adding deprecation)
c) expanding `eval6/asr_librispeech` to include targeted and targeted entailment attacks
d) various updates of docs

christopherwoodall:
Added fix for #1787

christopherwoodall:
I agree with the above, but I think we should scope that work out into a different PR.

davidslater:
Let's have `pytorch-deepspeech` be a different PR, but let's put `tf2` in this one, mapping it to PyTorch like you have done here with `carla-mot`. The current `tf2` and `pytorch` docker images are effectively identical.

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [project]
-requires-python = ">=3.7"
+requires-python = ">=3.8"
```
mwartell, line 2: pure fyi: the reason we deprecated 3.7 support was because we used `ubuntu-latest` in CI and when Github Actions bumped latest == 22.04 we lost python 3.7 there. While we could have patched it in from the `deadsnakes` PPA we thought that silly.

And, since we shouldn't claim support for that which we don't (or can't) test, then it is proper for us to declare 3.8+

pyproject.toml
```diff
@@ -115,8 +106,34 @@ deepspeech = [
...
+
+all = [
+    "armory-testbed[developer]",
+    "armory-testbed[engine]",
+    "armory-testbed[math]",
+    "armory-testbed[jupyter]",
+    "armory-testbed[datasets]",
+    "armory-testbed[datasets-builder]",
+    "armory-testbed[pytorch]",
+    # "armory-testbed[deepspeech]",
```
mwartell, line 132: dead code; should get a comment or be deleted

pyproject.toml
christopherwoodall: I'd like to add that support for 3.7 ends in 6 months as well.

armory/docker/images.py
```diff
@@ -15,22 +15,18 @@
...
-    PYTORCH,
+    ARMORY_IMAGE_NAME,
     PYTORCH_DEEPSPEECH,
-    TF2,
-    CARLA_MOT,
 )
 REPOSITORIES = tuple(x.split(":")[0] for x in ALL)
 IMAGE_MAP = {
-    "pytorch": PYTORCH,
+    "pytorch": ARMORY_IMAGE_NAME,
```
davidslater, line 26: Can you add a key-value pair `"armory": ARMORY_IMAGE_NAME,`?

This dict is used in `__main__` and in image lookups, and this would enable using `armory` on the command line or the scenario configs.

pyproject.toml
```diff
@@ -55,50 +54,36 @@ developer =[
...
+    "tensorboardx == 2.5.1",
 ]
 
 pytorch = [
     "armory-testbed[engine,datasets,math]",
     "torch",
-    "torchvision",
-    "torchaudio",
+    "torchaudio == 0.12.1",
+    "torchvision == 0.13.1",
```
davidslater, line 79: Why are these specified? If they are specified, `torch` version should be specified as well.

environment.yml
```diff
@@ -1,24 +1,42 @@
...
+
+  # ByteTrack (see below)
+  # TODO: Should this be added to pyproject.toml; e.g. [byte-track]?
+  - pytorch=1.12   # Required by ByteTrack, therefore must be installed pre-pip
+  - Cython         # ByteTrack
+  - lap            # ByteTrack
+
+  - pip:
+    # Hydra Lightning Configs
+    - 'git+https://github.com/romesco/hydra-lightning.git#subdirectory=hydra-configs-pytorch-lightning'
```
davidslater, line 34: No, this is part of pytorch-deepspeech.

environment.yml
```diff
@@ -1,24 +1,42 @@
...
-# yamllint disable rule:line-length
 name: armory-core
+
 channels:
-  - pytorch
-  - conda-forge
   - defaults
+  - conda-forge
+  - anaconda
+  - pytorch
```
davidslater, line 8: Why is the ordering changed?

pyproject.toml
```diff
@@ -55,50 +54,36 @@ developer =[
...
+    "boto3 == 1.24.89",      # Needed for armory.data.utils
+    "botocore == 1.27.89" ,  # Needed for armory.data.utils
+    "ffmpeg-python",         # Needed for armory.utils.export
+    "pydub",               # this is in ART's extra-requires
+    "tidecv",              # Needed for TIDE metrics
+    # Both `opencv-python` and `opencv-python-headless` must specify
+    # the same version.
+    "opencv-python == 4.5.5.62",          # Needed for CARLA baseline scenario
+    "opencv-python-headless == 4.5.5.62", # Needed for CARLA baseline scenario
+    "tensorboardx == 2.5.1",
```
davidslater, line 72: These are very specific version numbers. Why?

pyproject.toml
```diff
@@ -113,10 +98,43 @@ deepspeech = [
...
+    "armory-testbed[tensorflow]",
+    "tensorflow-datasets >= 4.6.0",
+    "protobuf",
+]
+
+datasets-builder = [
+    "armory-testbed[datasets,engine]",
+    "apache-beam >= 2.22.0",
+]
+
```
davidslater, line 119: Is there a reason these all got reordered? Hard to follow changes.

environment.yml
```diff
@@ -1,24 +1,42 @@
 ---
-# yamllint disable rule:line-length
 name: armory-core
```
davidslater, line 2: Changing the `environment.yml` will impact the base image. Not something we want to do with this mod.

davidslater (changes requested):
I have a lot of questions - probably better to discuss in person.

pyproject.toml
christopherwoodall: In order to speed up pip's resolver as well as to ensure compatibility with all the dependencies.

environment.yml
christopherwoodall: The changes do effect the base image, but all images continue to build as normal.

docker/Dockerfile-pytorch
```diff
@@ -21,7 +21,9 @@ RUN pip install --no-cache-dir \
         git+https://github.com/ifzhang/ByteTrack.git \
         thop \
         lap \
-        Cython && \
+        Cython \
+        # BURST dependency
+        tabulate && \
```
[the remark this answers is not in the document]
christopherwoodall: There is a dependency warning if not installed. My guess is that it is required by another package.

pyproject.toml
christopherwoodall: This change moves the frameworks to the top so that it is easier to find their dependencies.

davidslater (changes requested):
Please make this into 2 PRs:
1) A PR that does the basic changes to workflows, docker images, __main__, build and docs, but doesn't add any new features to `Dockerfile-armory` (e.g., docker user or port exposure) or make any mods to `environment.yml` or `pyproject.toml`.
2) A second PR (that can branch from the first one) that adds the new features, environmental changes, etc.?

christopherwoodall:
Split PR. See PR #1807

armory/docker/images.py
```diff
@@ -10,27 +10,19 @@
...
-ALL = (
-    PYTORCH,
-    PYTORCH_DEEPSPEECH,
-    TF2,
-    CARLA_MOT,
-)
-REPOSITORIES = tuple(x.split(":")[0] for x in ALL)
+TAG = version.to_docker_tag(armory.__version__)
+ARMORY_IMAGE_NAME = f"twosixarmory/armory:{TAG}"
+DEEPSPEECH_IMAGE_NAME = f"twosixarmory/deepspeech:{TAG}"
```
davidslater, line 18: Should be f"twosixarmory/pytorch-deepspeech:{TAG}"

environment.yml
```diff
@@ -21,4 +21,4 @@ dependencies:
   - numba  # ART requires numba >= 0.53.1
   - pandas
   - protobuf
-prefix: /opt/conda
+prefix: /opt/conda
```
davidslater, line 24: Add back in EOL

pyproject.toml
```diff
@@ -55,50 +61,35 @@ developer =[
...
+    "Pillow",              # Data dependencies
+    "boto3",               # Needed for armory.data.utils
+    "botocore" ,           # Needed for armory.data.utils
+    "ffmpeg-python",       # Needed for armory.utils.export
+    "pydub",               # this is in ART's extra-requires
+    "tidecv",              # Needed for TIDE metrics
+    # Both `opencv-python` and `opencv-python-headless` must specify
+    # the same version.
+    "opencv-python == 4.5.5.62",          # Needed for CARLA baseline scenario
+    "opencv-python-headless == 4.5.5.62", # Needed for CARLA baseline scenario
```
davidslater, line 78: Remind me why `opencv-python-headless` is needed

davidslater, line 78: Can we remove this pinning for this part and have it for part II instead?

davidslater (approved):
Two minor edits - but feel free to merge after fixing those.

Prior to merging, please build the armory container and push to dockerhub as `twosixarmory/armory:0.16.1`.

PR #1790, opened 2022-11-30
tfdsv4 resisc45

lcadalzo:


lcadalzo:
@davidslater do you recall why the dataset gets built to `<data_dir>/resisc45_split` rather than `<data_dir>/resisc45`? This discrepancy makes `load.load()` think the dataset doesn't exist locally, since it's looking for the path `<data_dir>/resisc45`

davidslater:
It was because the original dataset did not contain splits. However, I think that we're fine to just call it `resisc45`. 

You will want to incorporate `resisc45_dataset_partition.py` into your `_generate_examples method`, I think.

lcadalzo:
In `armory/data/resisc45/resisc45_split.py`, which I've copied to `armory/datasets`, the data is already split into 3 tar files. I'm able to build the data fine without touching `resisc45_dataset_partition.py`, but the data is built to `resisc45_split` dir:
```
I have no name!@b9bf7b1fac00:/workspace$ ls /armory/datasets/new_builds/resisc45_split/3.0.0/
dataset_info.json  resisc45_split-test.tfrecord-00000-of-00001   resisc45_split-train.tfrecord-00002-of-00004
features.json      resisc45_split-train.tfrecord-00000-of-00004  resisc45_split-train.tfrecord-00003-of-00004
label.labels.txt   resisc45_split-train.tfrecord-00001-of-00004  resisc45_split-validation.tfrecord-00000-of-00001


I have no name!@b9bf7b1fac00:/workspace$ ls /armory/datasets/new_builds/resisc45             
ls: cannot access '/armory/datasets/new_builds/resisc45': No such file or directory

```

if I change the dataset `"name"` in the config from `"resisc45"` to `"resisc45_split"`, I can load the data fine. But with the name `"resisc45"`, `load.load()` looks for `/armory/datasets/new_builds/resisc45` and throws an error.

davidslater:
Those URLs are what results from applying the resisc45_dataset_partition.py script to the original dataset `NWPU-RESISC45.tar.gz` and then breaking into separate files.

I think that we probably want to just reference `armory-public-data/resisc45/NWPU-RESISC45.tar.gz` and incorporate the script into the builder loop. Thoughts?

I can take a stab at it if you'd like.

lcadalzo:
after the above commit, I can build to `resisc45` dir without error, and this no longer uses the 3 separate files. Not quite sure why I needed to add the hardcoded `NWPU-RESISC45` in a couple places to get things working, though

armory/datasets/standard/resisc45/resisc45.py
```diff
@@ -0,0 +1,134 @@
...
+        for label in tf.io.gfile.listdir(f"{path}/NWPU-RESISC45"):
+            for idx, filename in enumerate(
+                tf.io.gfile.glob(f"{path}/NWPU-RESISC45/{label}/*.jpg")
+            ):
+                example = {
+                    "image": filename,
+                    "label": label,
+                    "filename": os.path.basename(filename),
+                }
+                if idx <= 500:
```
davidslater, line 127: There are three issues here. 

1. Currently, I'm not sure if glob will sort, so I would want to do a `sorted(enumerate(...` with your method (even if it does, that will make it explicit). 
2. Then you need to change the `<=` to `<`, since `enumerate` is 0-indexed but the dataset items are 1-indexed.
3. Since we are no longer enforcing explicit names, I would want to raise an error if `idx >= 700`. That would indicate that there is an erroneous file with a `.jpg` ending.

Alternatively, you could raise an error if `filename != f"{label}_{idx+1:03}.jpg"`. That would make it very clear what is going on.

armory/datasets/standard/resisc45/resisc45.py
```diff
@@ -0,0 +1,134 @@
...
+        """Returns SplitGenerators."""
+        path = dl_manager.download_and_extract(_URL)
+        return {
+            split: self._generate_examples(path, split)
+            for split in ["train", "validation", "test"]
+        }
+
+    def _generate_examples(self, path, split):
+        """Yields examples."""
+        for label in tf.io.gfile.listdir(f"{path}/NWPU-RESISC45"):
```
davidslater, line 118: Should we use something from `_LABEL` here? We should explicitly restrict to those, I would think.

Then you could do:
```python
for label in _LABEL:
    for idx, filename in sorted(enumerate(tf.io.gfile.glob(f"{path}/NWPU-RESISC45/{label}/*.jpg"))):
        ...
```

armory/datasets/standard/resisc45/checksums.tsv
```diff
@@ -0,0 +1,3 @@
+https://armory-public-data.s3.us-east-2.amazonaws.com/resisc45/resisc45_test.tar.gz	58428483	13670e5ce420c1bb9a2716a7d544405e4ee1b03247210d09da57a351ed46774d	resisc45_test.tar.gz
+https://armory-public-data.s3.us-east-2.amazonaws.com/resisc45/resisc45_train.tar.gz	291448728	4067ca9d0027ab69cae78d0b3f7c5596f83f586d144eb4bfb86ac7ad7e57db15	resisc45_train.tar.gz
+https://armory-public-data.s3.us-east-2.amazonaws.com/resisc45/resisc45_validation.tar.gz	58163558	db04359b48af71f8e27c62474e1d56f451e9f2c7861eea3e3d04fcf1414c906f	resisc45_validation.tar.gz
```
davidslater, line 3: This will need to be updated as well.

davidslater:
Builds fine for me.

lcadalzo:
Removing the WIP, I'm able to run a scenario and see expected benign/adv output. @davidslater ready for re-review

davidslater (changes requested):
Scenario works.

Can you add also do `package.add_to_cache("resisc45", public=True)` and commit the result (should change `armory/datasets/cached_datasets.json`).

lcadalzo:
done. Calling `add_to_cache()` also attempt to upload to s3, although this yielded an error for me. Not before `cached_datasets.json` was updated, though. Have you been using `upload()` successfully? I noticed that all the `url`'s in the json are `null`

davidslater:
Where did it error? Do you have `ARMORY_PRIVATE_S3_ID` and `ARMORY_PRIVATE_S3_KEY`?

davidslater:
You can break down that operation with:
```
from armory.datasets import package, upload
package.package("resisc45")
package.update("resisc45")
package.verify("resisc45")
upload.upload("resisc45", public=True)
```

PR #1794, opened 2022-12-01
fix intermittent macOS CI timing results

### LINKED ISSUE #1793: CI failure on macOS due to Action Runner scheduling
There is an intermittent error in where the time it takes to run different tests varies. https://github.com/twosixlabs/armory/actions/runs/3586468396/jobs/6035715746

In particular the expression `assert results[slow_key] > results[fast_key]` faults because sometimes the "fast" profile takes longer. This is in method `test_metric` of tests/unit/test_compute_metrics.py

mwartell:
Closes #1793 

With the meld of tf2 into pytorch (PR #1786) there is an intermittent problem with the slow test taking less average cpu on Github Action macOS runners despite slow expecting to be an order of magnitude slower.

Rather than try to refine the timing or predicate them by host OS, I've just removed the check. I've tagged @davidslater for review in case this warrants more clever action.

### DIFF
--- a/tests/unit/test_compute_metrics.py
+++ b/tests/unit/test_compute_metrics.py
@@ -37,7 +37,6 @@ def test_basic():
 
     assert fast_key in results
     assert slow_key in results
-    assert results[slow_key] > results[fast_key]
 
 
 def test_deterministic(caplog):


davidslater (approved):
This seems fine to me. You could turn it into a pytest warning instead of a failure.

PR #1795, opened 2022-12-01
Clear jupyter notebook output images

mwartell:
This has been bugging me since it was added. I've taken out two huge PNG output files embedded in the book.  This change dropped the size from 3800KB to 56KB.


christopherwoodall (approved):
LGTM

PR #1797, opened 2022-12-01
tfdsv4 xview

lcadalzo:
Can build the dataset successfully, still need to add preprocessing and ensure scenario / default attacks run successfully.


lcadalzo:
Current issue I'm hitting is that preprocessing returns `y` as a `dict` rather than `List[dict]` (of length 1) that Armory is expecting. Haven't figured out why `carla_over_obj_det` preprocessing returns `List[dict]` but `xview` differs. Running with `--skip-attack` for now, this is causing the following error in metrics computation:

```
 File "/workspace/armory/metrics/task.py", line 637, in object_detection_AP_per_class
    _check_object_detection_input(y_list, y_pred_list)
    │                             │       └ [{'boxes': array([[1.57085159e+02, 3.68701096e+01, 1.92630753e+02, 6.53191986e+01],
    │                             │                [2.45423927e+01, 1.08855011e+02, 6...
    │                             └ ['area', 'area', 'area']
    └ <function _check_object_detection_input at 0x7f38119eb820>

  File "/workspace/armory/metrics/task.py", line 480, in _check_object_detection_input
    f"y must contain the following keys: {REQUIRED_LABEL_KEYS}. The following keys were found: {y.keys()}"

AttributeError: 'str' object has no attribute 'keys'

```

lcadalzo:
still need to update other configs but otherwise this is ready for review @davidslater

PR #1804, opened 2022-12-05
1802: enable package() to handle non-standard dataset paths

### LINKED ISSUE #1802: TFDSv4: Handle non-standard dataset paths
We assume that datasets are stored in `<dataset_dir>/<name>/<version>/`. However, some have additional directories that may be related to builder configs (e.g., ucf101) or community datasets.

lcadalzo:
Closes #1802

### DIFF
--- a/armory/datasets/package.py
+++ b/armory/datasets/package.py
@@ -23,9 +23,8 @@ def package(
     )
 
     data_dir = Path(data_dir)
-    expected_dir = data_dir / name / version
-    if not expected_dir.is_dir():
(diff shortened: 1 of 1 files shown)


davidslater (approved):
This will work for default TFDS builder config. However, for non-standard configs (e.g., `ucf101_1` or so2sat `all`), the config needs to be piped through, I think.

Your call on whether to add that in a subsequent PR on in this one.

lcadalzo:
Supplanted by #1842

PR #1811, opened 2022-12-06
create parent directories if needed

davidslater:


davidslater:
This is probably the first error you will run into if you're trying to load or create a new dataset

lcadalzo:
@jprokos26 I'm lacking context on this one, can this be merged?

davidslater:
Very little context. The first time you use the cache functionality, it will throw a directory does not exist error. This makes the cache base directory if it doesn't exist.

jprokos26:
I did not run into this error while building `carla_video_tracking` datasets. I suspect the parent directory had already been created when I integrated Eval6 datasets from Mitre. Regardless adding to cache works as expected with this change.

mwartell (approved):

Issue #1620, opened 2022-09-29
remove references to getenv("ARMORY_VERSION") throughout


mwartell:
There are places e.g. `environment.py` which look to the environment for a version string. They should be re-pointed to use the 0.16 version primitives.

davidslater:
I think the only reason this was done was to log a warning if the armory launcher has a different version than armory running the scenario:
https://github.com/twosixlabs/armory/blob/master/armory/scenarios/main.py#L211-L216

christopherwoodall:
Just to start discussion around the issue, is `armory_launcher` a [different application](https://12factor.net/codebase)  from `armory_scenarios` or just a way to interact with the scenario runner "API"? If the later is true(the launcher is an API client) should the version of the client matter or of the API?

davidslater:
There isn't a clean distinction or API at the moment. Probably will become less of an issue once we do a better job of separating the two.

davidslater:
Rename to `CLIENT_ARMORY_VERSION`? Used to log a warning if the version used in the container differs from the version used in the client (since we don't have an API versioning mechanism)

PR #1820, opened 2022-12-10
Fix for Issue 1620 - remove references to getenv("ARMORY_VERSION")

christopherwoodall:
Removes references to `getenv("ARMORY_VERSION")`.

see Issue #1620

### DIFF
--- a/armory/docker/management.py
+++ b/armory/docker/management.py
@@ -27,11 +27,11 @@ class ArmoryInstance(object):
     ):
         self.docker_client = docker.from_env(version="auto")
 
-        host_paths = paths.HostPaths()
-        docker_paths = paths.DockerPaths()
-
         envs = envs if envs else {}
         envs["ARMORY_VERSION"] = armory.__version__
+
+        host_paths = paths.HostPaths()
+        docker_paths = paths.DockerPaths()
 
         mounts = [
             docker.types.Mount(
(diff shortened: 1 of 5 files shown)


armory/docker/management.py
```diff
@@ -20,7 +20,7 @@ def __init__(
         self,
         image_name,
         runtime: str = "runc",
-        envs: dict = None,
+        envs: dict = {},
```
davidslater, line 23: Having a dict as a default parameter, since it is mutable (and hence can have weird side-effect problems), and is generally not recommended.

Instead, I recommend that you keep it None in the method definition and then add this below:
```
if envs is None:
    envs = {}
```

armory/docker/management.py
```diff
@@ -30,6 +30,8 @@ def __init__(
         host_paths = paths.HostPaths()
         docker_paths = paths.DockerPaths()
 
+        envs["ARMORY_VERSION"] = armory.__version__
+
```
davidslater, line 34: Having this as an override to what the evaluator adds seems both redundant and hard to follow. I would recommend you set this in one or the other.

armory/scenarios/main.py
```diff
@@ -208,7 +208,7 @@ def run_config(*args, **kwargs):
     args = parser.parse_args()
     update_filters(args.log_level, args.debug)
     log.trace(f"main.py called update_filters({args.log_level} debug: {args.debug})")
-    calling_version = os.getenv(environment.ARMORY_VERSION, "UNKNOWN")
+    calling_version = os.environ.get("ARMORY_VERSION", "UNKNOWN")
```
davidslater, line 211: I'm not a fan of replacing `environment.ARMORY_VERSION` with a static string `"ARMORY_VERSION"` everywhere. That makes it more difficult to change in the future.

Instead, let's just rename the attribute in environment to be more understandable, like:
```
CLIENT_ARMORY_VERSION = "ARMORY_VERSION"
```
and then use `CLIENT_ARMORY_VERSION` throughout.

davidslater (changes requested):

PR #1821, opened 2022-12-12
Instrumentation Examples with Probes, Meters, and Writers

ppark-twosixtech:
This PR was the outcome of discussions on how to make `Probe`s and `Meter`s more approachable to the user, with the conclusion that more documentation with specific examples would be more beneficial then defining hooking methods in the `Probe` class for different packages (i.e. `pytorch`, `art`, etc.). We determined that defining hooking methods in this manner would not be sustainable since they would have to be maintained by the GARD program should any of those packages change in the future. The templates introduced in this documentation show a flexible approach in how `Probe`s and `Meter`s can fit into other packages/frameworks, and avoids the need to figure out how other packages need to fit into `armory`. Code and documentation referring to any defined hooking methods for `Probe`s have been removed as well.


docs/probe.md
```diff
@@ -0,0 +1,177 @@
+# Probes and Meters: Examples for Adding to Code
+For an introduction to `Probe`s and `Meter`s, please refer to [Measurement Overview](./metrics.md).
+
+## Context
+To monitor particular aspects of an `armory run` session, the user needs to know the following factors:
+- What am I measuring?
+- When should I measure it?
+- Where should my custom monitoring script go?
+
+We assume the user is modifying code for models or attacks and wishes to use `Probe`s and `Meter`s to monitor certain variables within the code.
```
lcadalzo, line 10: instead of `modifying code for models or attacks` could we be more descriptive like `capturing artifacts from the model or attack`

docs/probe.md
```diff
@@ -0,0 +1,177 @@
+# Probes and Meters: Examples for Adding to Code
+For an introduction to `Probe`s and `Meter`s, please refer to [Measurement Overview](./metrics.md).
+
+## Context
```
lcadalzo, line 4: do we discuss at all where to find the outputted results of what we're measuring?

docs/probe.md
```diff
@@ -0,0 +1,177 @@
...
+        output = F.log_softmax(x, dim=1)
+        return output
+
+...
+```
+
+
+
+#### Step 1
+After importing `get_probe`, `probe = get_probe("my_model")` creates a `Probe` object with the name `"my_model"`, which is what the user can refer to later to apply processing functions through a `Meter` object.
```
lcadalzo, line 77: maybe add a pointer to where in the code i.e `In line X...`

docs/probe.md
```diff
@@ -0,0 +1,177 @@
...
+
+#### Step 2
+The only difference between `CustomAttack` and `CARLADapricotPatch` is that `_augment_images_with_patch` has been redefined to call on `CARLADapricotPatch._augment_images_with_patch` and then have `probe` update the value for `x_patch` that results from that call. An updated value of `x_patch` is stored in `attack_output` via `probe.update(attack_output=x_patch)`. Like the `Probe` name `"my_attack"`, `attack_output` can be referenced by the user later to apply additional processing functions through a `Meter` object.
+
+### `Meter` Example Code
+Now that a `Probe` instance has been created, we need to create a `Meter` object to accept any updated values from `Probe` and apply further processing that the user desires. Suppose the user created a script located at `armory/user_init.py`:
+```python
+from armory.instrument import get_hub, Meter
+
+def user_init_function():
```
lcadalzo, line 147: rename `user_init_function()` to something more descriptive a la `create_meter()` or `set_up_meter()`?

lcadalzo: and same for above

docs/probe.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Context
+To monitor particular aspects of an `armory run` session, the user needs to know the following factors:
+- What am I measuring?
+- When should I measure it?
+- Where should my custom monitoring script go?
+
+We assume the user is modifying code for models or attacks and wishes to use `Probe`s and `Meter`s to monitor certain variables within the code.
+
+Recall the steps for a minimal working example (in [Measurement Overview](./metrics.md)):
```
lcadalzo, line 12: I'd link specifically to `/metrics.md#instrumentation`

lcadalzo, line 12: in `metrics.md`, where we say `Meters and writers can be instantiated in your initial setup`, can we link to the user_init docs here?

docs/probe.md
```diff
@@ -0,0 +1,177 @@
...
+`x_out = x.detach().cpu().numpy()` is taking the layer output of interest (second `relu` activation output) and converting the tensor to `numpy` array on the CPU, which will be passed to `probe`. An updated value of `x_out` is stored in `layer_output` via `probe.update(layer_output=x_out)`. Like the `Probe` name `"my_model"`, `layer_output` can be referenced by the user later to apply additional processing functions through a `Meter` object.
+
+### `Meter` Example Code
+Now that a `Probe` instance has been created, we need to create a `Meter` object to accept any updated values from `Probe` and apply further processing that the user desires. Suppose the user created a script located at `armory/user_init.py`:
+```python
+from armory.instrument import get_hub, Meter
+
+def user_init_function():
+    meter = Meter(
+        "my_layer_identity", lambda x: x, "my_model.layer_output"
```
lcadalzo, line 89: would `"my_arbitrary_meter_name"` be clearer than `"my_layer_identity"`? Same for example 2

docs/probe.md
```diff
@@ -0,0 +1,177 @@
...
+We assume the user is modifying code for models or attacks and wishes to use `Probe`s and `Meter`s to monitor certain variables within the code.
+
+Recall the steps for a minimal working example (in [Measurement Overview](./metrics.md)):
+1. Create `Probe` via `get_probe(name)`
+2. Place `Probe` actions
+3. Create `Meter` with processing functions that take input from created `Probe`
+4. Connect `Meter` to `Hub` via `get_hub().connect_meter(meter)`
+
+The examples will show how each of these steps are accomplished.
+
```
lcadalzo, line 19: the two examples look good. Just to confirm, you did successfully run these correct?

docs/probe.md
```diff
@@ -0,0 +1,177 @@
+# Probes and Meters: Examples for Adding to Code
```
lcadalzo, line 1: I think `probe.md` could use a better name, but not sure exactly what's right. This doc to me feels like "examples of using Armory's instrumentation feature"

docs/probe.md
ppark-twosixtech: by default, output will go to the `json` file, but there's a separate issue where we run in to record size restrictions such as:
> 2022-12-16 19:34:36 30s WARNING  armory.instrument.instrument:_write:856 record (name=my_layer_identity, batch=0, result=...) size > max_record_size 1048576. Dropping.
which will depend on what the user is trying to capture. It just so happens that the examples shown here do run into those issues, but that should be its own section about defining a new class of `Meter` to:
1. work around the size limitation
2. and save to a `pkl` file rather than `json`

ppark-twosixtech: that said, I guess we should just mention that outputs will go to the `json` file while we figure out a better alternative?

docs/probe.md
ppark-twosixtech: random question, is there a good way to add line numbers to code blocks?

docs/probe.md
```diff
@@ -0,0 +1,177 @@
...
+
+Recall the steps for a minimal working example (in [Measurement Overview](./metrics.md)):
+1. Create `Probe` via `get_probe(name)`
+2. Place `Probe` actions
+3. Create `Meter` with processing functions that take input from created `Probe`
+4. Connect `Meter` to `Hub` via `get_hub().connect_meter(meter)`
+
+The examples will show how each of these steps are accomplished.
+
+## Example 1: Model Layer Output
```
lcadalzo, line 20: `Model Layer Output` could be clearer as `Measuring a Model Layer's Output`. Simiarly `Attack Layer Output` --> `Measuring Attack Artifact` or something along those lines that's more explicit as to what the example is doing

docs/probe.md
ppark-twosixtech: the examples run, but there was the relevant question asked before about what to expect for outputs, which will depend on what the `Probe` is monitoring...

docs/probe.md
lcadalzo: I'm actually starting to wonder if a better home for this kind of demo is in the `armory/notebooks` dir

docs/probe_hooking_notebook.md
```diff
@@ -0,0 +1,316 @@
+# Probes and Meters: Advanced Hooking Examples for Notebooks
```
lcadalzo, line 1: I like the idea of having some sort of hooking example in the docs. Ideally it could go in the same doc (or notebook?) and be a more advanced Example 3

docs/probe.md
ppark-twosixtech: resolved with 89d200a6

docs/probe.md
ppark-twosixtech: resolved with 89d200a6

docs/probe.md
ppark-twosixtech: resolved with 89d200a6

docs/probe.md
ppark-twosixtech: resolved with 89d200a6

docs/probe.md
ppark-twosixtech: resolved with 89d200a6

docs/probe.md
ppark-twosixtech: > I think `probe.md` could use a better name, but not sure exactly what's right. This doc to me feels like "examples of using Armory's instrumentation feature"

any thoughts on updates in 89d200a6?

docs/probe.md
ppark-twosixtech: checkout f402f30d

docs/probe.md
ppark-twosixtech: check out f402f30d

docs/probe_hooking_notebook.md
ppark-twosixtech: Tabling for now based on brief discussion of how to insert a `Probe` from a notebook without hooking, but redefining a function (e.g. `forward`) of a model or attack. The exact steps for this particular approach has not been determined yet.

docs/instrumentation_example.md
```diff
@@ -1,4 +1,4 @@
-# Probes and Meters: Examples for Adding to Code
+# Armory Instrumentation Examples: Capture Artifacts from Existing Code
```
lcadalzo, line 1: I might change `Capture Artifacts from Existing Code` to `Measuring Experiment Artifacts Using Probes and Meters`. And `instrumentation_example.md` --> `instrumentation_examples.md`. But otherwise this commit looks good

docs/probe.md
ppark-twosixtech: Resolved with dd462e2b

docs/instrumentation_example.md
```diff
@@ -174,4 +174,60 @@ Last but not least, the config file passed to `armory run` needs to updated for
...
+
+        self.max_batches = max_batches
+        self.metric_args = metric_args
+        self.output_dir = output_dir
+        self.iter = 0
+
+        if not os.path.exists(self.output_dir):
+            os.mkdir(self.output_dir)
+
+    def measure(self, clear_values=True):
```
lcadalzo, line 215: it seems like perhaps rather than overwriting `measure()`, a cleaner solution would be to add an image-specific `Writer` class to the `Meter`, so that `writer.write()` would do its thing like normal (although you'd need a custom `Writer` class that overwrites `write()` I believe)

docs/probe.md
ppark-twosixtech: Added line number references with 1371a49b - would be more useful if line numbers could be included in the code block, but I couldn't find a way to do that...

docs/instrumentation_example.md
ppark-twosixtech: resolved with ee52a9e7

docs/instrumentation_example.md
ppark-twosixtech: resolved with 72577b7

docs/instrumentation_examples.md
```diff
@@ -0,0 +1,214 @@
...
+
+Users who have tried the examples in this document, however, may run into some of the following warning logs:
+> 2022-12-16 19:34:36 30s WARNING  armory.instrument.instrument:_write:856 record (name=my_arbitrary_meter_name, batch=0, result=...) size > max_record_size 1048576. Dropping.
+
+Outputs are saved to a `json` file because of a default `ResultWriter` class tied to the `Meter` class, which has a `max_record_size` limit for each record. Any record that exceeds `max_record_size` will not save to the `json` file. That the outputs exceed a size limit also suggests that a `json` file may not be the best file type to save to. To work around these behaviors, we can define a new `Writer` subclass (`ResultWriter` is also a `Writer` subclass) to work with our examples that does not have a size limit and will save to another filetype such as a `pkl` file. Below is an updated `user_init.py` for Example 2 with a new `PickleWriter` class and `set_up_meter_writer` function that will be executed with the `user_init` block:
+```python
+from armory.instrument import get_hub, Meter, Writer
+import pickle
+
+class PickleWriter(Writer):
```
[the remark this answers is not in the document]
ppark-twosixtech: resolved with dd8feb7f

lcadalzo:
@ppark-twosixtech `develop` is now clean, so once you add a commit to remove probe hook tests, we should expect to see all tests passing

docs/instrumentation_examples.md
```diff
@@ -0,0 +1,214 @@
+# Armory Instrumentation Examples: Measuring Experiment Artifacts Using Probes and Meters
```
lcadalzo, line 1: I think we should also update this link to `metrics.md#instrumentation`

lcadalzo:
this predates your PR, but in `metrics.md` we say 
```
Strings must be a valid armory metric from armory.utils.metrics
```

Given our now support for custom metrics, this would be a good time to remove this

docs/instrumentation_examples.md
```diff
@@ -0,0 +1,214 @@
...
+For an introduction to `Probe`s and `Meter`s, please refer to [Measurement Overview](./metrics.md).
+
+## Context
```
lcadalzo, line 4: I think you can delete the `Context` header, the three bullets, and the sentence above the bullets. 

Originally I was thinking the third bullet may be confusing in that seems to imply you need to create some separate custom monitoring script, as opposed to dropping probes into your already-existing code (I understand there's also the `user_init` piece, but don't think we need to throw that at the reader at the jump). After removing this bullet, I think it may be more straightforward to just jump straight into `We assume the user...` followed by `Recall the steps...`

docs/instrumentation_examples.md
```diff
@@ -0,0 +1,214 @@
...
+1. Create `Probe` via `get_probe(name)`
+2. Place `Probe` actions
+3. Create `Meter` with processing functions that take input from created `Probe`
+4. Connect `Meter` to `Hub` via `get_hub().connect_meter(meter)`
+
+The examples will show how each of these steps are accomplished.
+
+## Example 1: Measuring a Model Layer's Output
+### User Story
+I am interested in layer output from the second `relu` activation of a `forward` method located in `armory/baseline_models/pytorch/cifar.py`.
```
lcadalzo, line 22: `in layer output` --> `in the layer output`

docs/instrumentation_examples.md
```diff
@@ -0,0 +1,214 @@
...
+from armory.instrument import get_hub, Meter
+
+def set_up_meter():
+    meter = Meter(
+        "my_arbitrary_meter_name", lambda x: x, "my_model.layer_output"
+    )
+    get_hub().connect_meter(meter)
+```
+#### Step 3
+In this particular example, the `Meter` accepts 3 inputs (line 5): a meter name, a metric/function for processing, and a argument name to pass the metric/function.
```
lcadalzo, line 94: I would omit `line 5` here, `line 7` below, etc.

docs/instrumentation_examples.md
```diff
@@ -0,0 +1,214 @@
...
+
+
+
+#### Step 1
+After importing `get_probe` in line 11, `probe = get_probe("my_model")` in line 12 creates a `Probe` object with the name `"my_model"`, which is what the user can refer to later to apply processing functions through a `Meter` object.
+
+#### Step 2
+`x_out = x.detach().cpu().numpy()` in line 36 is taking the layer output of interest (second `relu` activation output) and converting the tensor to `numpy` array on the CPU, which will be passed to `probe`. An updated value of `x_out` is stored in `layer_output` via `probe.update(layer_output=x_out)` in line 37. Like the `Probe` name `"my_model"`, `layer_output` can be referenced by the user later to apply additional processing functions through a `Meter` object.
+
+### `Meter` Example Code
```
lcadalzo, line 82: my proposed rewriting of the text below `Meter Example Code`:

```
Now that a Probe instance has been created, we need to create a Meter object to accept any updated values from Probe and apply further processing that the user desires. We can create the Meter in a function added to a local Python script, then pass this function to the `user_init` field in our JSON config to ensure the code is run during scenario initialization, prior to scenario execution. Please refer to etc. etc. for more details on the user_init field and how to configure it.
```

docs/instrumentation_examples.md
```diff
@@ -0,0 +1,214 @@
...
+- The argument passed to the metric/function follows a `.`-separated format (`"my_model.layer_output"`), which needs to be consistent with `Probe` setup earlier:
+    - `my_model` matches input in `probe = get_probe("my_model")`
+    - `layer_output` matches variable name in `probe.update(layer_output=x_out)`
+
+#### Step 4
+For the scope of this document, we don't dwell on what `armory` is doing with `get_hub().connect_meter(meter)` in line 7 other than to mention this step is necessary for establishing the connection between `meter` created in `armory/user_init.py` and `probe` created in the modified version of `armory/baseline_models/pytorch/cifar.py`.
+
+### Config Setup
+Last but not least, the config file passed to `armory run` needs to be updated for these changes to take effect. Assuming the `"model"` block has been changed appropriately, the other block that needs to be added is `"user_init"` (please refer to [User Initialization](./scenarios.md#user-initialization) for more details about using the `user_init` block):
```
lcadalzo, line 105: the `model` field shouldn't need to be changed, no?

lcadalzo, line 105: we just linked user_init docs, so I'd say no need to do it again so shortly after

docs/instrumentation_examples.md
```diff
@@ -0,0 +1,214 @@
...
+3. Create `Meter` with processing functions that take input from created `Probe`
+4. Connect `Meter` to `Hub` via `get_hub().connect_meter(meter)`
+
+The examples will show how each of these steps are accomplished.
+
+## Example 1: Measuring a Model Layer's Output
+### User Story
+I am interested in layer output from the second `relu` activation of a `forward` method located in `armory/baseline_models/pytorch/cifar.py`.
+### `Probe` Example Code
+The code below is an example of how to accomplish steps 1 and 2 (note the lines of code with `# added` comments at the end) for a model code that the user is modifying.
```
lcadalzo, line 24: I would trim as many lines in the code snippet as possible that aren't showing what we really care about here. I.e. the docstring at the top, the various non-instrument imports, `DEVICE` assignment, etc.

docs/instrumentation_examples.md
```diff
@@ -0,0 +1,214 @@
...
+        x_patch, patch_target, transformations = return_value
+        probe.update(attack_output=x_patch)
+
+        return return_value
+```
+#### Step 1
+This step is the same as before, except `Probe` name is set to`"my_attack"` in line 3, which is what the user can refer to later to apply processing functions through a `Meter` object.
+
+#### Step 2
+The only difference between `CustomAttack` and `CARLADapricotPatch` is that `_augment_images_with_patch` in line 6 has been redefined to call on `CARLADapricotPatch._augment_images_with_patch` and then have `probe` update the value for `x_patch` that results from that call (line 7-9). An updated value of `x_patch` is stored in `attack_output` via `probe.update(attack_output=x_patch)`. Like the `Probe` name `"my_attack"`, `attack_output` can be referenced by the user later to apply additional processing functions through a `Meter` object.
```
lcadalzo, line 140: I'd delete the line references since it's a short snippet, a bummer that github doesn't let you include line numbers

docs/instrumentation_examples.md
```diff
@@ -0,0 +1,214 @@
...
+- Again, `Meter` will record variables monitored by `Probe` as-is (thus `lambda x: x`)
+- The argument passed to the metric/function follows a `.`-separated format (`"my_attack.attack_output"`), which needs to be consistent with `Probe` setup earlier:
+    - `my_attack` matches input in `probe = get_probe("my_attack")`
+    - `attack_output` matches variable name in `probe.update(attack_output=x_patch)`
+
+#### Step 4
+Again, `get_hub().connect_meter(meter)` in line 7 is necessary for establishing the connection between `meter` created in `armory/user_init.py` and `probe` created in `armory/custom_attack.py`.
+
+### Config Setup
+Last but not least, the config file passed to `armory run` needs to be updated for these changes to take effect. Assuming the `"attack"` block has been changed appropriately, the other block that needs to be added is `"user_init"` (please refer to [User Initialization](./scenarios.md#user-initialization) for more details about using the `user_init` block):
```
lcadalzo, line 165: attack field shouldn't need to be "changed". We're already assuming the user is pointing to the attack that the probe is added to

docs/instrumentation_examples.md
```diff
@@ -0,0 +1,214 @@
...
+> 2022-12-16 19:34:36 30s WARNING  armory.instrument.instrument:_write:856 record (name=my_arbitrary_meter_name, batch=0, result=...) size > max_record_size 1048576. Dropping.
+
+Outputs are saved to a `json` file because of a default `ResultWriter` class tied to the `Meter` class, which has a `max_record_size` limit for each record. Any record that exceeds `max_record_size` will not save to the `json` file. That the outputs exceed a size limit also suggests that a `json` file may not be the best file type to save to. To work around these behaviors, we can define a new `Writer` subclass (`ResultWriter` is also a `Writer` subclass) to work with our examples that does not have a size limit and will save to another filetype, such as a `png` file, since we are saving data for an image. Below is an updated `user_init.py` for Example 2 with a new `ImageWriter` class, which uses the `export` method of `ImageClassificationExporter` to save an image, and a `set_up_meter_writer` function that will be executed with the `user_init` block:
+```python
+from armory.instrument import get_hub, Meter, Writer
+from armory.instrument.export import ImageClassificationExporter
```
lcadalzo, line 188: minor point here but since Example2 is object detection, can we use an object detection exporter? Should work the same way

docs/instrumentation_examples.md
```diff
@@ -0,0 +1,214 @@
...
+class ImageWriter(Writer):
+    def __init__(self, output_dir):
+        super().__init__()
+        self.output_dir = output_dir
+        self.iter_step = 0
+        self.batch = 0
+        self.exporter = ImageClassificationExporter(self.output_dir)
+
+    def _write(self, name, batch, result):
+        if batch != self.batch:
```
lcadalzo, line 199: is there a clearer way than `if batch != self.batch` to handle whatever this conditional block is doing?

docs/probe_hooking_notebook.md
lcadalzo, line 1: I think particularly examples 2 and 3 are far too complex and lengthy to include in our official docs. Perhaps we could add example 1 to `instrumentation_examples.md`, although I think we should follow the same protocol of `user_init` rather than show a drastically different approach of using `get_scenario()` and `s.run_attack()` etc. We have separate docs for showing how to run scenarios this way instead of the normal `armory run`

docs/instrumentation_examples.md
ppark-twosixtech: > my proposed rewriting of the text below `Meter Example Code`:
> 
> ```
> Now that a Probe instance has been created, we need to create a Meter object to accept any updated values from Probe and apply further processing that the user desires. We can create the Meter in a function added to a local Python script we'll name `user_init.py`. In `instrumentation_examples.md#config-setup` shortly below, we'll show how to ensure this code is run during scenario initialization.
> ```
> 
> this defers mentioning of the user_init feature until later (to not bombard the reader with new terminology) but also satisfies a reader immediately wondering "well how do I ensure this is run"

Removed first link to address above comment, so will keep this one.

docs/instrumentation_examples.md
ppark-twosixtech: resolved with c24837a7

docs/instrumentation_examples.md
ppark-twosixtech: resolved with c24837a7

docs/instrumentation_examples.md
ppark-twosixtech: resolved with c24837a7

docs/instrumentation_examples.md
ppark-twosixtech: resolved with c24837a7

docs/instrumentation_examples.md
ppark-twosixtech: resolved with 9d35fa8c

docs/instrumentation_examples.md
ppark-twosixtech: resolved with 445a2ab4

docs/instrumentation_examples.md
ppark-twosixtech: resolved with 445a2ab4

docs/instrumentation_examples.md
ppark-twosixtech: thought that the user may move model file or name it differently to keep original file - removed phrase on changing `"model"` block in 8cf85780

docs/instrumentation_examples.md
ppark-twosixtech: resolved with 8cf85780

ppark-twosixtech:
> this predates your PR, but in `metrics.md` we say
> 
> ```
> Strings must be a valid armory metric from armory.utils.metrics
> ```
> 
> Given our now support for custom metrics, this would be a good time to remove this

resolved with c24837a

docs/instrumentation_examples.md
ppark-twosixtech: resolved with f1fbb0c3

docs/instrumentation_examples.md
ppark-twosixtech: resolved with 30a5b921

docs/probe_hooking_notebook.md
ppark-twosixtech: removed for now (b282870d) after discussion with @lcadalzo to avoid confusion

docs/instrumentation_examples.md
```diff
@@ -0,0 +1,181 @@
...
+
+Users who have tried the examples in this document, however, may run into some of the following warning logs:
+> 2022-12-16 19:34:36 30s WARNING  armory.instrument.instrument:_write:856 record (name=my_arbitrary_meter_name, batch=0, result=...) size > max_record_size 1048576. Dropping.
+
+Outputs are saved to a `json` file because of a default `ResultWriter` class tied to the `Meter` class, which has a `max_record_size` limit for each record. Any record that exceeds `max_record_size` will not save to the `json` file. That the outputs exceed a size limit also suggests that a `json` file may not be the best file type to save to. To work around these behaviors, we can define a new `Writer` subclass (`ResultWriter` is also a `Writer` subclass) to work with our examples that does not have a size limit and will save to another filetype, such as a `png` file, since we are saving data for an image. Below is an updated `user_init_script.py` for Example 2 with a new `ImageWriter` class, which uses the `export` method of `ImageClassificationExporter` to save an image, and a `set_up_meter_writer` function that will be executed with the `user_init` block:
```
lcadalzo, line 152: one super minor comments: we changed the code to `ObjectDetectionExporter`, but the text still says `ImageClassificationExporter`. We can merge once this change is made, no need to wait for CI again

ppark-twosixtech: resolved with 6de62f2

lcadalzo (approved):
looks good

PR #1827, opened 2022-12-19
Eval6 - CARLA MOT test data

yusong-tan:


lcadalzo:
@swsuggs let's hold off on merging until the issue with CI on `develop` is addressed, but this can still otherwise now be reviewed

swsuggs (approved):
The test succeeds, the data loads, and the mot scenario still runs.

lcadalzo:
re-running CI that should now pass with the updates to develop, and we can merge once it does

PR #1833, opened 2022-12-21
remove optional art dependency

mwartell:
ART requires `numba` only to support BrendelBethgeAttack which is only loaded if numba is present.


mwartell:
Related to #1835. That fixes changes demanded by numpy 1.24.0 (recently released) and this removes the (unused) numba library which was forcing numpy version.

lcadalzo:
LGTM, just making sure CI doesn't yield any new non-numpy errors

lcadalzo (approved):

Issue #1832, opened 2022-12-21
CI Numpy Failures


lcadalzo:
See [here](https://github.com/twosixlabs/armory/actions/runs/3743705462) for example action for how CI is currently failing due to numpy errors, likely resulting from the new numpy 1.24.0 release on 12/18

Bumping to numpy 1.24.0 may not be trivial as I tried installing it in an `--interactive` Armory container and got the following:
```
numba 0.56.2 requires numpy<1.24,>=1.18, but you have numpy 1.24.0 which is incompatible.
```

Then as a result when I ran the scenario I got:
```

Traceback (most recent call last):
  File "/opt/conda/lib/python3.9/runpy.py", line 188, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/opt/conda/lib/python3.9/runpy.py", line 111, in _get_module_details
    __import__(pkg_name)
  File "/workspace/armory/scenarios/__init__.py", line 4, in <module>
    from armory import delayed_imports
  File "/workspace/armory/delayed_imports.py", line 42, in <module>
    paths.set_art_data_path()
  File "/workspace/armory/paths.py", line 35, in set_art_data_path
    from art import config
  File "/opt/conda/lib/python3.9/site-packages/art/__init__.py", line 7, in <module>
    from art import attacks
  File "/opt/conda/lib/python3.9/site-packages/art/attacks/__init__.py", line 8, in <module>
    from art.attacks import evasion
  File "/opt/conda/lib/python3.9/site-packages/art/attacks/evasion/__init__.py", line 17, in <module>
    from art.attacks.evasion.brendel_bethge import BrendelBethgeAttack
  File "/opt/conda/lib/python3.9/site-packages/art/attacks/evasion/brendel_bethge.py", line 50, in <module>
    from numba.experimental import jitclass
  File "/opt/conda/lib/python3.9/site-packages/numba/__init__.py", line 42, in <module>
    from numba.np.ufunc import (vectorize, guvectorize, threading_layer,
  File "/opt/conda/lib/python3.9/site-packages/numba/np/ufunc/__init__.py", line 3, in <module>
    from numba.np.ufunc.decorators import Vectorize, GUVectorize, vectorize, guvectorize
  File "/opt/conda/lib/python3.9/site-packages/numba/np/ufunc/decorators.py", line 3, in <module>
    from numba.np.ufunc import _internal
SystemError: initialization of _internal failed without raising an exception

```

[...]

PR #1835, opened 2022-12-21
fix numpy ci failures

lcadalzo:
Addressing CI failures from #1832 resulting from jump to numpy 1.24

### DIFF
--- a/armory/metrics/statistical.py
+++ b/armory/metrics/statistical.py
@@ -117,7 +117,7 @@ def make_contingency_tables(
         A map (Dict[int, np.ndarray]) of the per-class contingency tables.
     """
 
-    y = np.array(y).astype(np.int).flatten()
+    y = np.array(y).astype(np.int64).flatten()
     flagged_A = np.array(flagged_A).astype(np.bool_).flatten()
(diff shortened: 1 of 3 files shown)


lcadalzo:
@mwartell this PR does not intend to address any of the numba issues (as I understand you're working on that). The sole purpose of this PR is to fix the numpy CI failures

mwartell (approved):
Good pending tests. Those are some pretty basic interface elements to change in a numpy minor release.

lcadalzo:
Linking the Scenario Runner tests [here](https://github.com/lcadalzo/armory/actions/runs/3752128457). I would expect these to fail due to numba incompatibility with numpy 1.24.0, although...

@mwartell do you know the expected behavior of Scenario Runner if the config JSON's it runs are using `twosixarmory:pytorch` and `twosixarmory:carla-mot`, etc. images that are no longer built by `develop` after we consolidated images? Will it simply pull the `twosixarmory:pytorch:0.16.1` image? If so, I'd expect Scenario Runner to pass since that has numpy 1.23

mwartell (approved):

mwartell:
@lcadalzo Yes that is the proper expectation.

PR #1829, opened 2022-12-20
TFDSv4 carla_video_tracking upgrade.

jprokos26:
Throws an error when run without flags but completes successfully with `--skip-attack`. See https://github.com/twosixlabs/armory/issues/1814 for more information.

Status:
- [x] `carla_video_tracking_dev`
- [x] `carla_video_tracking`

Both uploaded to s3 and ready for review.

Issue #1814, opened 2022-12-06
Update object detection attacks for tfdsv4 label format


lcadalzo:
The tfdsv4 pipeline outputs object detection labels with a batch dimension, e.g. for an image with N boxes, the new pipeline outputs scores with a shape of `(1, N)` as opposed to shape `(N,)`. This will require minor modification to attacks in `armory.art_experimental` to account for this (as well as in metrics code?). 

To yield the errors currently present on the `tfdsv4` branch, simply run any carla object detection config. As part of this issue, we'll need to figure out if this impacts all object detection datasets or just certain ones. This issue can be considered fixed once we can run all object detection configs without error.

Issue #1803, opened 2022-12-05
TFDSv4: store contextual information from datasets, such as sample rate and frame rate


davidslater:
Currently, this information is needed in the following scenarios:
```
audio_asr.py:            self.test_dataset.context.sample_rate,
audio_classification.py:            self.test_dataset.context.sample_rate,
carla_video_tracking.py:            frame_rate=self.test_dataset.context.frame_rate,
carla_video_tracking.py:            frame_rate=self.test_dataset.context.frame_rate,
video_ucf101_scenario.py:            frame_rate=self.test_dataset.context.frame_rate,
```

This needs to be captured somewhere in the data loading pipeline.

PR #1836, opened 2022-12-21
TFDSv4 carla_video_tracking upgrade.

jprokos26:
Throws an error when run without flags but completes successfully with --skip-attack. See https://github.com/twosixlabs/armory/issues/1814 for more information.

Both `dev` and `test` datasets uploaded to s3 and ready for review.

Note, duplicate of https://github.com/twosixlabs/armory/pull/1829 to properly apply rebase.


armory/datasets/preprocessing.py
```diff
@@ -57,6 +58,28 @@ def carla_over_obj_det_dev(element, modality="rgb"):
...
+
+@register
+def carla_video_tracking_test(element, max_frames=None):
+    return carla_video_tracking(
+        element["video"], max_frames=max_frames, split="test"
+    ), carla_video_tracking_labels(
+        element["video"],
+        (element["bboxes"], element["patch_metadata"]),
+        max_frames=max_frames,
+    )
```
jprokos26, line 80: I am assuming `max_frames` gets passed in as a `dataset_kwargs` but this is untested as I cannot find an example of it being used previously. Planning to circle back and test over the next couple days.

jprokos26 (commented):
Reapplying review comments

armory/datasets/generator.py
```diff
@@ -164,3 +167,13 @@ def wrap_generator(armory_data_generator):
...
+
+
+class EvalGenerator:
+    """
+    Wraps a specified number of batches in a DataGenerator to allow for evaluating on
+    part of a dataset when running through a scenario
+    """
+
+    def __init__(self, armory_generator, num_eval_batches):
+        raise NotImplementedError("EvalGenerator not implemented")
```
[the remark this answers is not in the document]
lcadalzo: removed

scenario_configs/eval5/carla_video_tracking/carla_video_tracking_goturn_advtextures_defended.json
```diff
@@ -14,11 +14,11 @@
         "use_label": true
     },
     "dataset": {
-        "batch_size": 1,
-        "eval_split": "dev",
-        "framework": "numpy",
-        "module": "armory.data.adversarial_datasets",
-        "name": "carla_video_tracking_dev"
+        "test": {
```
lcadalzo, line 17: did you mean to change this from `dev` to `test`?

docs/metrics.md
```diff
@@ -73,6 +73,10 @@ The `armory.metrics` module contains functionality to measure a variety of metri
 We have implemented the metrics in numpy, instead of using framework-specific metrics, to prevent expanding the required set of dependencies.
 Please see the relevant submodules in [armory/metrics](../armory/metrics/) for more detailed descriptions.
 
+### Custom Metrics
```
lcadalzo, line 76: we have a custom metrics section [here](https://github.com/twosixlabs/armory/blob/53708efa13f9582f317a4358215cd7a368c8c874/docs/metrics.md#custom-metrics-1). As far as I can tell, this TODO should be deleted

scenario_configs/eval5/carla_video_tracking/carla_video_tracking_goturn_advtextures_defended.json
jprokos26: No this should be `dev`, I changed it to `test` for testing. Will fix in my next push.

docs/metrics.md
jprokos26: You are right, I'm not sure why my rebase didn't apply properly, I am going through to see where it diverged now.

armory/scenarios/scenario.py
```diff
@@ -355,6 +355,11 @@ def evaluate_all(self):
     def next(self):
```
lcadalzo, line 355: I think you need to be updating `next()` in the CarlaVideoTracking class, not the base Scenario. The former already overrides `next()` and separates the `y` from `y_patch_metadata` in the 2-tuple [here](https://github.com/twosixlabs/armory/blob/master/armory/scenarios/carla_video_tracking.py#L29).

armory/datasets/generator.py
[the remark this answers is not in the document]
jprokos26:
```suggestion
```

armory/scenarios/scenario.py
lcadalzo: also the conversion from `Dict` to `List[Dict]` I'd think should be as simple as something like

```
if isinstance(y, dict)`:
    y = [y]
```

I assume there should be a cleaner way than need to reference specific keys in the base scenario class

armory/datasets/preprocessing.py
```diff
@@ -145,6 +168,145 @@ def carla_over_obj_det_image(x, modality="rgb"):
...
+        if batch.dtype == object:
+            raise NotImplementedError(
+                "<object> dtype not yet supported for variable image processing."
+            )
+        raise ValueError(
+            f"input dtype {batch.dtype} not in ({context.input_type}, 'O')"
+        )
+    check_shapes(tuple(batch.shape), context.x_shape)
+    assert batch.dtype == context.input_type
+    return batch
```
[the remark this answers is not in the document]
lcadalzo: we should be removing as much as possible from the old preprocessing. I don't believe `preprocessing_chain`, `label_preprocessing_chain`, etc. should be ported over. I believe the only function we should be adding to `preprocessing.py` is the registered `carla_video_tracking` preprocessor

armory/datasets/preprocessing.py
```diff
@@ -145,6 +168,145 @@ def carla_over_obj_det_image(x, modality="rgb"):
...
+        return boxes[:, : self.max_frames, :]
+
+    def clip_metadata(self, patch_metadata_dict):
+        return {
+            k: v[:, : self.max_frames, ::] for (k, v) in patch_metadata_dict.items()
+        }
+
+    def __call__(self, x, labels):
+        boxes, patch_metadata_dict = labels
+        return self.clip_boxes(boxes), self.clip_metadata(patch_metadata_dict)
```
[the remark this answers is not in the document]
lcadalzo: I believe they should be removed, and have the `max_frames` kwarg of the dataset function do the indexing

armory/scenarios/scenario.py
jprokos26: To your former comment, yes I can implement it within `CarlaVideoTracking` but I wasn't sure if this should be a fix for other dataloaders, I will move it to this function.

Regarding conversion from Dict the problem is that the tensors within the dictionary have an extra dimension to represent batches (batch, frames, boxes). I am iterating through that first dimension with the for loop and sending each "batch" to its own dictionary with two dimensions (frames, boxes) as each value in the resultant list. I am assuming I should still try to support multi-batch correct?

armory/datasets/preprocessing.py
lcadalzo: it should be passed as a `preprocessor_kwarg`. See the configs in #1788 as an example

armory/datasets/context.py
```diff
@@ -0,0 +1,34 @@
+import tensorflow as tf
+
+
+class ImageContext:
```
lcadalzo, line 4: I don't believe we'd actually be using any of the info in contexts except for `frame_rate`. We used to use the context objects for preprocessing but we aren't in the new setup. 

To my understanding, the issue here is that exporters need the dataset frame rate, and there's no way to access that from the dataset object. Is that correct? If that's so, we should make this piece as minimal as possible

jprokos26: Yes you are correct that is the issue; I can reduce this down to be solely VideoContext containing frame-rate; I was unsure if other scenarios would require this context but I can selectively re-add as necessary.

lcadalzo: I think audio has sample_rate which is needed by the audio exporter for the same reason, see #1803

armory/scenarios/scenario.py
lcadalzo: inevitably, I think, we need a way to handle this extra batch dimension that the new tf preprocessing adds. Note with the CARLA OD PR from a while earlier, I [punted](https://github.com/twosixlabs/armory/pull/1788#issue-1470111022) on this question for now. Wherever we end up adding the logic to handle it should be as specific to the issue as possible (i.e. would rather not add to a base `scenario.py` that every Armory run touches). Ideally it would be added to the carla preprocessing code but I'm not sure how to get that to work. If we can find a way to do it there, that'd be best. Offhand, the other places I can think of would be whichever inherited `Scenario` class is most appropriate. And also in metrics code, but that changes our API as far as expected format

armory/datasets/context.py
jprokos26: Addressed by 0bdb3ec3ce074b5e33d590ea83ef7e0fd29da408. AudioContext can be added in a separate PR.

armory/datasets/generator.py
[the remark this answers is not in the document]
jprokos26: Addressed in e198f8134009f420a7ed565181acec459a343bb3.

armory/datasets/preprocessing.py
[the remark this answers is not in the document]
jprokos26: Addressed by d9c0a2234a46607689f5bf7a677a3cba707c722b.

armory/datasets/preprocessing.py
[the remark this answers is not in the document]
jprokos26: Addressed by d9c0a2234a46607689f5bf7a677a3cba707c722b.

scenario_configs/eval5/carla_video_tracking/carla_video_tracking_goturn_advtextures_defended.json
jprokos26: Addressed by d9862ce7a7409946eb90f9a513a6df4150871adc.

docs/metrics.md
jprokos26: Addressed in f07ae034e7a4415bf0bfeb3c543133bdf4e7b194.

armory/datasets/preprocessing.py
```diff
@@ -57,6 +58,54 @@ def carla_over_obj_det_dev(element, modality="rgb"):
     )
 
 
+def carla_video_tracking_preprocess(element, max_frames, split):
```
lcadalzo, line 61: I'd suggest separating the video preprocessing and the label preprocessing into separate functions to make things more modular

armory/datasets/preprocessing.py
```diff
@@ -57,6 +58,54 @@ def carla_over_obj_det_dev(element, modality="rgb"):
...
+    x = element["video"]  # tf.Tensor [B, H, W, C]
+    y = element["bboxes"]  # tf.Tensor [B, 4]
+    y_meta = element["patch_metadata"]  # Dict
```
lcadalzo, line 64: for consistency with the rest of the codebase, I'd suggest renaming `y_meta` to `y_patch_metadata`

armory/datasets/context.py
```diff
@@ -0,0 +1,16 @@
+import tensorflow as tf
+
+
+class VideoContext:
+    def __init__(self, x_shape, frame_rate):
+        self.x_shape = x_shape
```
lcadalzo, line 6: do we expect `x_shape` and `input_type` to ever actually be used? If not, probably clearer to remove so in the future we're not wondering where/if it's used

jprokos26: Yes, they are both used in the preprocessor here: https://github.com/twosixlabs/armory/blob/f07ae034e7a4415bf0bfeb3c543133bdf4e7b194/armory/datasets/preprocessing.py#L84-L89

armory/datasets/preprocessing.py
jprokos26: Addressed in 9716c588a48d001b1afab6bbe06cb351c0fbe710. Tested working with `max_frames` 2 and 10. Throws error due to dimension mismatch when `max_frames=1`.

armory/datasets/preprocessing.py
jprokos26: Addressed in 8db643f627562c3b68265725f1a7bee85d06ada6.

armory/datasets/preprocessing.py
```diff
@@ -57,6 +58,54 @@ def carla_over_obj_det_dev(element, modality="rgb"):
...
+
+
+@register
+def carla_video_tracking_dev(element, max_frames=None):
+    return carla_video_tracking_preprocess(element, max_frames, "dev")
+
+
+@register
+def carla_video_tracking_test(element, max_frames=None):
+    return carla_video_tracking_preprocess(element, max_frames, "test")
```
jprokos26, line 107:
```suggestion
def carla_video_tracking_preprocess(x, max_frames, context):
    # Clip
    if max_frames:
        max_frames = int(max_frames)
        if max_frames <= 0:
            raise ValueError(f"max_frames {max_frames} must be > 0")
        x = x[:max_frames, :]
    # Validate input
    if x.dtype != context.input_type:
        if x.dtype == object:
            raise NotImplementedError(
                "<object> dtype not yet supported for variable image processing."
            )
        raise ValueError(f"input dtype {x.dtype} not in ({context.input_type}, 'O')")
    assert isinstance(
        context.x_shape, tuple
    ), f"target shape {context.x_shape} is not a tuple"
    assert len(x.shape) == len(context.x_shape)
    for a, t in zip(x.shape, context.x_shape):
        assert (
            t is None or a == t
        ), f"shape {x.shape} does not match shape {context.x_shape}"
    assert x.dtype == context.input_type
    return x


def carla_video_tracking_preprocess_labels(y, y_patch_metadata, max_frames):
    # Clip
    if max_frames:
        max_frames = int(max_frames)
        if max_frames <= 0:
            raise ValueError(f"max_frames {max_frames} must be > 0")
        y = y[:max_frames, :]
        y_patch_metadata = {k: v[:max_frames, :] for (k, v) in y_patch_metadata.items()}
    # Update labels
    box_array = tf.squeeze(y, axis=0) if y.shape[0] == 1 else y
    y = {"boxes": box_array}
    y_patch_metadata = {
        k: (tf.squeeze(v, axis=0) if v.shape[0] == 1 else v)
        for k, v in y_patch_metadata.items()
    }
    return y, y_patch_metadata


@register
def carla_video_tracking_dev(element, max_frames=None):
    return carla_video_tracking_preprocess(
        element["video"],
        max_frames=max_frames,
        context=contexts["carla_video_tracking_dev"],
    ), carla_video_tracking_preprocess_labels(
        element["bboxes"], element["patch_metadata"], max_frames=max_frames
    )


@register
def carla_video_tracking_test(element, max_frames=None):
    return carla_video_tracking_preprocess(
        element["video"],
        max_frames=max_frames,
        context=contexts["carla_video_tracking_test"],
    ), carla_video_tracking_preprocess_labels(
        element["bboxes"], element["patch_metadata"], max_frames=max_frames
    )
```

jprokos26: Addressing https://github.com/twosixlabs/armory/pull/1836/files/f07ae034e7a4415bf0bfeb3c543133bdf4e7b194#r1055888776

armory/datasets/preprocessing.py
jprokos26: See https://github.com/twosixlabs/armory/pull/1836/files/f07ae034e7a4415bf0bfeb3c543133bdf4e7b194#r1055907094.

armory/datasets/preprocessing.py
jprokos26: See https://github.com/jprokos26/armory/commit/612fe1e7dea3880bce13e3832c592fbe5fe3df17 for a full commit.

armory/scenarios/scenario.py
jprokos26: Unfortunately there doesn't seem to be a way to implement this within the preprocessor code as the data is not batched when it passes through there; I have moved this fixup to `CarlaVideoTracking`, see https://github.com/jprokos26/armory/commit/4837db0e3da65322ebc3624688d9b1140222bacf.

Additionally to avoid referencing a specific key, it can be changed to:
```python
    def next(self):
        super().next()
        self.y, self.y_patch_metadata = self.y
        # Fix for tfdsv4 upgrade - separate batches into list
        if isinstance(self.y, dict):
            assert (
                len(self.y.keys()) == 1
            ), f"Expected only one key in labels dict, got {self.y.keys()}"
            key = next(iter(self.y))
            self.y = [{key: batch} for batch in self.y[key]]
        self.probe.update(y=self.y, y_patch_metadata=self.y_patch_metadata)
```
Which do you think?

armory/datasets/context.py
lcadalzo: I think we should make preprocessing as simple as possible by not feeding in the context object (and removing the corresponding assertions). I'd think the context should only be needed for grabbing frame_rate when the exporter is instantiated

armory/datasets/preprocessing.py
```diff
@@ -57,6 +58,55 @@ def carla_over_obj_det_dev(element, modality="rgb"):
     )
 
 
+def carla_video_tracking_preprocess(element, max_frames, split):
```
lcadalzo, line 61: I'd suggest having `max_frames` be a kwarg that defaults to None

armory/datasets/preprocessing.py
jprokos26: Accepted with https://github.com/twosixlabs/armory/pull/1836/commits/612fe1e7dea3880bce13e3832c592fbe5fe3df17.

armory/scenarios/scenario.py
jprokos26: Addressed in 4837db0e3da65322ebc3624688d9b1140222bacf. @lcadalzo let me know if I should use the code snippet above instead to avoid referencing `"boxes"`.

armory/datasets/preprocessing.py
jprokos26: It defaults to none in the registered functions:
https://github.com/twosixlabs/armory/blob/742d6cb0fcf3fcdc00242172668ad5652c81486c/armory/datasets/preprocessing.py#L106
https://github.com/twosixlabs/armory/blob/742d6cb0fcf3fcdc00242172668ad5652c81486c/armory/datasets/preprocessing.py#L117

lcadalzo: I'd have it also be a kwarg that defaults to None in the functions the registered one is calling, this way you're not forced to pass max_frames if you're using one of those functions directly and not via the registered one

armory/datasets/context.py
jprokos26: Addressed in 18ce134d72cb1318a7a3489b827a98e687945530. Without the input validation should the preprocessor function still be split into two functions to handle labels separately or can I merge these now?

armory/datasets/preprocessing.py
jprokos26: Addressed in 9f71947cee22b7f5c941312dabe21acf37418ebc.

jprokos26:
@lcadalzo Are you alright with the changes in 4a29dba48a1228abf409ed5235f6a4d1e915bcad? I felt it unnecessary to split the preprocessor up by data & label due to https://github.com/twosixlabs/armory/pull/1836#discussion_r1056621894.

lcadalzo:
I would still keep `x` preprocessing separate from label preprocessing since it's more modular this way

jprokos26:
> I would still keep `x` preprocessing separate from label preprocessing since it's more modular/reusable this way

Reverted in 7c2da860db01e3cf717354d3bb9b9d10fa94f9e1.

armory/datasets/preprocessing.py
```diff
@@ -57,6 +57,40 @@ def carla_over_obj_det_dev(element, modality="rgb"):
...
+        element["video"],
+        element["bboxes"],
+        element["patch_metadata"],
+        max_frames=max_frames,
+    )
+
+
+@register
+def carla_video_tracking_test(element, max_frames=None):
+    return carla_video_tracking_dev(element, max_frames)
```
lcadalzo: minor refactor suggestion: if `_test`'s preprocessor is just calling `dev`'s, this suggests we can rename `carla_video_tracking_dev` to just `carla_video_tracking`. Then as we do for mnist/resisc/etc:

```
carla_video_tracking_dev = register(carla_video_tracking, "carla_video_tracking_dev")
carla_video_tracking_test = register(carla_video_tracking, "carla_video_tracking_test")
```

and this is a bit simpler by not needing to define functions for both `_dev` and `_test`

jprokos26: Addressed in 3ecba7229fafad9140058585b044831932b06122

armory/datasets/preprocessing.py
```diff
@@ -58,6 +58,47 @@ def carla_over_obj_det_dev(element, modality="rgb"):
     )
 
 
+def carla_video_tracking_preprocess(x, max_frames=None):
```
lcadalzo, line 61: input preprocessing also needs to be scaling from [0, 255] to [0, 1]

lcadalzo:
can we keep the old files as-is in `armory/data`? We haven't started removing any code from that directory yet, so I'd rather do it at once at a later time

armory/datasets/preprocessing.py
```diff
@@ -58,6 +58,47 @@ def carla_over_obj_det_dev(element, modality="rgb"):
...
+def carla_video_tracking_preprocess_labels(y, y_patch_metadata, max_frames=None):
+    # Clip
+    if max_frames:
+        max_frames = int(max_frames)
+        if max_frames <= 0:
+            raise ValueError(f"max_frames {max_frames} must be > 0")
+        y = y[:max_frames, :]
+        y_patch_metadata = {k: v[:max_frames, :] for (k, v) in y_patch_metadata.items()}
+    # Update labels
+    box_array = tf.squeeze(y, axis=0) if y.shape[0] == 1 else y
```
lcadalzo, line 80: are we ever expecting the squeeze condition `y.shape[0] == 1` to be True? To my knowledge, no, and if this is the case we should remove these lines as it only would then be adding potential confusion

armory/scenarios/carla_video_tracking.py
```diff
@@ -27,6 +27,9 @@ def load_test_dataset(self):
     def next(self):
         super().next()
         self.y, self.y_patch_metadata = self.y
+        # Fix for tfdsv4 upgrade - separate batches into list
+        if isinstance(self.y, dict) and "boxes" in self.y:
+            self.y = [{"boxes": batch} for batch in self.y["boxes"]]
```
lcadalzo, line 32: r.e. the issue of tf preprocessing outputting the extra batch dimension: here we take a half-step towards addressing it, by removing the batch dim for `self.y`. We don't touch anything in `self.y_patch_metadata` yet, so the attack will still error out

I see a few options for this PR:
1. Address it for both `y` and `y_patch_metadata` by further modifying scenario code to convert from `(1, N, 4)` to `(N, 4)` within `y_patch_metadata` too
2. Don't address it for either yet (i.e. just do `self.y = [self.y]` in the `if` block)
3. Address it for both `y` and `y_patch_metadata` but instead of modifying scenario code to remove the dimension, we could modify metrics to accept either `(N, 4)` or `(1, N, 4)`. We actually already do this for mAP, see [here](https://github.com/twosixlabs/armory/blob/master/armory/metrics/task.py#L651). For this particular model, we'd also need to overwrite [this](https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/art/estimators/object_tracking/pytorch_goturn.py#L625) `predict()` method to remove the batch dimension in `y_init`

The pros/cons of (3) vs (1) are what we need to weigh. For context, this batch dimension issue affects all tracking and detection scenarios, and I don't think we should do a mix of (3) and (1). 

(1) may be the quickest path to get things running, but it feels unusual or unexpected to have scenario code doing further dataset preprocessing. It also may have unexpected side effects in the future when users are subclassing the scenario. If we go (3), though, we're telling users "your model + attack must be compatible with this slightly awkward `(1, N, 4)` format"

Thoughts? There may be a possible option (4) where we apply preprocessing not as a tf map operation but instead after the tf dataset is created (enabling us to do preprocessing as we used to), although this would require some rearchitecting

armory/datasets/preprocessing.py
jprokos26: Addressed with e466546acf85008b8edf5815816c4822ee733764

armory/datasets/preprocessing.py
jprokos26: Addressed in a82f34fcfaf1c03e55a14257fa98452790a8ccda

armory/scenarios/carla_video_tracking.py
lcadalzo: @jprokos26 revisiting this, I think (1) is less bad than (3), bc it at least maintains an interface of sensible shape `(N, 4)` over the awkward `(1, N, 4)`. If we went (3), to reiterate, we're forcing anyone creating an attack or metric to work with the awkward shape.

Happy to discuss this more, but if you agree please update `next()` in `carla_video_tracking.py` to also remove the batch dimension for `y_patch_metadata`

jprokos26: Addressed with 4dcd2ca9ff5fd42e1415355a564e10ffa75bb073

armory/scenarios/carla_video_tracking.py
```diff
@@ -24,9 +24,18 @@ def load_test_dataset(self):
             raise ValueError("batch_size must be 1 for evaluation.")
         super().load_test_dataset(test_split_default="dev")
 
+    def _split_batches_into_list(self, x: dict) -> list:
+        """Fix for tfdsv4 upgrade - separate batches into list"""
+        if not isinstance(x, dict):
+            return x
+        expected_batches = list(x.values())[0].shape[0]
+        if not all(isinstance(x[k], np.ndarray) and x[k].shape[0] == expected_batches for k in x):
+            raise ValueError(f"Expected all batches to have the same length, but got {x}")
```
jprokos26, line 33:
```suggestion
            raise ValueError(f"Expected all values to have the same length (batches), but got {x}")
```

armory/datasets/generator.py
```diff
@@ -129,6 +127,9 @@ def __init__(
             # SEE: tfds.as_numpy
             # https://github.com/tensorflow/datasets/blob/v4.7.0/tensorflow_datasets/core/dataset_utils.py#L141-L176
             raise NotImplementedError(f"framework {framework}")
+        
+        if info.metadata is None:
+            info.metadata = tfds.core.MetadataDict()
```
lcadalzo, line 132: this line yields an error when the conditional is met. Try testing with e.g. cifar dataset:

```

 File "/workspace/armory/datasets/generator.py", line 132, in __init__
    if info.metadata is None:
       │    └ <property object at 0x7fabca4d7270>
       └ tfds.core.DatasetInfo(
             name='cifar10',
             full_name='cifar10/3.0.2',
             description="""
             The CIFAR-10 dataset consis...

AttributeError: can't set attribute

```

lcadalzo:
LGTM

PR #1838, opened 2022-12-22
replace carla-mot and pytorch image names to "armory" in scenario_configs

### LINKED ISSUE #1834: Update eval6 configs docker images
The configs in `scenario_configs/eval6` use `pytorch` and `carla-mot` docker images, which have been consolidated/removed on `develop`. I would think that before a 0.16.2 release we need to update these configs to no longer point to deprecated images, can you confirm @mwartell?

mwartell:
Fixes #1834

Because the `twosixarmory/carla-mot` and `twosixarmory/pytorch` have been merged into `twosixarmory/armory`, replace the image names in the `scenario_configs/eval6". 

Older scenario configs (for example `scenario_configs/eval5/*` or `scenario_configs/mnist_baseline.json` have been changed in separate commits f1e1957 and b3743a8 in case this was not wanted by #1834


lcadalzo:
looks like [these](https://github.com/mwartell/armory-1/tree/mwartell/isue1834-remove-dead-flavors/tests/scenarios/broken) configs could use an update too?

lcadalzo:
[these](https://github.com/mwartell/armory-1/tree/mwartell/isue1834-remove-dead-flavors/tests/scenarios/tf1) configs should either be deleted if we're no longer supporting tf1, or renamed to `"armory"`

lcadalzo:
while we're looking at this, let's update the very outdated example config in [this](https://github.com/mwartell/armory-1/blob/mwartell/isue1834-remove-dead-flavors/docs/configuration_files.md#configuration-files) file

mwartell:
> looks like [these](https://github.com/mwartell/armory-1/tree/mwartell/isue1834-remove-dead-flavors/tests/scenarios/broken) configs could use an update too?

I believe those were put in `broken` because they didn't work. I have simply removed them because they'd been effectively deprecated in v0.15.x

lcadalzo:
any reason you deleted the `broken` configs rather than updating the image name?

lcadalzo:
by deprecated you mean our CI is no longer using them? In theory it seems like what those are testing (should get an error if your config isn't set up right) is still relevant

lcadalzo:
> by deprecated you mean our CI is no longer using them?

if I'm interpreting the CI failures here correctly, it is still expecting those files to exist

lcadalzo:
> Yep, the broken scenarios were for checking expected failures in CI. I have restored them.

but their docker images should be changed from `tf1` to `armory`, correct?

mwartell:
> but their docker images should be changed from `tf1` to `armory`, correct?

Those files are there to test config loading and validation, so the image name doesn't actually matter. I've changed them to "armory" for consistency, though.

lcadalzo (approved):
LGTM

PR #1840, opened 2022-12-22
offload scenario-runner to self-hosted

mwartell:
The scenario-runner test exhausts resources on Github-hosted action runners. This causes them to use our self hosted runner.


lcadalzo:
can we test this by launching a `Scenario Runner` action on the `self-hosted-runner` branch and seeing if all tests pass?

lcadalzo:
is this intended to go on a different branch?

mwartell:
Supplanted by #1844

PR #1844, opened 2022-12-27
runs scenario checks on self-hosted runner

mwartell:
This repoints the scenario-runner CI to execute on our self-hosted runner because the Github hosted runners are too small for those tests.


ppark-twosixtech (approved):
Approving to test

PR #1846, opened 2022-12-28
ucf101

ppark-twosixtech:


ppark-twosixtech:
Some observations regarding `frame_rate=self.test_dataset.context.frame_rate` for `ucf101` and #1803 in general:
- `frame_rate` and `sample_rate` only used for exporting media (`VideoClassificationExporter`, `VideoTrackingExporter`, `AudioExporter`)
- `ImageContext`, `VideoContext`, `AudioContext` only seem to be used as a method for storing information like `frame_rate` and `sample_rate` for different datasets that only get used at export

Note that `frame_rate` and `sample_rate` are not exposed to the user, and they seem to be just hardcoded values for export. If we really need to reference these values in some meaningful way, recommend a mechanism other than using a `Context` class. If it's only for export, would recommend just hardcoding the values for each scenario when the `Exporter` class is initialized e.g:
```
class Ucf101(Scenario):
    ...
    def _load_sample_exporter(self):
        return VideoClassificationExporter(
            self.export_dir,
            # frame_rate=self.test_dataset.context.frame_rate,
            frame_rate=10,
        )
```

lcadalzo:
Looks like we'll need to merge the latest from `tfdsv4` branch into this one and perhaps clean up `package.py`. I imagine once we do so the diff here should be pretty minor

ppark-twosixtech:
because we don't build `ucf101` ourselves (`ucf101` is a standard `tfds` dataset), `frame_rate` will not be available as metadata, so we rely on `DEFAULT_FRAME_RATE` in `video_ucf101_scenario.py`

jprokos26 (approved):
Tested working with `ucf101_baseline_finetune.json` and `carla_video_tracking_goturn_advtextures_defended.json`

PR #1848, opened 2022-12-30
audio poisoning baseline results

swsuggs:
Results for speech commands poisoning with all 4 triggers.

Still running Sleeper Agent.


swsuggs:
The sleeper agent runs have finished but there is something fishy about the numbers.  I will look into it, but this can be merged now for the upcoming release

mwartell (approved):
Documentation changes accepted.

PR #1849, opened 2023-01-02
WIP: german_traffic_sign

ppark-twosixtech:


ppark-twosixtech:
Requires updating data loading for poisoning scenarios (where `german_traffic_sign` is referenced most) for further testing after loading

PR #1850, opened 2023-01-03
coco

ppark-twosixtech:
Including all configs for `coco`:
- `2014`
- `2017`
- `2017_panoptic`


lcadalzo:
@ppark-twosixtech can you merge tfdsv4 into this branch? I believe the only issue blocking this dataset was 1802, which is resolved, so hopefully this one's good to go after that

lcadalzo:
@ppark-twosixtech please request review once this is in a place where things are working

ppark-twosixtech:
Like PRs #1872, #1873, #1874, running `evaluate` in Python with attack throws an error about an extra dimension. `check_run` will pass with `skip_attack` set to `True` (used config file for `xview` as an example).

lcadalzo:
I'm able to run a scenario with `--skip-attack` without error. I was using the xView config (and model) and swapped in the coco dataset.

I noticed a mAP of 0 over 10 batches, and as a quick check I turned off the `tf.lookup()` that's re-mapping the labels. After doing so, the mAP was still 0, fwiw.

Were you able to get the Armory coco baseline model running? If so, it'd be worthwhile to run over this dataset and confirm a non-zero mAP. Overall the code here LGTM though

Issue #1858, opened 2023-01-11
[BUG]: Depth images are saved incorrectly


dxoigmn:
### Description of the bug

It is possible for depth images to have non-integer values when passed to:
https://github.com/twosixlabs/armory/blob/066db8467a425403b980c5d408e0d6dbdca5c7d1/armory/instrument/export.py#L68
For example, the value of the depth image pixel (in RGB) could be `[55.255993 61.999996  9.999999]`. In the ideal case these would be rounded to the nearest integer, however, `get_sample` always rounds down:
https://github.com/twosixlabs/armory/blob/066db8467a425403b980c5d408e0d6dbdca5c7d1/armory/instrument/export.py#L100
As a result of rounding down, this causes noise in the saved depth images on disk and can cause large perturbations as a result.

### Steps To Reproduce

1. Examine `255*x_i_mode` just before this line:
https://github.com/twosixlabs/armory/blob/066db8467a425403b980c5d408e0d6dbdca5c7d1/armory/instrument/export.py#L100
3. In the case of depth images, the pixel values around the patch will be non-integer (after multiplying by 255).

lcadalzo:
impending 0.16.3 patch release will contain this fix

PR #1860, opened 2023-01-12
round to nearest int when exporting, rather than rounding down

lcadalzo:
Addresses #1858

### DIFF
--- a/armory/instrument/export.py
+++ b/armory/instrument/export.py
@@ -98,7 +98,7 @@ class ImageClassificationExporter(SampleExporter):
         else:
             raise ValueError(f"Expected 1, 3, or 6 channels, found {x.shape[-1]}")
         image = Image.fromarray(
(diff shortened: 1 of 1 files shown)


lcadalzo:
@dxoigmn thanks for reporting this issue. If you're able to, please let us know if this PR mitigates the issue by rounding to nearest int rather than down

dxoigmn:
I believe round to nearest integer mitigates the issue.

mwartell (approved):
This is a straight-forward conversion of returns `astype(int)`

mwartell:
> @dxoigmn @lcadalzo
> 
> You need to convert to `np.uint8` instead of `int`. `PIL` will not produce the expected images if you give `int`.

I didn't account for this in my review. I've tagged myself to look at this again when this comment is considered.

lcadalzo:
thanks @mzweilin for pointing this out. I've also now confirmed this is the case and have added a commit to address it

mwartell (approved):
Properly changes return types to `np.uint8` as intended.

PR #1863, opened 2023-01-13
Sleeper agent fix

swsuggs:
Corrections: scale the patch to [0,1], and find the correct poison indices.

Note: running the current configs will return low attack success because the attack parameters are tuned for a slightly different resnet version.

### DIFF
--- a/armory/scenarios/poisoning_sleeper_agent.py
+++ b/armory/scenarios/poisoning_sleeper_agent.py
@@ -167,4 +167,19 @@ class SleeperAgentScenario(Poison):
                 np.array([]),
             )
 
+        # poison_index returned by attack is the index within the target class, not the whole dataset.
+        # In addition, it may contain images that aren't actually perturbed.
+        # Find the true poison indices
+        true_poison_inds = np.array(
+            [
+                i
+                for i in range(len(self.x_clean))
+                if (self.x_clean[i] != self.x_poison[i]).all()
+            ]
+        )
+        n_target = (self.y_clean == self.target_class).sum()
+        log.info(
+            f"Actual amount of poison returned by attack: {len(true_poison_inds)} samples or {len(true_poison_inds)/n_target} percent"
+        )
+        self.poison_index = true_poison_inds
         self.record_poison_and_data_info()
--- a/armory/scenarios/poisoning_sleeper_agent.py
+++ b/armory/scenarios/poisoning_sleeper_agent.py
@@ -151,15 +151,27 @@ class SleeperAgentScenario(Poison):
     def poison_dataset(self):
         self.hub.set_context(stage="poison")
         if self.use_poison:
-            (
-                self.x_poison,
-                self.y_poison,
-                poison_index,
-            ) = self.poisoner.poison_dataset(
+            self.x_poison, self.y_poison = self.poisoner.poison_dataset(
                 self.x_clean,
                 self.label_function(self.y_clean),
-                return_index=True,
+                return_index=False,
             )
+
+            # Manually find the poison indices.  Although the attack can return them, they
+            # will be the index within the target class, not the whole dataset.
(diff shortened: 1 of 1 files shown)


lcadalzo:
@swsuggs is there a good reason not to modify the attack parameters in the default configs to whatever is most appropriate?

swsuggs:
Finding them could take days of testing, and performers (namely MIT) are interested in getting the corrections before the submission deadline

armory/scenarios/poisoning_sleeper_agent.py
```diff
@@ -161,4 +167,19 @@ def poison_dataset(self):
...
+                for i in range(len(self.x_clean))
+                if (self.x_clean[i] != self.x_poison[i]).all()
+            ]
+        )
+        n_target = (self.y_clean == self.target_class).sum()
+        log.info(
+            f"Actual amount of poison returned by attack: {len(true_poison_inds)} samples or {len(true_poison_inds)/n_target} percent"
+        )
+        self.poison_index = true_poison_inds
         self.record_poison_and_data_info()
```
lcadalzo, line 185: does the updated `self.poison_index` only matter for the `record_poison_and_data_info()` call?

armory/scenarios/poisoning_sleeper_agent.py
```diff
@@ -161,4 +167,19 @@ def poison_dataset(self):
...
+                i
+                for i in range(len(self.x_clean))
+                if (self.x_clean[i] != self.x_poison[i]).all()
+            ]
+        )
+        n_target = (self.y_clean == self.target_class).sum()
+        log.info(
+            f"Actual amount of poison returned by attack: {len(true_poison_inds)} samples or {len(true_poison_inds)/n_target} percent"
+        )
+        self.poison_index = true_poison_inds
```
lcadalzo, line 184: earlier in the function should we be setting `self.poison_index` if it's overriden immediately with `true_poison_inds`?

swsuggs: it is not necessary to set self.poison_index earlier.  I can update it.

armory/scenarios/poisoning_sleeper_agent.py
swsuggs: I believe the fairness metrics use it as well.  It looks like ```record_poison_and_data_info``` creates a new variable ```self.poisoned``` which is essentially the same as self.poison_index.  Simplifying this is desirable but not necessarily in the scope of this PR

armory/scenarios/poisoning_sleeper_agent.py
```diff
@@ -154,14 +154,14 @@ def poison_dataset(self):
             (
                 self.x_poison,
                 self.y_poison,
-                self.poison_index,
+                poison_index,
```
lcadalzo, line 157: if you set `self.x_poison, self.y_poison, _` instead it will be clearer that the third output isn't used

swsuggs: Didn't see this before making the next commit.  See if you are fine with that, otherwise I can use _

swsuggs: Didn't see this before making the next commit.  See if you are fine with that, otherwise I can use _

lcadalzo: I'd use `_` so it's clear that the third output isn't used

lcadalzo: to clarify, I'd still keep the real variable we care about defined as `poison_index` rather than `true_poison_index`

swsuggs: check the latest

lcadalzo (approved):

PR #1869, opened 2023-01-25
Speech commands explanatory model

swsuggs:
Adds explanatory model for poisoning fairness metrics.

Also corrects for an ART bug where triggered audio samples go out of range.

### DIFF
--- a/armory/art_experimental/attacks/poison_loader_audio.py
+++ b/armory/art_experimental/attacks/poison_loader_audio.py
@@ -60,6 +60,7 @@ class CacheTrigger:
             raise ValueError("Shift + Backdoor length is greater than audio's length.")
 
         audio[shift : shift + bd_length] += self.scaled_trigger
+        audio = np.clip(audio, -1.0, 1.0)
         return audio.astype(original_dtype)
 
 
--- a/armory/metrics/poisoning.py
+++ b/armory/metrics/poisoning.py
@@ -16,7 +16,6 @@ EXPLANATORY_MODEL_CONFIGS = explanatory_model_configs = {
         "model_framework": "tensorflow",
         "module": "armory.baseline_models.tf_graph.audio_resnet50",
         "name": "get_unwrapped_model",
-        "preprocess_kwargs": {},
         "weights_file": "speech_commands_explanatory_model_resnet50_bean.h5",
     },
     "cifar10_explanatory_model": {
@@ -27,14 +26,11 @@ EXPLANATORY_MODEL_CONFIGS = explanatory_model_configs = {
         },
         "module": "armory.baseline_models.pytorch.resnet18_bean_regularization",
         "name": "get_model",
-        "preprocess_kwargs": {},
         "weights_file": "cifar10_explanatory_model_resnet18_bean.pt",
     },
     "gtsrb_explanatory_model": {
-        "model_kwargs": {},
         "module": "armory.baseline_models.pytorch.micronnet_gtsrb_bean_regularization",
         "name": "get_model",
-        "preprocess_kwargs": {},
         "weights_file": "gtsrb_explanatory_model_micronnet_bean.pt",
     },
     "resisc10_explanatory_model": {
@@ -62,7 +58,7 @@ class ExplanatoryModel:
         data_modality="image",
         model_framework="pytorch",
         activation_layer=None,
-        preprocess_kwargs={},
+        preprocess_kwargs=None,
     ):
         """
         explanatory_model: A callable pytorch or tensorflow model used to produce
@@ -84,7 +80,7 @@ class ExplanatoryModel:
         self.data_modality = data_modality
         self.model_framework = model_framework
         self.activation_layer = activation_layer
-        self.preprocess_kwargs = preprocess_kwargs
+        self.preprocess_kwargs = preprocess_kwargs if preprocess_kwargs else {}
 
         if self.activation_layer is not None:
             if self.model_framework == "tensorflow":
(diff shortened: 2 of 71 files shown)


swsuggs:
Most of the non-trivial changes here were needed because the code for the fairness metrics assumed that the explanatory model was a pytorch model, but the speech commands one is tensorflow.  It also assumed image data, while speech commands is audio.

This should work for eval6.  It makes no pretense to general applicability, meaning It is possible that future model architectures might require additional changes; but that's hard to predict.

swsuggs:
@lcadalzo could you review when you get a chance

armory/art_experimental/attacks/poison_loader_audio.py
```diff
@@ -60,6 +60,7 @@ def insert(self, x: np.ndarray) -> np.ndarray:
             raise ValueError("Shift + Backdoor length is greater than audio's length.")
 
         audio[shift : shift + bd_length] += self.scaled_trigger
+        audio = np.clip(audio, -1.0, 1.0)
```
lcadalzo, line 63: just to get more context, was the previous absence of clipping here a bug?

armory/metrics/poisoning.py
```diff
@@ -54,11 +66,20 @@ def __init__(
...
         self.explanatory_model = explanatory_model
+        self.data_modality = data_modality
+        self.model_framework = model_framework
+        self.activation_layer = activation_layer
         self.resize_image = bool(resize_image)
         self.size = size
         self.resample = resample
         self.device = device
 
+        if self.model_framework == "tensorflow" and self.activation_layer is not None:
```
lcadalzo, line 77: could you add a comment here? This code looks a bit confusing without context.

armory/metrics/poisoning.py
```diff
@@ -46,6 +55,9 @@ class ExplanatoryModel:
     def __init__(
         self,
         explanatory_model,
+        data_modality="image",
+        model_framework="pytorch",
+        activation_layer=None,
```
lcadalzo, line 60: it's unclear what `activation_layer` is supposed to be

armory/metrics/poisoning.py
```diff
@@ -96,24 +126,31 @@ def get_activations(self, x, batch_size: int = None):
...
+        for i in range(0, len(x), batch_size):
+            x_batch = x[i : i + batch_size]
+
+            if self.model_framework == "pytorch":
+                with torch.no_grad():
+                    x_batch = self.preprocess(x_batch)
+                    activation, _ = self.explanatory_model(x_batch)
+                    activations.append(activation.detach().cpu().numpy())
+
+            elif self.model_framework == "tensorflow":
```
lcadalzo, line 145: what's supposed to happen if `model_framework` is neither `pytorch` or `tensorflow`? Should we raise a ValueError?

armory/art_experimental/attacks/poison_loader_audio.py
swsuggs: David copied and modified this from ART because their version did something very slowly.  Clipping does not happen in the ART version either, but I believe it is a bug to return an audio sample that exceeds the valid range.  I filed an issue in ART for this as well.

armory/metrics/poisoning.py
```diff
@@ -58,27 +62,41 @@ def __init__(
         data_modality="image",
         model_framework="pytorch",
         activation_layer=None,
-        resize_image=True,
-        size=(224, 224),
-        resample=Image.BILINEAR,
-        device=DEVICE,
+        preprocess_kwargs={},
```
lcadalzo, line 65: I suggest having `preprocess_kwargs` default to `None` rather than a dict. See, e.g., [here](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/) for a description of potential issues with default mutable args

You could then do:
```
if preprocess_kwargs is None:
    preprocess_kwargs = {}
```
You should then be able to remove all the various
```
        "preprocess_kwargs": {},
```

that are in `EXPLANATORY_MODEL_CONFIGS`, which would be cleaner

swsuggs: Agree, done

armory/metrics/poisoning.py
swsuggs: Error raised.

armory/metrics/poisoning.py
swsuggs: Added docstrings

lcadalzo:
I had to change `"compute_fairness_metric"` in the config but now that I'm running the code affected by this PR, I'm seeing the following issue where the script is being killed by the `self.explanatory_model()` call:

On laptop, using `breakpoint()`:
```
> /workspace/armory/metrics/poisoning.py(156)get_activations()
-> activation = self.explanatory_model(x_batch, training=False)
(Pdb) n
Killed
I have no name!@1d23884bdaeb:/workspace$
```

And on remote cluster using a GPU:
```
2023-02-01 17:57:03 49s INFO     armory.scenarios.poison:compute_explanatory:630 Computing fairness metrics
2023-02-01 17:57:09.852144: F ./tensorflow/core/util/gpu_launch_config.h:160] Check failed: work_element_count > 0 (0 vs. -1591109476)
Aborted (core dumped)
```

Have you been able to run this code successfully? Are you able to reproduce this?

swsuggs:
If you are running an audio config you may have to use ```fit_generator:true``` to avoid OOM issues
```
    ...
    "scenario": {
        "kwargs": {
            "fit_generator": true
        },
    ...
```
Not sure if that's the issue you are hitting; it's not a very descriptive error.

swsuggs:
Ok I was finally able to replicate your error by skipping training and not using fit_generator.  If I train but don't use fit_generator I get OOM.

swsuggs:
@lcadalzo In case I wasn't clear, I don't believe the error you hit is an issue because it can be avoided by using fit_generator.

lcadalzo:
What's the rationale behind fit_generator being a configurable scenario kwarg?

lcadalzo (approved):
LGTM

PR #1872, opened 2023-01-27
carla_obj_det_train

ppark-twosixtech:


armory/datasets/standard/carla_obj_det_train/carla_obj_det_train_test.py
```diff
@@ -0,0 +1,25 @@
+"""carla_obj_det_train dataset."""
```
lcadalzo, line 1: is this the file you were asking about? You can look [here](https://github.com/twosixlabs/armory/tree/tfdsv4/armory/datasets/standard) for examples of what already-integrated datasets look like. We should have tests for this dataset already in `tests/`, so this file along with `TODO-add_fake_data...` I believe can be removed

ppark-twosixtech: Resolved with 941be98

lcadalzo:
@ppark-twosixtech is this PR meant to still be a Draft? I can build the dataset ok and run with `--skip-attack` without error. Note that there's still the issue of the extra batch dimension (which this PR doesn't address, hence using tide metrics here fails), but we have a separate issue open for that. I'm ok merging this while keeping that issue open

PR #1879, opened 2023-02-02
Add sleeper agent resnet

swsuggs:
Sleeper Agent doesn't work very well with the resnet18 already in armory.  Rather than attempt to tune hyperparameters, the poisoning group desired us to add this resnet version on which we know the attack succeeds.


swsuggs:
I'm just running some configs with the full attack to make sure everything works.  @lcadalzo Feel free to begin reviewing the code

lcadalzo (approved):
LGTM

PR #1883, opened 2023-02-03
Carla metrics bug fix

swsuggs:
Mike's description of the error from slack:

> it seems that per example metrics are always collected for OD scenarios even if ```metric -> record_metric_per_sample``` is ```False```. In addition, with the exception of ```AP_per_class```, the other metrics no longer report the mean statistic even if ```metric -> mean``` is ```True```.

This is essentially because the ```GlobalMeter``` class used for ```@populationwise``` metric functions assumes that the functions already return the mean over all samples.  This is true for ```AP_per_class```, but not the others.

Easy fix: just have the metric functions return the mean over all samples.  This seems to be what the writer of this code (David) expected.

But that leaves questions about why the carla OD functions are marked @populationwise in the first place and whether a user would ever want to see per-image results.  In that case, should this be controllable from the config, or should we just return both?

### DIFF
--- a/armory/metrics/task.py
+++ b/armory/metrics/task.py
@@ -1225,74 +1225,98 @@ def object_detection_hallucinations_per_image(
 
 @populationwise
 def carla_od_hallucinations_per_image(
-    y_list, y_pred_list, iou_threshold=0.5, score_threshold=0.5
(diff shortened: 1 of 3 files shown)


swsuggs:
The present change simply returns both mean and per-image.  @lcadalzo Please weigh in, if you'd like, on what the behavior of this should be.

PR #1886, opened 2023-02-10
Yet another sleeper agent bug fix

swsuggs:
Fixes bug in sleeper agent scenario where it was incorrectly finding the indices of poisoned data.

Also includes a couple less crucial improvements (invalid variable name in infrequently-reached code; improved parameter values for dp-instahide baseline configs)


jprokos26 (approved):
Attack succeeds with success rate `0.041` when tested.

PR #1889, opened 2023-02-20
MOT export bug: removing decrement of index to avoid KeyError

### LINKED ISSUE #1888: [BUG]: MOT exporting should not adjust dataset timesteps
### Description of the bug

Given that a[ preprocessing function ](https://github.com/twosixlabs/armory/blob/f529d887e95725bb730c11c5810a68dbe117cc11/armory/data/adversarial_datasets.py#L1019)was added to the MOT dataset loader to 0-index the timesteps, the exporting code no longer needs to [adjust](https://github.com/twosixlabs/armory/blob/f529d887e95725bb730c11c5810a68dbe117cc11/armory/instrument/export.py#L331-L336) for 1-indexed timesteps.

### Steps To Reproduce

Run any MOT config (that doesn't use coco format data) with "export_batches":true; you should see a key error because the timestep indices are off by one

### Additional Information

_No response_

lcadalzo:
Closes #1888

### DIFF
--- a/armory/instrument/export.py
+++ b/armory/instrument/export.py
@@ -327,13 +327,11 @@ class VideoTrackingExporter(VideoClassificationExporter):
 
         if y.ndim == 1:
             y = np.expand_dims(y, 0)
-        y_dict = {
-            int(timestep - 1): [] for timestep in set(y[:, 0])
-        }  # 0-index timesteps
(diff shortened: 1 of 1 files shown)


swsuggs (approved):
Looks great, thanks

PR #1890, opened 2023-02-24
New armory utils rgb-convert utility

jprokos26:
Added 

Added `__main__.py` file to armory.utils to allow user to view saved depth images interactively using matplotlib in various formats (original, linear, log).

When using X11 forwarding, can run the following to produce a matplotlib figure that can switch between the globbed images as well as between their depth formats:
```
armory utils rgb-convert ~/git/.armory/outputs/<RUN_ID>/saved_samples
```
Otherwise, can supply the `--save` or `--output` flag to write to disk.

TODO:
- Add this to the docs somewhere

### Demo Usage
# Demo


jprokos26:
@christopherwoodall Assigning you for tracking, feel free to make the TODO change otherwise I plan to finish this PR by EOW and will request review then

jprokos26:
TODO, update structure to be like #1928

christopherwoodall (approved):
Looks good. I would note, for the future, we should monitor `tools.py`. If it grows too much longer we may need to start breaking the individual commands/methods into their own modules; e.g. something like `armory/cli/tools/rgb_converter.py`.

PR #1892, opened 2023-02-28
Yet another sleeper agent bug fix (#1886) (#1887)

mwartell:
* correction to finding the poison indices

* fix variable name, also raise error instead of warning

* better noise values for dp instahide defense

* removing leftover line

* sleeper agent baseline results update


swsuggs (approved):

swsuggs:
This is just to get the 0.16.6 commit tag on develop.

Issue #1897, opened 2023-03-08
Advance ART to 1.14.0


swsuggs:
ART will be released around the 15th, plus or minus a day.

swsuggs:
Completed #1901

PR #1901, opened 2023-03-17
advance art to 1.14.0

swsuggs:
for #1897


swsuggs:
@mwartell are there any tests/checks that need to be run manually?  I see that was done last time but not sure if it's still necessary

mwartell (approved):
I confirm that `14 > 13` and that the CI has completed successfully.

PR #1904, opened 2023-03-23
Generic Masked Patch Attack

jprokos26:
The goal of this PR is to implement a generic masked patch module to be integrated into existing attacks with minimal code change to the attack module

## How to use this code
See the docs

To illustrate what this does see the example below:

## Example Run
<details open>
<summary>Attack kwargs from config</summary>
<br>
<pre>
"attack": {
        "knowledge": "white",
        "kwargs": {
            "batch_size": 1,
            "learning_rate": 0.1,
            "max_iter": 2,
            "optimizer": "pgd",
            "targeted": false,
            "verbose": true,
            "patch_base_image": "https://github.com/jprokos26/armory/blob/public-base-images/armory/art_experimental/attacks/base_images/%5BCARLA_OD%5D_stable_diffusion_gemstones.png?raw=true",
            "patch_mask": {
                "shape": "grid",
                "shape_kwargs": {
                    "num_circles": 5,
                    "circle_radius": 0.1,
                    "circle_spacing": 2
                },
                "invert": false,
                "fill": "https://github.com/jprokos26/armory/blob/public-base-images/armory/art_experimental/attacks/base_images/white_w_black_text.png?raw=true"
            }
        },
        "module": "armory.art_experimental.attacks.carla_obj_det_adversarial_patch",
        "name": "CARLAAdversarialPatchPyTorch",
        "use_label": true
    },
</pre>
</details>

<details open>
<summary>Resultant image</summary>
<br>
image
</details>


armory/art_experimental/attacks/carla_obj_det_utils.py
```diff
@@ -97,3 +106,208 @@ def rgb_depth_to_linear(r, g, b):
     depth_m = r_ + g_ * 256 + b_ * 256 * 256
     depth_m = depth_m * 1000.0 / (256**3 - 1)
     return depth_m
+
+
+@dataclass
+class PatchMask:
```
jprokos26, line 112: I'm not sure if this file is the correct place for this class since we do not have a dedicated 'attack utils' file. Should this be moved to its own file?

armory/utils/shape_gen.py
```diff
@@ -0,0 +1,410 @@
...
+from io import BytesIO
+import os
+from pathlib import Path
+from typing import Union
+
+from PIL import Image
+import cairosvg
+import matplotlib.pyplot as plt
+import numpy as np
+import svg
```
jprokos26, line 13: Which subgroup should cairosvg and svg.py be added to in the pyproject.toml for code that will be potentially called during an armory run? Should this be in engine or pytorch? Which subgroup gets installed on the base docker image?

jprokos26:
@christopherwoodall Still WIP, not worth you trying to debug the current issue with patch_base_image but feel free to address other comments; will request review by EOW with fixes

armory/art_experimental/attacks/carla_obj_det_utils.py
jprokos26: Should be moved to be within `armory.utils`

jprokos26:
@mwartell I am running into [python package issues](https://github.com/twosixlabs/armory/actions/runs/5135916795/jobs/9241967362?pr=1904#step:8:75) when installed _not_ in editable mode but it works as expected when it is. Specifically armory cannot find `armory.cli.tools` after installation (failing [here](https://github.com/twosixlabs/armory/pull/1904/commits/46ef56c8bff961c92e89197491e360ae45c7eea5#diff-2fbe9c250c28022874410d5003b3350258e33e60f269e76ac3d91d62b7b78176R1)), I haven't had time to track down exactly why this is the case, I have tried relative and absolute imports to no avail, what might I be doing wrong? Do I need to make changes to our `pyproject.toml` to have it find the `tools` package properly?

mwartell:
The culprit is an overzealous exclusion in [pyproject.toml](https://github.com/twosixlabs/armory/blob/5a5daedc20671cd0ae6a50e91c6e8bc50776761b/pyproject.toml#L160)

I'm looking for the syntax to fix that, likely `exclude = ["/tools"]`

armory/cli/tools/generate_shapes.py
```diff
@@ -0,0 +1,59 @@
+import argparse
+from pathlib import Path
+
+
+def generate_shapes(command_args, prog, description):
+    from armory.utils.shape_gen import (  # move to lazy load to avoid PIL import error
+        Shape,
+    )
```
jprokos26, line 8: Is there a better way to handle this? In particular, Pillow is listed under `engine` and `develop` but when you install armory-testbed standard it is not included and will error when trying to run `armory --help`. This fixes that problem but creates another that if `generate_shapes` is called at runtime from an environment without PIL the error message will not be useful. However since this is strictly code that is called from either the cli in which the user will know how to fix the issue or from the `carla_adversarial_patch_pytorch` code which must be running with `engine` installed and PIL must also be installed I left it this way.

christopherwoodall: Totally in agreement that all required libraries should be in the `dependencies` section.

christopherwoodall: Totally in agreement that all required libraries should be in the dependencies section.

christopherwoodall (approved):
Let's get those CI tests passing, otherwise, it looks good to me.

PR #1905, opened 2023-03-24
Integrate BadDet

### LINKED ISSUE #1898: Integrate BadDet poisoning attacks
[Paper](https://arxiv.org/pdf/2205.14497.pdf)

[Demo notebook](https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/notebooks/poisoning_attack_bad_det.ipynb)

[Attack code](https://github.com/Trusted-AI/adversarial-robustness-toolbox/tree/main/art/attacks/poisoning/bad_det)

Probably most of the work will be on metrics since this is an object detection attack.  The fairness metrics are built around image classification tasks.

swsuggs:
Integrating 4 object detection poisoning attacks with associated metrics to close #1898.  Opted for one scenario with minor case switching rather than 4 separate scenarios inheriting a 5th.  

Will be adding model (YoloV3) and data as well.

Fairness metrics will be addressed in a separate PR after additional discussion with the poisoning team.


lcadalzo:
running the config included in this PR, I'm getting the following error when `fit()` is run:
```

  File "/workspace/armory/scenarios/poison.py", line 359, in fit
    self.model.fit(
    │    │     └ <function PyTorchObjectDetector.fit at 0x7f744c5e84c0>
    │    └ PyTorchFasterRCNN(channels_first=False, model=FasterRCNN(
    │        (transform): GeneralizedRCNNTransform(
    │            Normalize(mean=[0.48...
    └ <armory.scenarios.poisoning_obj_det.ObjectDetectionPoisoningScenario object at 0x7f7522feedf0>

  File "/opt/conda/lib/python3.9/site-packages/art/estimators/object_detection/pytorch_object_detector.py", line 463, in fit
    raise ValueError("An optimizer is needed to train the model, but none for provided.")

ValueError: An optimizer is needed to train the model, but none for provided.

```

lcadalzo:
got it, if the changes were temporary then please Slack me them

armory/metrics/task.py
```diff
@@ -1311,6 +1311,210 @@ def carla_od_AP_per_class(y_list, y_pred_list, iou_threshold=0.5, mean=True):
     )
 
 
+def _object_detection_poisoning_get_targeted_mr_dr(
+    y_list, y_pred_list, source_class=None, target_class=None, iou_threshold=0.5, score_threshold=0.5,
```
lcadalzo, line 1315: just so I understand, what's prompted the need for a new function rather than using the disappearance/misclassification rate code we already have?

armory/metrics/task.py
```diff
@@ -1311,6 +1311,210 @@ def carla_od_AP_per_class(y_list, y_pred_list, iou_threshold=0.5, mean=True):
...
+        else:
+            gt_boxes = y["boxes"]
+            gt_labels = y["labels"]
+
+        # initialize count of disappearances
+        num_disappearances = 0
+        num_gt_boxes = len(gt_boxes)
+
+        # Initialize array to count misclassified
+        true_positive_array = np.zeros((num_gt_boxes,))
```
lcadalzo, line 1359: doesn't appear `true_positive_array` is used anywhere

armory/metrics/task.py
```diff
@@ -1311,6 +1311,210 @@ def carla_od_AP_per_class(y_list, y_pred_list, iou_threshold=0.5, mean=True):
...
+        targeted_misclassification_array = np.zeros((num_gt_boxes,))
+        disappearance_array = np.zeros((num_gt_boxes,))
```
lcadalzo, line 1361: doesn't appear `disappearance_array` is used anywhere

armory/metrics/task.py
```diff
@@ -1311,6 +1311,210 @@ def carla_od_AP_per_class(y_list, y_pred_list, iou_threshold=0.5, mean=True):
...
+    returns: a tuple of length 2 (MR, DR) where each element is a list of length equal
+        to the number of images.
+    """
+
+    targeted_misclassification_rate_per_img = []
+    disappearance_rate_per_img = []
+
+
+    for img_idx, (y, y_pred) in enumerate(zip(y_list, y_pred_list)):
+        if source_class:
```
lcadalzo, line 1345: I think this maybe needs to be changed to `if source_class is not None`, bc if `source_class` is 0, then the else block is reached, which I believe isn't the desired behavior. I've set a breakpoint() and noticed this happening, and so `gt_labels` includes other label ids:

```
(Pdb) source_class
0
(Pdb) gt_labels
array([1, 1, 1, 1, 1, 1, 1, 1, 2, 2])

```

armory/metrics/task.py
swsuggs: Mainly the targeted nature of this metric.  For regional misclassification, we only count a box if the gt label is source and it was predicted as target.  For global misclassification, it only counts if the gt label is not target but it is predicted as target.  The existing code did not support this.  

For disappearance, it would be possible to use the existing code with ```class_list=source```, but it does something slightly different than my interpretation of what this metric is trying to capture.  I simply want to count source class boxes with no overlapping predicted box (over the threshold).  The existing code uses ```1 - TPR - misclassification_rate```.  I believe this could result in too low a disappearance rate if a GT box has multiple overlapping predictions.  For example, imagine two GT boxes, one with no prediction, and one with two predictions of which one is correct and one is wrong.  Disappearance rate would be 0 even though one GT box had nothing predicted.

Although a comment in the original code states ```Convert these arrays to binary to avoid double-counting (i.e. when multiple predicted boxes overlap with a single ground-truth box)```, this only avoids double counting two misclassifications or two TPs, not one of each.

I am happy to discuss this more or change my view if I'm missing something.

armory/art_experimental/attacks/poison_loader_dlbd.py
```diff
@@ -33,11 +33,26 @@ def mod(x):
...
+            y_shift = 0
+        elif "x_shift" in kwargs and "y_shift" in kwargs:
+            x_shift = kwargs.get("x_shift")
+            y_shift = kwargs.get("y_shift")
+        elif "base_img_size_x" in kwargs and "base_img_size_y" in kwargs:
+            # default to center of image
+            base_img_size_x = kwargs.get("base_img_size_x")
+            base_img_size_y = kwargs.get("base_img_size_y")
+            x_shift = (base_img_size_x - size[0]) // 2
+            y_shift = (base_img_size_y - size[1]) // 2
```
jprokos26, line 51: Are these kwargs documented anywhere? I can't seem to find what they are doing although I can make an educated guess. Might be worth adding something to https://github.com/twosixlabs/armory/blob/develop/docs/poisoning.md#configuration-files about what kwargs are available.

armory/art_experimental/attacks/poison_loader_obj_det.py
```diff
@@ -0,0 +1,33 @@
+from armory.art_experimental.attacks.poison_loader_dlbd import poison_loader_dlbd
+from armory.utils import config_loading
+
+bad_det_attacks = [
+    "BadDetRegionalMisclassificationAttack",
+    "BadDetGlobalMisclassificationAttack",
+    "BadDetObjectGenerationAttack",
+    "BadDetObjectDisappearanceAttack",
```
jprokos26, line 8: Should these be listed in https://github.com/twosixlabs/armory/blob/develop/docs/poisoning_object_detection.md#configuration-files?
Not a huge deal since error should inform user of options.

armory/baseline_models/pytorch/yolov3.py
```diff
@@ -0,0 +1,49 @@
...
+    detector = PyTorchYolo(
+        model=model_wrapper,
+        device_type=DEVICE,
+        input_shape=(416, 416, 3),
+        optimizer=optimizer,
+        clip_values=(0, 1),
+        channels_first=False,
+        attack_losses=("loss_total",),
+    )
+    return detector
```
jprokos26, line 49: You aren't using `model_kwargs` or `wrapper_kwargs` at all here. I'm not sure what _should_ be kwarg'able but these options seem appropriate to me:

```suggestion
def get_art_model(
    model_kwargs: dict, wrapper_kwargs: dict, weights_path: Optional[str] = None
) -> PyTorchYolo:

    model_path = "armory/baseline_models/pytorch/yolov3.cfg"
    model = load_model(model_path=model_path, weights_path=weights_path)
    model_wrapper = Yolo(model)

    params = [p for p in model_wrapper.parameters() if p.requires_grad]
    optimizer = torch.optim.SGD(
        params,
        lr=model_kwargs.get("learning_rate", 0.01),
        momentum=model_kwargs.get("momentum", 0.9),
        nesterov=model_kwargs.get("nesterov", True),
    )

    detector = PyTorchYolo(
        model=model_wrapper,
        device_type=DEVICE,
        input_shape=(416, 416, 3),
        optimizer=optimizer,
        clip_values=(0, 1),
        channels_first=False,
        attack_losses=wrapper_kwargs.get("attack_losses", ("loss_total",)),
    )
    return detector
```

armory/scenarios/poisoning_obj_det.py
```diff
@@ -0,0 +1,529 @@
+import copy
+import imgaug.augmenters as iaa
+from imgaug.augmentables.bbs import BoundingBox, BoundingBoxesOnImage
```
jprokos26, line 3: In lieu of https://twosixtechnologies.slack.com/archives/G021C1YDUN5/p1682968354706919 it may be worth having a try catch block here since yoloattacks is not in the default dependencies list.

```suggestion
try:
    import imgaug.augmenters as iaa
    from imgaug.augmentables.bbs import BoundingBox, BoundingBoxesOnImage
except ImportError:
    raise ImportError(
        "imgaug is not installed. Please install via pip install armory-testbed[yolo]."
    )
```

armory/scenarios/poisoning_obj_det.py
```diff
@@ -0,0 +1,529 @@
...
+        self.attack_variant = self.config["attack"]["kwargs"]["attack_variant"]
+        self.train_epochs = adhoc_config["train_epochs"]
+        self.fit_batch_size = adhoc_config.get(
+            "fit_batch_size", self.config["dataset"]["batch_size"]
+        )
+        self.config["dataset"]["batch_size"] = 1
+        # Set batch size to 1 for loading data, this simplifies filtering small boxes.
+        # Note, training is unaffected since self.fit_batch_size is already set above.
+
+        self.label_function = lambda y: y
```
jprokos26, line 154: What is the purpose of this? I don't see any way for this to be set elsewhere, can likely remove this line and replace line 296 (the only place using this)
```
                            self.label_function(
                                aug_y_train[i : i + self.fit_batch_size]
                            ),
```
with
```
                            aug_y_train[i : i + self.fit_batch_size],
```
I've made a separate comment with this change.


```suggestion
```

armory/scenarios/poisoning_obj_det.py
```diff
@@ -0,0 +1,529 @@
...
+                )
+
+                # Manually call model.fit with small batches
+                for i in range(0, len(aug_y_train), self.fit_batch_size):
+                    if i + self.fit_batch_size < len(aug_y_train):
+                        self.model.fit(
+                            aug_x_train[i : i + self.fit_batch_size],
+                            self.label_function(
+                                aug_y_train[i : i + self.fit_batch_size]
+                            ),
```
jprokos26, line 293:
```suggestion
                            aug_y_train[i : i + self.fit_batch_size],
```

armory/scenarios/poisoning_obj_det.py
```diff
@@ -0,0 +1,529 @@
...
+        y_pred_adv = [self.non_maximum_supression(pred) for pred in y_pred_adv]
+        self.probe.update(x_adv=x_adv, y_adv=y_adv, y_pred_adv=y_pred_adv)
+
+        self.x_adv = x_adv
+        self.y_pred_adv = y_pred_adv
+
+    def load_fairness_metrics(self):
+        raise NotImplementedError(
+            "The fairness metrics have not been implemented for object detection poisoning"
+        )
```
jprokos26, line 501: Should there be an issue tracking this implementation or should this never be implemented?

armory/art_experimental/attacks/poison_loader_dlbd.py
swsuggs: They are the kwargs for the art class, so the documentation would be there in ART.  I'll add a note in the doc

armory/scenarios/poisoning_obj_det.py
jprokos26: Nvmd I see it is used in https://github.com/twosixlabs/armory/blob/1b88fd21e27ddac619955dae30a6062a53c1c379/armory/scenarios/poisoning_witches_brew.py#L305

swsuggs: That's witches brew, but actually it is used in the inherited function [filter_dataset](https://github.com/twosixlabs/armory/blob/1b88fd21e27ddac619955dae30a6062a53c1c379/armory/scenarios/poison.py#L271)

armory/scenarios/poisoning_obj_det.py
swsuggs: There's no plan to implement.  As currently defined, they only apply to image classification.  But I can clarify the error message.

armory/baseline_models/pytorch/yolov3.py
swsuggs: Actually those dicts are supposed to be passed through to load_model and PytorchYolo respectively; thanks for catching that.  In fact I will use model_kwargs to pass the path to the config file instead of hard coding it there.

armory/scenarios/poisoning_obj_det.py
swsuggs: This won't be the only thing someone's missing if they didn't install the [yolo] version.  but it can't hurt

Issue #1909, opened 2023-04-05
Annotate exported OD images with predicted labels


swsuggs:
It would be useful, especially with the upcoming OD poisoning scenario, if the predicted labels were displayed on the exported images next to their respective boxes.

PR #1910, opened 2023-04-05
Label bbox in exported OD images

swsuggs:
For #1909.

This simply writes the (integer) class label on a 10x10 box next to the bbox.  Advantages: simple; no kwargs or font files to include.  Disadvantages: size may not be appropriate for all image/bbox sizes; string labels could be more useful than integers.


jprokos26:
> For #1909.
> 
> This simply writes the (integer) class label on a 10x10 box next to the bbox. Advantages: simple; no kwargs or font files to include. Disadvantages: size may not be appropriate for all image/bbox sizes; string labels could be more useful than integers.

I really like the idea of keeping this simple so that there are no kwargs to deal with. I don't think embedding the label name is a good idea since there isn't a lot of space, I think an easy fix here would be to log out to the user what each class value represents the first time you write a sample such that the user can confirm what the values mean from the `armory_log.txt`.

Additionally the class label may get lost when there are many overlapping hallucinations (see below). I see two ways to resolve this issue:
1. Dynamically choose font size based on bbox size, realistically we can probably go down to ~5pt font while still keeping 1-9 distinguishable by zooming in (default seems to be 10?). Would require minor logic updates and would keep it relatively simple although I'm not sure how legible smaller font remains without additional testing
2. Remove the label text alltogether and instead color code the boxes and again output a legend to `armory_log`. Additionally could label the largest bbox (or just y_true boxes) for each label with its name to act as a "legend" such that a user can decipher what each color means by looking at the image. This would be sizably more logic changes than the previous suggestion and could get confusing when there are many labels/colors for someone new looking at the export.
3. Leave it as is, I'm not sure how much of a problem it is to have overlapping labels since in this example they are all the same predicted class. Might not be worth the effort to complicate this logic when it works fine as is.

batch_0_ex_0_scenario x_adv_with_boxes

swsuggs:
Yeah logging out the class names would be nice, but I don't think those are even available anywhere, other than possibly a comment in each dataset builder file.  So we'd have to maintain a dictionary somewhere and select the labels for the current dataset.  Is that worth the trouble?

For overlapping boxes, my thought is that this is most useful for misclassification errors, in which case the labels will still be legible since ground truth boxes aren't that close.  With hallucinations, I care less about each individual box.  Maybe it's good to be able to see we are mostly hallucinating pedestrians, but I think we can get that sense even when some are covered.

So overall I don't feel like additional readability would justify the trouble.  I'd prefer to start here and see if it satisfies users; we can always improve it later.

jprokos26 (approved):
Tested working, LGTM!

PR #1911, opened 2023-04-10
avoid tfds version 4.9.0

swsuggs:
Tfds version 4.9.0 has trouble importing on macos apparently, causing problems with our github tests.  See this [issue](https://github.com/tensorflow/datasets/issues/4852) for some context. 

We should be able to dodge this problem by not using that tfds version


christopherwoodall (approved):

christopherwoodall:
LGTM!

PR #1913, opened 2023-04-17
Update notebook Google Colab configuration

christopherwoodall:
Update notebook to run in Google Colab environment w/out error.


mwartell:
This still fails for me and that first cell is really daunting for a newbie. Let's discuss

image

PR #1914, opened 2023-04-24
Remove unsupported option from precommit hooks

christopherwoodall:
Removes the ` --indent=4` option until Python 3.8 support is dropped.


mwartell (approved):
The thing that slightly bugs me about this change is that it may alter the canonical form of our config.json files. Since they are deprecated anyway, it is probably okay.

Issue #1922, opened 2023-05-02
Automatic GPU selection


jprokos26:
Currently when gpu is not specified, armory defaults to running the attack on GPU0 and loading some other data onto all other GPUs. It would be useful to be able to launch armory without having to check gpustat.

When gpu is not set to a specific integer value and `use_gpu` is set to True, utilize the following function to automatically detect the best gpu to use as done [elsewhere](https://github.com/twosixlabs/gard-wp/blob/main/yoloattacks/utils.py#L50-L81):

```python
def get_device(device=None) -> torch.device:
    """
    If None is passed, the GPU with the most free memory will be used.
    If False is passed, the CPU will be used.
    Otherwise, the device will be set to the specified GPU.
    """
    if device is False:
        log.info("Using CPU")
        return torch.device("cpu")
    elif device is None:
        # Get the current memory usage of all GPUs using nvidia-smi
        memory = {
            i: int(
                os.popen(
                    f"nvidia-smi --query-gpu=memory.free --format=csv,noheader,nounits --id={i}"
                ).read()
            )
            for i in range(torch.cuda.device_count())
        }

        # Get the GPU with the most free memory
        device = torch.device(f"cuda:{max(memory, key=memory.get)}")
        log.info(
            f"Autoselected {device} with {memory[device.index]} MB bytes free"
        )
        return device
    else:
        try:
            return torch.device(f"cuda:{device}")
        except Exception as e:
            log.error(f"Error setting device to {device}: {e}")
            return torch.device("cpu")
```

swsuggs:
@christopherwoodall this was not completed by 1925, is this no longer desired?

Issue #1902, opened 2023-03-17
[BUG]: `--no-gpu` does not work with `--no-docker`


jprokos26:
### Description of the bug

When running a config (which has `use_gpu` set to false) with both the `--no-docker` and `--no-gpu` flag, all available gpus are used. 

### Steps To Reproduce

Run `armory run ~/git/twosixlabs/armory/scenario_configs/cifar10_baseline.json --no-docker --no-gpu` which will cause:
```
$ nvidia-smi 
+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
|    0   N/A  N/A   3172002      C   ...s/.venv/py310/bin/python3     1095MiB |
|    1   N/A  N/A   3172002      C   ...s/.venv/py310/bin/python3      337MiB |
|    2   N/A  N/A   3172002      C   ...s/.venv/py310/bin/python3      277MiB |
|    3   N/A  N/A   3172002      C   ...s/.venv/py310/bin/python3      159MiB |
|    4   N/A  N/A   3172002      C   ...s/.venv/py310/bin/python3      159MiB |
|    5   N/A  N/A   3172002      C   ...s/.venv/py310/bin/python3      159MiB |
|    6   N/A  N/A   3172002      C   ...s/.venv/py310/bin/python3      159MiB |
|    7   N/A  N/A   3172002      C   ...s/.venv/py310/bin/python3      159MiB |
+-----------------------------------------------------------------------------+
```

### Additional Information

Run using
```
$ which python
/home/jonathan.prokos/.venv/py310/bin/python
$ armory --version
0.16.6.post4+gdbb73b3d.d20230310
```

jprokos26:
@christopherwoodall I have zero context as to how GPUs get grabbed, I believe it is due to CUDA environment variables not being set properly; this is not a high priority bug but would be nice to fix for debugging

PR #1925, opened 2023-05-04
set cuda devices for no-docker mode

### LINKED ISSUE #1923: Add global torch.device variable set based on `--gpus` flag to prevent issues between docker & no-docker device ids
Related to #1902 and #1922 

We use `torch.device("cuda")` a lot in our code:
https://github.com/search?q=repo%3Atwosixlabs%2Farmory%20device&type=code

There should instead be a single device variable to be inherited such that we do not need to track the device-id discrepancy between being in a docker container or not. This will also allow for us to set `torch.device("cpu")` to address #1902.

swsuggs:
I believe this retains the existing behavior when using docker.
For no-docker mode, this sets CUDA_VISIBLE_DEVICES to the list of gpus, if specified, or to -1 if the gpu is not to be used.  
It still defaults to using all gpus if use_gpu is True but no gpus are specified.

@christopherwoodall @jprokos26 Do you have any comments about this approach?  Any cases I'm overlooking?

Should fix #1923 and #1902.  Does not address #1922.

### DIFF
--- a/armory/eval/evaluator.py
+++ b/armory/eval/evaluator.py
@@ -95,7 +95,15 @@ class Evaluator(object):
         if self.config["sysconfig"].get("use_gpu", None):
             gpus = self.config["sysconfig"].get("gpus")
             if gpus is not None:
-                self.extra_env_vars["NVIDIA_VISIBLE_DEVICES"] = gpus
+                if self.no_docker:
+                    self.extra_env_vars["CUDA_VISIBLE_DEVICES"] = gpus
+                else:
+                    self.extra_env_vars["NVIDIA_VISIBLE_DEVICES"] = gpus
+        else:
+            if self.no_docker:
+                # Must be explicitly set for no-docker
+                self.extra_env_vars["CUDA_VISIBLE_DEVICES"] = "-1"
+
         if self.config["sysconfig"].get("set_pythonhashseed"):
             self.extra_env_vars["PYTHONHASHSEED"] = "0"
 
--- a/armory/eval/evaluator.py
+++ b/armory/eval/evaluator.py
@@ -96,10 +96,10 @@ class Evaluator(object):
             True if os.getenv("CUDA_VISIBLE_DEVICES") is not None else False
         )
         if cuda_var_exists and self.no_docker:
+            # Existing value should override armory flags
             log.warning(
                 "CUDA_VISIBLE_DEVICES is set; any Armoy gpu instructions will be ignored"
             )
-            # Existing value should override armory flags
 
         # Set visible gpus
         if self.config["sysconfig"].get("use_gpu", None):
--- a/armory/eval/evaluator.py
+++ b/armory/eval/evaluator.py
@@ -95,7 +95,9 @@ class Evaluator(object):
         cuda_var_exists = (
             True if os.getenv("CUDA_VISIBLE_DEVICES") is not None else False
         )
-        # Existing value should override armory flags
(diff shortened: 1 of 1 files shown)


dxoigmn:
We use `CUDA_VISIBLE_DEVICES` to control what GPUs a process can see and thus run on. It would be good if Armory respects an already set value of `CUDA_VISIBLE_DEVICES` in no-docker mode. We could switch to using Armory's flags, but `CUDA_VISIBLE_DEVICES` works regardless of process (assuming such processes do not muck with that environment variable).

swsuggs:
@dxoigmn thanks for the input, let me know if this revision will satisfy your need.

dxoigmn:
> @dxoigmn thanks for the input, let me know if this revision will satisfy your need.

LGTM!

armory/eval/evaluator.py
```diff
@@ -92,10 +92,24 @@ def _gather_env_variables(self):
...
+
+        # Set visible gpus
         if self.config["sysconfig"].get("use_gpu", None):
             gpus = self.config["sysconfig"].get("gpus")
             if gpus is not None:
-                self.extra_env_vars["NVIDIA_VISIBLE_DEVICES"] = gpus
+                if self.no_docker and not cuda_var_exists:
+                    self.extra_env_vars["CUDA_VISIBLE_DEVICES"] = gpus
+                if not self.no_docker:
+                    self.extra_env_vars["NVIDIA_VISIBLE_DEVICES"] = gpus
```
jprokos26, line 107: Would be useful to at a minimum show a warning (or throw error) when CUDA_VISIBLE_DEVICES is already set but a gpu flag is passed in as running `CUDA_VISIBLE_DEVICES=3 python3 -m armory run scenario_configs/cifar10_baseline.json --no-docker --gpus 1` will load everything onto gpu 3 without any warnings.

Change below tested to work locally.


```suggestion
                if self.no_docker and not cuda_var_exists:
                    self.extra_env_vars["CUDA_VISIBLE_DEVICES"] = gpus
                elif not self.no_docker:
                    self.extra_env_vars["NVIDIA_VISIBLE_DEVICES"] = gpus
                else:
                    log.warning(
                        "CUDA_VISIBLE_DEVICES already set, ignoring gpus argument"
                    )
```

jprokos26 (changes requested):
Will approve once addressed.

armory/eval/evaluator.py
swsuggs: I've pulled the warning up a level so it will be logged whether use_gpu is true or false.

armory/eval/evaluator.py
```diff
@@ -92,10 +92,28 @@ def _gather_env_variables(self):
...
             self.extra_env_vars["VERIFY_SSL"] = "false"
 
+        cuda_var_exists = (
+            True if os.getenv("CUDA_VISIBLE_DEVICES") is not None else False
+        )
+        if cuda_var_exists and self.no_docker:
+            log.warning(
+                "CUDA_VISIBLE_DEVICES is set; any Armoy gpu instructions will be ignored"
+            )
+            # Existing value should override armory flags
```
jprokos26, line 102: Very minor formatting change
```suggestion
            # Existing value should override armory flags
            log.warning(
                "CUDA_VISIBLE_DEVICES is set; any Armoy gpu instructions will be ignored"
            )
```

jprokos26 (approved):
LGTM w/ minor formatting change

PR #1934, opened 2023-05-16
removed tools/run_scenario_configs.sh

mwartell:
this was causing new git clones to be dirty immediately after cloning. soooo weird! (msw & cjw)


christopherwoodall (approved):
LGTM!

Issue #1930, opened 2023-05-12
Modify attacks which utilize the `attacks.kwargs.patch_base_image` config option to accept files not in the same directory as the module.


jprokos26:
Currently the only way to use `patch_base_image` is to clone armory or use a custom attack module. This is done because it will definitely be included in the docker container. Maybe make it so you can supply a url and that will be enough?

PR #1939, opened 2023-05-25
Add support for patch_base_image when not running locally

jprokos26:
Addresses #1930 

Searches for `patch_base_image` in more than one place:
1. Check if file exists in module path
2. Check if file exists in `armory.paths.runtime_paths().cwd`
3. Check if supplied `patch_base_image` is actually a url to an image we can download
Else fail

### DIFF
--- a/armory/art_experimental/attacks/carla_obj_det_adversarial_patch.py
+++ b/armory/art_experimental/attacks/carla_obj_det_adversarial_patch.py
@@ -54,24 +54,37 @@ class CARLAAdversarialPatchPyTorch(AdversarialPatchPyTorch):
         patch_base_image_path = os.path.abspath(
             os.path.join(module_folder, self.patch_base_image)
         )
-        # if the image does not exist, check paths.DockerPaths().cwd and paths.HostPaths().cwd
+        # if the image does not exist, check paths.DockerPaths().cwd
         if not os.path.exists(patch_base_image_path):
+            # TODO: determine if in docker mode smartly
             docker_cwd = paths.DockerPaths().cwd
             host_cwd = paths.HostPaths().cwd
             patch_base_image_path = os.path.abspath(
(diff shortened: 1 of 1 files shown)


jprokos26:
Whenever I tried to `pip install -e .` with these changes and move to a different directory when I launch the docker container the armory being run does not have these changes so I was unable to test that the cwd gets mapped properly to `runtime_paths().cwd` when running as a pip installed package not locally.

armory/art_experimental/attacks/carla_obj_det_adversarial_patch.py
```diff
@@ -48,12 +49,34 @@ def create_initial_image(self, size, hsv_lower_bound, hsv_upper_bound):
...
+        if not os.path.exists(patch_base_image_path):
+            patch_base_image_path = os.path.abspath(
+                os.path.join(paths.runtime_paths().cwd, self.patch_base_image)
+            )
+        # image not in cwd or module, check if it is a url to an image
+        if not os.path.exists(patch_base_image_path):
+            import requests
+
+            try:
+                response = requests.get(self.patch_base_image, allow_redirects=True)
```
jprokos26, line 67: Any worry about safety of links? i.e. zipbombs or something?

mwartell: cf. https://stackoverflow.com/questions/14270698/get-file-size-using-python-requests-while-only-getting-the-header for sample usage

jprokos26: Resolved with https://github.com/twosixlabs/armory/pull/1939/commits/67e002049b4c5c5a03f7b1074231c5346da52ae7

christopherwoodall (approved):
LGTM!

PR #1940, opened 2023-05-25
Update Google Colab notebook

christopherwoodall:
Updates cells relevant to Google Colab in order to handle numpy version mismatch. See, https://numpy.org/devdocs/user/troubleshooting-importerror.html#c-api-incompatibility


mwartell (approved):
presumed working

PR #1942, opened 2023-05-25
Minor BadDet improvements

swsuggs:
Fixed:

- [x] Potentially misleading error statement
- [x] source/target class in configs
- [x] final batch loop does not skip remaining data if there is less than ```batch_size```


christopherwoodall (approved):
LGTM!

PR #1944, opened 2023-06-06
clarify local repo path in config documentation

swsuggs:
At least one performer experienced confusion about the ```local_repo_path``` variable; I think this should help.


christopherwoodall (approved):
LGTM!

PR #1948, opened 2023-06-19
OD poison baseline results

swsuggs:


deprit (approved):
Object detection poisoning baselines all complete

PR #1952, opened 2023-07-03
simplify release process

mwartell:
release only from `master` instead of requiring a version branch like
`r0.17.2`
add convenience script for triggering release


christopherwoodall (approved):
LGTM!

PR #1954, opened 2023-07-03
Carla attack update

yusong-tan:
This PR updates two items:

1. The depth patch region now matches the RGB patch region
2. Added a targeted attack option for the CARLA object detection evasion scenarios. This includes changes to the attack code, new targeted attack configs, and a random target generator.


swsuggs (approved):
Looks great, thank you Mike!

PR #1957, opened 2023-07-10
Save original config

### LINKED ISSUE #1333: Results JSON doesn't copy all of the config from the input JSON
I observed this with certain fields in `wrapper_kwargs` being omitted. My guess is that these configs are modified in place by the consuming model. To fix this, we should make a defensive copy of the submitted config and drop that into the results json.

swsuggs:
Fixes #1333.  Ensure that the config saved in the results json is identical to the initiating config.  In some cases the scenario, attack, or model may modify it.


christopherwoodall (approved):
LGTM!

PR #1964, opened 2023-07-18
Add support for pre-computed fairness majority masks

f4str:
Modify the poisoning scenario to accept the `majority_masks` field which will attempt to read a pre-computed `npz` file which contains the majority masks for the train and test data. This will override the explanatory model if provided. 

The `npz` file must be placed in the root of the armory directory (under `armory/`). So the provided path:
```json
"adhoc": {
    ...
    "compute_fairness_metrics": true,
    "majority_masks": "audio_majority_masks.npz",
    ...
}
```
will look for `armory/audio_majority_masks.npz`. The file will have two keyword arg arrays: `train` and `test` which are boolean arrays the same length as the train and test sets, respectively.

This works on both docker and non-docker mode.


f4str:
A zip file containing the majority masks for the speech commands dataset and a sample config file for testing: [majority_masks.zip](https://github.com/twosixlabs/armory/files/12089186/majority_masks.zip)

swsuggs:
Thank you Farhan, this is great.  I'm just adding the npz file to its own directory with a bit of extra code to facilitate loading it.

swsuggs (approved):

PR #1959, opened 2023-07-12
Additional eval7 baselines

swsuggs:
Still need to run Carla OD test set 2.0.0 after we receive it from Mike.

PR #1966, opened 2023-07-19
CARLA overhead object detection test data

yusong-tan:


swsuggs:
@jprokos26 could you double check that you can get the new dataset when you run a config?
This version has two splits ```"test_hallucination"``` and ```"test_disappearance"``` so just set the config accordingly:
```
"dataset": {
        . . .
        "eval_split": "test_hallucination",
        "name": "carla_over_obj_det_test"
        . . .
```

swsuggs:
I will add baseline results to PR #1959

jprokos26 (approved):

PR #1967, opened 2023-07-24
Bring tag to develop from master

mwartell:
There was a setuptools_scm unhappiness because the v0.18.0 tag lived on master but not develop.

Because a tag applies to one specific commit, merging master into develop gets the tagged commit into branch develop thus allowing work on that branch to see v0.18.0

This happened because I forgot to merge master→develop, and then forgot that I forgot.  😢


christopherwoodall (approved):
LGTM!

PR #1969, opened 2023-07-28
CARLA Attack Update with Adam Optimizer

yusong-tan:
Updated CARLA object detection and MOT attacks with the option to use Adam optimizer, which is found to create stronger attacks. The default scenario configs have also been modified to use the optimizer.


swsuggs (approved):
Looks great, thanks a lot Mike

PR #1973, opened 2023-08-03
Triggers and config kwargs

swsuggs:
- Test triggers for OD poisoning.

- Updated kwargs for targeted Carla configs based on discussions with Mike and updates that have already been applied to the other Carla configs.


deprit (approved):
Ran --check on the modified scenarios -- everything looks good.

PR #1975, opened 2023-08-07
Update split names in Carla dataset test

swsuggs:
Use the new split names in the Carla Overhead OD dataset unit test ("test_hallucination" and "test_dataset")


christopherwoodall (approved):
LGTM!

PR #1980, opened 2023-08-11
update np.bool reference

### LINKED ISSUE #1978: [BUG]: np has no attribute bool
### Description of the bug

armory/scenarios/poison.py", line 312, in filter_dataset:

```self.removed = (1 - self.indices_to_keep).astype(np.bool) ```

Throws:

```
AttributeError: module 'numpy' has no attribute 'bool'.                                                 
`np.bool` was a deprecated alias for the builtin `bool`. To avoid this error in existing code, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy s
calar type, use `np.bool_` here.                    
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations 
```

### Steps To Reproduce

Run a poison config with latest docker

### Additional Information

Related to #1971

swsuggs:
To fix #1978

### DIFF
--- a/armory/scenarios/poison.py
+++ b/armory/scenarios/poison.py
@@ -309,7 +309,7 @@ class Poison(Scenario):
         self.x_train = self.x_poison[indices_to_keep]
         self.y_train = self.y_poison[indices_to_keep]
         self.indices_to_keep = indices_to_keep
-        self.removed = (1 - self.indices_to_keep).astype(np.bool)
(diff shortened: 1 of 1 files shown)


christopherwoodall (approved):
LGTM~

Issue #1982, opened 2023-08-17
[BUG]: `export_batches` errors when using model with tensor output instead of numpy


jprokos26:
### Description of the bug

Need to sanitize model output with postprocessing function to ensure it is a numpy array (assuming this is the expected type). Additionally need some more verbose type information 😅 

### Steps To Reproduce

Run MITM objdet with `export_batches` on

### Additional Information

$ python -m armory run ./eval7/mitm_eval7_objdet/advpatch_rgb.json --num-eval-batches=1
2023-08-17 17:51:58  0s INFO     armory.eval.evaluator:__init__:30 Constructing Evaluator Object
2023-08-17 17:51:58  0s INFO     armory.eval.evaluator:__init__:61 Retrieving Environment Variables
2023-08-17 17:51:58  0s SUCCESS  armory.docker.images:is_image_local:133 found docker image mitm_eval7_objdet:1.1
2023-08-17 17:51:59  2s INFO     armory.docker.management:__init__:62 ARMORY Instance 7d8b44bd5d1b created.
2023-08-17 17:51:59  2s INFO     armory.eval.evaluator:_run_config:287 Running evaluation script
2023-08-17 17:52:00  0s INFO     armory.delayed_imports:<module>:7 Importing and configuring torch, tensorflow, and art, if available. This may take some time.
2023-08-17 17:52:04  4s INFO     armory.delayed_imports:<module>:30 Setting tf.config.experimental.set_memory_growth to True on all GPUs
2023-08-17 17:52:04  4s INFO     art.config:set_data_path:55 set ART_DATA_PATH to /tmp/.art/data
2023-08-17 17:52:05  5s INFO     art.config:set_data_path:55 set ART_DATA_PATH to /armory/saved_models/art
2023-08-17 17:52:07  7s INFO     matplotlib.font_manager:_load_fontmanager:1624 generated new fontManager
2023-08-17 17:52:11 10s INFO     armory.data.utils:maybe_download_weights_from_s3:47 Using available [MITM-objdet_rgb.pt](http://mitm-objdet_rgb.pt/) in Armory `saved_model_dir`
2023-08-17 17:52:16 16s INFO     armory.scenarios.scenario:load_dataset:257 Loading test dataset carla_over_obj_det_dev...
[...]

PR #1983, opened 2023-08-18
1982 export hotfix

jprokos26:
See #1982

### DIFF
--- a/armory/scenarios/scenario.py
+++ b/armory/scenarios/scenario.py
@@ -9,8 +9,8 @@ import sys
 import time
 from typing import Optional
 
-from tqdm import tqdm
 from torch import Tensor
+from tqdm import tqdm
 
 import armory
 from armory import Config, metrics, paths
--- a/armory/scenarios/scenario.py
+++ b/armory/scenarios/scenario.py
@@ -10,6 +10,7 @@ import time
 from typing import Optional
 
 from tqdm import tqdm
+from torch import Tensor
 
 import armory
 from armory import Config, metrics, paths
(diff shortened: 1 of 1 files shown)


jprokos26:
@swsuggs I couldn't find documentation about what the model is _supposed_ to output but when testing with other configs they all output numpy arrays; do we notate what the model output is required to be anywhere? Is there supposed to be a postprocessing function to handle this?

swsuggs:
@jprokos26 I think Armory's use of numpy is mainly to conform with ART.  I found this in the [doc](https://github.com/twosixlabs/armory/blob/4a5e808bfa02952618da9ca7d5ae06f793775a11/docs/faqs.md?plain=1#L32):

> Currently, the ART toolbox only accepts NumPy arrays as inputs to attacks, defenses
and model fitting. It is understood that this is an inefficient way to utilize data
since it requires a conversion from FrameworkTensor -> ndarray -> FrameworkTensor.

Which doesn't mention model output, but I would assume it applies to output as well.  I haven't dug into the ART documentation though.  This is a performer model that's outputting a tensor?  Perhaps they are doing some post-processing within the "model" but outside of the art wrapper that converts it back to a tensor?  If so, I think it would be better to edit their code (in gard-submissions) to return numpy rather than making armory conform to their submission.

jprokos26:
@swsuggs Got it, I suspected this was a performer issue, I just was not sure where the post-processing occurred. I'll make that change in gard-submissions and close this.
In the meantime anyone experiencing this issue can use [jprokos26:1982-export-hotfix](https://github.com/jprokos26/armory/tree/1982-export-hotfix).

PR #1985, opened 2023-08-22
iterate through sys.path instead of just checking cwd

jprokos26:
Makes it so you can put `patch_base_images` for carla od adv patch in any folder added to `local_repo_dir` in config.

### DIFF
--- a/armory/art_experimental/attacks/carla_obj_det_adversarial_patch.py
+++ b/armory/art_experimental/attacks/carla_obj_det_adversarial_patch.py
@@ -1,4 +1,5 @@
 import os
+import sys
 from typing import Optional
 
 from art.attacks.evasion.adversarial_patch.adversarial_patch_pytorch import (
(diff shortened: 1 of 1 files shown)


armory/art_experimental/attacks/carla_obj_det_adversarial_patch.py
```diff
@@ -57,11 +58,18 @@ def create_initial_image(self, size, hsv_lower_bound, hsv_upper_bound):
         patch_base_image_path = os.path.abspath(
             os.path.join(module_folder, self.patch_base_image)
         )
-        # if the image does not exist, check cwd
+        # if the image does not exist iterate through path
         if not os.path.exists(patch_base_image_path):
```
swsuggs, line 62: Would it make sense to combine this conditional (line 62) with the similar one on line 74?

jprokos26: This can certainly be cleaned up to avoid fallthroughs like this but the second check on 74 is needed to determine if it was in the path. Another option is to use for-break iterating through the path and checking for url only if the loop finishes.

Regardless, I am cleaning this code path up in #1904:
https://github.com/twosixlabs/armory/pull/1904/commits/bff3f021eee42482226231b9dac3da9deda52a97#diff-43528b7021f0d40c4760148beb2bdbb8e81013288e0d00e42ff57940600b7d12R124-R167

jprokos26: I created this PR more as a hotfix so people can use images anywhere off their armory git directory by including it in the `local_repo_path` since that gets added to your pythonpath. Happy to move the `fetch_file_or_url` function into this PR to prevent additional changes with #1904

swsuggs: Got it.  Fine either way

swsuggs (approved):

PR #1988, opened 2023-08-29
updating utils docs to include shape-gen

jprokos26:
Tiny change


christopherwoodall (approved):
Errors in CI are caused by underlying `tf-models` requirements seem to have reappeared.

PR #1974, opened 2023-08-07
Remove np.object references

swsuggs:
To fix #1971, the three references to ```np.object``` are replaced with ```object```, as advised by the error statement:

> To avoid this error in existing code, use `object` by itself. Doing this will not modify any behavior and is safe.

PR #2001, opened 2023-11-06
remove additional np.object references

swsuggs:
See #1974.  It seems that some were overlooked


mwartell (approved):
more straightforward type changes as in #1974

PR #2002, opened 2023-11-14
Fix np error

swsuggs:
We were creating an array out of a list of ints and single-element arrays.  Latest numpy does not support this.

This PR also brings a couple changes from master into develop.


mwartell (approved):
This is mostly back-merges to develop from 0.19.x with a valid one line change in posion.py

